PR #4, opened 2016-09-21
Added a DjangoToManyField to handle to-many relationships

conradev:
To-many relationships return a manager, which needs to be turned into an iterable (its queryset) using `maybe_queryset`. Let me know if this is a reasonable solution!

To-many relationships that are not a connections are currently broken.


syrusakbary:
Make sense!

syrusakbary:
(test should be passing on `master` when merging).

PR #6, opened 2016-09-21
Fixed a bug that was causing graphql_schmea command to not import properly

BossGrand:
There is a bug that is happening when running the graphql_schema command

AttributeError: 'Schema' object has no attribute 'startswith'

It was trying to import the schema even though it was already grabbing it with the add_arguments function


syrusakbary:
Makes sense! Thanks for the fix :)

PR #23, opened 2016-10-14
Fix #19 by using choice keys as enum keys, not choice descriptions

### LINKED ISSUE #19: Use choice keys, not descriptions, for enums
Currently, the automatically generated choices for Django fields correspond to the text labels passed in not the keys.

This is problematic because changing labels in the future would lead to all your enum values being updated. It's also contrary to what I would expect as a user.

morgante:
This fixes #19 so that only the keys (as used in the database) for choice enums. This is superior because descriptions might change over time but you don't want the underlying keys to change.

It also auto-prefixes the keys with "A_" if they would otherwise violate ENUM rule names.

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -7,6 +7,7 @@ from graphene.relay import is_node
 from graphene.types.datetime import DateTime
 from graphene.types.json import JSONString
 from graphene.utils.str_converters import to_const
+from graphql import assert_valid_name
 
 from .compat import (ArrayField, HStoreField, JSONField, RangeField,
                      RelatedObject, UUIDField)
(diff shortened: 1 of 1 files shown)


syrusakbary:
Good point! I will merge and then fix the tests 👍

graphene_django/converter.py
```diff
@@ -17,7 +18,12 @@
 
 
 def convert_choice_name(name):
-    return to_const(force_text(name))
+    name = to_const(force_text(name))
+    try:
+        assert_valid_name(name)
+    except AssertionError:
+        name = "A_%s" % name
```
Ian-Foote: This could break badly if python is run with optimisations on (`python -O`) because that disables `assert` statements, as used in `assert_valid_name`. The possible fixes are to update `assert_valid_name` to raise a different exception (`ValueError` probably) or to add a parallel function to `assert_valid_name` for use like this.

PR #25, opened 2016-10-17
Command indent

stegben:
It seems to be a missing part since this command is mentioned [here](http://docs.graphene-python.org/projects/django/en/latest/introspection.html#advanced-usage).

```bash
python manage.py graphql_schema --indent 2
```
or
```python
# in settings.py
GRAPHENE = {
    'SCHEMA': 'cookbook.schema.schema',
    'SCHEMA_INDENT': 2,
}
```


syrusakbary:
👍

Issue #8, opened 2016-09-28
cookbook example doesn't work?


zamesilyasa:
I am trying to run cookbook example and when I try to get item by ID, web client raises an error:

query {
ingredient(id: "SW5ncmVkaWVudE5vZGU6MQ==") { # actual id is takef rom the previous request
name
}
}
<img width="1018" alt="screen shot 2016-09-28 at 11 49 07" src="https://cloud.githubusercontent.com/assets/2279909/18909081/e52dcd50-857a-11e6-8e28-778b4692b630.png">

jtaylor32:
I followed the docs on [Cookbook](http://docs.graphene-python.org/projects/django/en/latest/tutorial.html) as well and it seems that those are not updated for the newest django-filter updates. Any idea what would be changing for the new filtering release?

Thanks!

rm--:
I also followed the docs to build the cookbook example.

An error occurred, so I checked the issues and ended here. After that I tried @jtaylor32 [graphene-tutorial repository](https://github.com/jtaylor32/graphene-tutorial) that looks very similar to my code. The same error occurs.

```
# $ git clone https://github.com/jtaylor32/graphene-tutorial                                                                                                                           Cloning into 'graphene-tutorial'...
remote: Counting objects: 22, done.
remote: Compressing objects: 100% (20/20), done.
remote: Total 22 (delta 0), reused 22 (delta 0), pack-reused 0
Unpacking objects: 100% (22/22), done.

# $ cd graphene-tutorial

# $ mkvirtualenv -a . graphene-tutorial
Setting project for graphene-tutorial to /Users/rm/used_repos/graphene-tutorial
pyenv-virtualenv: deactivate must be sourced. Run 'source deactivate' instead of 'deactivate'

# $ pip install -r requirements.txt
Collecting Django==1.10.2 (from -r requirements.txt (line 1))
  Using cached Django-1.10.2-py2.py3-none-any.whl
Collecting django-filter==0.15.2 (from -r requirements.txt (line 2))
  Using cached django_filter-0.15.2-py2.py3-none-any.whl
Collecting graphene==1.0.2 (from -r requirements.txt (line 3))
  Using cached graphene-1.0.2.tar.gz
Collecting graphene-django==1.0 (from -r requirements.txt (line 4))
  Using cached graphene-django-1.0.tar.gz
Collecting graphql-core==1.0 (from -r requirements.txt (line 5))
  Using cached graphql-core-1.0.tar.gz
[...]

PR #32, opened 2016-10-22
fix order_by_field bug

stegben:
#8 

I test on my own project with django-filter 0.15.2. My own test is pass now.


stegben:
It seems I misunderstand sth.

PR #38, opened 2016-10-31
Add support for batching several requests into one

michalochman:
Batch format compatible with [nodkz/react-relay-network-layer](https://github.com/nodkz/react-relay-network-layer).

Tested only `POST` requests with `Content-Type: application/json` since I do not think it would be used differently.

### DIFF
--- a/graphene_django/tests/test_views.py
+++ b/graphene_django/tests/test_views.py
@@ -8,13 +8,15 @@ except ImportError:
     from urllib.parse import urlencode
 
 
-def url_string(**url_params):
-    string = '/graphql'
-
+def url_string(string='/graphql', **url_params):
     if url_params:
         string += '?' + urlencode(url_params)
 
     return string
+
+
+def batch_url_string(**url_params):
+    return url_string('/graphql/batch', **url_params)
 
 
 def response_json(response):
@@ -22,6 +24,7 @@ def response_json(response):
 
 
 j = lambda **kwargs: json.dumps(kwargs)
+jl = lambda **kwargs: json.dumps([kwargs])
 
 
 def test_graphiql_is_enabled(client):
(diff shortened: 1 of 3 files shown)


michalochman:
Not sure why the `UnboundLocalError` suddenly occurred in my build on Travis. `master` tests probably never checked the `else` branch of `if execution_result` condition. Anyway, the issue is now fixed.

syrusakbary:
Looks great!!

Thanks for such a great PR!

jole78:
First of all, thanks a lot for adding this. 
A question/request though.
As I understand, this is made for a specific client implementation of batched queries...??

I'm using [apollo-client](https://github.com/apollostack/apollo-client) which also supports query batching but it expects the results to come back the same way as "normal" queries, in a data prop.
From what I can see the response format for a batched query looks like this:
```
                response = {
                     'id': id,
                     'payload': response,
                     'status': status_code,
                 }
```

where response['data'] is overwritten with this 'new' response instead.

I did make this work by sub classing the GraphQLView and basically removed those [lines](https://github.com/graphql-python/graphene-django/pull/38/files#diff-ccb88e1bb3e2dd99eeb680ebfdc3e736R162) but it seems a bit cumbersome since I have to copy/paste this entire function:
`def get_response(self, request, data, show_graphiql=False):`

and make the changes in there.

Would it be an option to make the response format more pluggable in some way...perhaps like a function that we can overwrite instead that takes the execution result and returns the response, and that would default to response['data']. Then one could sub class the GraphQLView and just overwrite that function to return a different response format if so desired.

michalochman:
@jole78 try setting response transform function as `batch` property of `GraphQLView` and then transforming the actual response using that function in the code you linked.

I would also suggest making the current implementation default to not break existing code.

PR #39, opened 2016-11-03
tutorial was missing import

chriscauley:
I know this is really minor, but any friction reduction is good, right?


syrusakbary:
Any friction reduction is good and this contributions helps a lot for moving forward!

PR #43, opened 2016-11-03
fixes #8, base import of field no longer works for query property. Us…

### LINKED ISSUE #8: cookbook example doesn't work?

chriscauley:
…e Node.Field instead


syrusakbary:
Great improvement to the cookbook!!
Thanks for the PR :)

Merging 👍

PR #44, opened 2016-11-05
Fix graphql query guide url

drzix:
Hi,

The original path refers an unresolved url.
maybe it should be fixed...?

Thanks.


syrusakbary:
Thanks for the fix!

Issue #46, opened 2016-11-10
Fields that end with '+' won't serialize properly


timothyjlaurent:
https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.ForeignKey.related_query_name

Django fields with '+' on the end do not make backreferences and also break graphene-django

syrusakbary:
Fixed with #47.

Closing the issue :)

PR #47, opened 2016-11-10
solves #46 don\'t use fields that end in a plus

timothyjlaurent:


graphene_django/types.py
```diff
@@ -26,9 +26,12 @@ def construct_fields(options):
         is_not_in_only = only_fields and name not in options.only_fields
         is_already_created = name in options.fields
         is_excluded = name in exclude_fields or is_already_created
-        if is_not_in_only or is_excluded:
+        # https://docs.djangoproject.com/en/1.10/ref/models/fields/#django.db.models.ForeignKey.related_query_name
+        is_no_backref = str.endswith(str(name), '+')
```
syrusakbary: `str(name).endswith('+')` I think is a little bit more legible :)

syrusakbary:
Besides the small nit, everything looks great! :)

timothyjlaurent:
OK I changed the `endswith` per your suggestion, @syrusakbary

PR #65, opened 2016-11-23
Updated Django Filter integration to 1.0

syrusakbary:
This PR updates the integration with Django Filter, deprecating the usage of `order_by` and `filter_order_by` in Graphene Django.

The only compatible `django-filter` version will be `1.0` and above, if your project requires `django-filter` `0.1X` please use `graphene-django<="1.1"`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,9 +22,8 @@ before_install:
 install:
 - |
   if [ "$TEST_TYPE" = build ]; then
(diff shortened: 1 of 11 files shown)


Svoka:
Why did you do this?

syrusakbary:
@Svoka could be possible to know what you mean exactly? Was this PR (from 2016) causing some regression on your side?

Svoka:
@syrusakbary somehow yes. We are in process of updating old deps and now we have to implement it.

I don't follow much, but maybe you have something like roadmap or blog post where you explaining why you decide to remove this features?

Issue #63, opened 2016-11-19
`ManyToOneRel` fields are wrongly named, causing the field resolver to return every single record.


yen223:
<This is on Django 1.8.16>

A foreign key on model `Article` to model `Reporter` will automatically create a reverse field called `article_set` on the `Reporter` model. However for some reason, the corresponding related field name does not contain the `_set` suffix:

```
>> Reporter.article_set.related.name
'article'
```

This causes the constructed field on the ReporterObjectType to be wrongly named `article` at https://github.com/graphql-python/graphene-django/blob/master/graphene_django/types.py#L25, which in turn causes the `Reporter.article` field to be unresolvable, as the Reporter instance lacks the `article` attribute. 

That in turn causes https://github.com/graphql-python/graphene-django/blob/master/graphene_django/fields.py#L49 to default to the `Article` model's manager, which results in every Article instance being returned.

nickhudkins:
@yen223 just ran into this with User Models, where a ManyToManyRel results in a field called "user" being created on "user", returning all users. Have you tried to address this?

yen223:
@nickhudkins My workaround was to exclude all m2m fields, and redefine the m2m fields with explicit resolvers.

nickhudkins:
@yen223 same here 👍  thanks for the feedback.

PR #74, opened 2016-12-13
#63 Get name of reverse_fields from model.__dict__

momamene:
If related name is not set, graphene-django cannot find reverse related objects. 

It is because `default_resolver` in `graphene.typemap` just get `attname` from `root`. 

Actually, the name of related objects in django model is assigned as `get_accessor_name`, not just `name` or `attname`. 

So it is needed to get name from information of `model.__dict__.items`, not `field.name`.

By modifiing like this, schema can find the corresponding related field name containing the `_set` suffix.

### DIFF
--- a/graphene_django/types.py
+++ b/graphene_django/types.py
@@ -33,8 +33,6 @@ def construct_fields(options):
             # Or when there is no back reference.
             continue
         converted = convert_django_field_with_choices(field, options.registry)
-        if not converted:
-            continue
         fields[name] = converted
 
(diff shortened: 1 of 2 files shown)


syrusakbary:
Great PR! I think is quite useful.

However if we can cover this case with a new test case would be great, as it assures this issue will not happen in the future!

syrusakbary:
@momamene let me know if you want to add a test case, or you need some assistance from me :)

Issue #30, opened 2016-10-21
DjangoFilterConnectionField and resolve support


drakon:
I found that the following example from the [docs](http://docs.graphene-python.org/projects/django/en/latest/authorization.html) doesn't work correctly.

The thing is that if there is a resolver for a DjangoFilterConnectionField like in the example, the filtering and ordering created by DjangoFilterConnectionField will be ignored. Details below.

```
class Query(ObjectType):
    my_posts = DjangoFilterConnectionField(CategoryNode)

    def resolve_my_posts(self, args, context, info):
        # context will reference to the Django request
        if not context.user.is_authenticated():
            return Post.objects.none()
        else:
            return Post.objects.filter(owner=context.user)
```

The issues seems to be that DjangoFilterConnectionField ignores what the resolver returns. 

```
@staticmethod
def connection_resolver(resolver, connection, default_manager, filterset_class, filtering_args,
                        root, args, context, info):
    filter_kwargs = {k: v for k, v in args.items() if k in filtering_args}
    order = args.get('order_by', None)
    qs = default_manager.get_queryset()
    if order:
        qs = qs.order_by(order)
    qs = filterset_class(data=filter_kwargs, queryset=qs)

    return DjangoConnectionField.connection_resolver(resolver, connection, qs, root, args, context, info)
```

Here the QuerySet is built just fine but the resolver does not have an impact on the QuerySet (which is fine if no resolver is used anyway).

The problem arrises when the DjangoConnectionField resolves the query.

```
@staticmethod
def connection_resolver(resolver, connection, default_manager, root, args, context, info):
    iterable = resolver(root, args, context, info)
    if iterable is None:
        iterable = default_manager
    iterable = maybe_queryset(iterable)
    if isinstance(iterable, QuerySet):
        _len = iterable.count()
    else:
        _len = len(iterable)
    connection = connection_from_list_slice(
        iterable,
        args,
        slice_start=0,
        list_length=_len,
        list_slice_length=_len,
        connection_type=connection,
        edge_type=connection.Edge,
        pageinfo_type=PageInfo,
    )
    connection.iterable = iterable
    connection.length = _len
    return connection
```

Here we consider the default_manager (which is the QuerySet that was built by DjangoFilterConnectionField) **only** if there resolver is None. 

So you either can have the DjangoFilterConnection field **or** a resolver. But it should be possible to use both (to limit an initial QuerySet and check of authorisation, etc.). 
### Solution

[...]

Issue #21, opened 2016-10-13
Filtering fields ignored inside fragments


felix-zg:
Hello guys,

I'm trying to run this query and it's completely ignore the 'employeeType'(or any other) filter:

```
{
 business(id: "") {
    ...F2
  }
}

fragment F2 on BusinessNode {
  employee(employeeType:":ceo") {
    edges {
      node {
        status
        employeeType
        id
      }
    }
  }
}
```

On the second hand, this query works perfectly fine:

```
employee(employeeType:":ceo") {
    edges {
     node {
       status
       employeeType
       id
      }
    }
  }
```

Both of the objects(BusinessNode and EmployeeNode) are DjangoObjectType.

Thanks,
Felix

KaRi94:
Same problem +1

syrusakbary:
That's a weird issue. Maybe is a bug in `graphql-core` or `graphene` itself.

Could you post a fully working example that raises the issue? (a gist or a PR with a failing test would also work :) )

felix-zg:
I used the cookbook example, and it's works perfectly fine:

```
 {
  allIngredients(notes: "Comes from a cow") {
    edges {
      node {
        id
        notes
        name
      }
    }
  }
}
```

but with fragment:
```
{
  allCategories(name: "Dairy") {
    edges {
      node {
        ...F2
      }
    }
  }
}

fragment F2 on CategoryNode {
  ingredients(notes: "Comes from a cow") {
    edges {
      node {
        id
        notes
        name
      }
    }
  }
}
```

BossGrand:
Seeing the same issue

BossGrand:
It doesn't even need to be a fragment for this to happen. If you run this query on the cookbook example graphiql page you can see the bug. The name filter is completely ignored 
    

  query{
      allCategories {
        edges {
           node {
               id,
               ingredients(name_Icontains:"beef") {
                  edges {
                     node {
                        id,
                        name
                     }
                  }
               }
            }
         }
      }
   }

tom-zeit:
This issue prevents us from upgrading to graphene 1.0 :/ 
Anyway has a lead on how to solve it?

timothyjlaurent:
I'm just noticing this too after writing tests for our app migration... This is a real problem... Need to be able to to filter on nested nodes. 

```
            query ($docid: String!, $label: String!) {
                allGene(docid_In: $docid) {  <-- this filter works
                    edges {
                        node {
                            id
                            docid
[...]

PR #82, opened 2016-12-29
Fixes filtering in nested nodes.

pchinea:
Fixes error filtering sub-nodes. The relationships were resolved but not filtered as it was notified in issues #21 and #30.

For example:

```
query{
  allCategories {
    edges {
       node {
           id,
           ingredients(name_Icontains:"beef") {
              edges {
                 node {
                    id,
                    name
                 }
              }
           }
        }
     }
  }
}
```

### DIFF
--- a/graphene_django/fields.py
+++ b/graphene_django/fields.py
@@ -49,6 +49,7 @@ class DjangoConnectionField(ConnectionField):
             iterable = default_manager
         iterable = maybe_queryset(iterable)
         if isinstance(iterable, QuerySet):
+            iterable &= maybe_queryset(default_manager)
             _len = iterable.count()
         else:
             _len = len(iterable)
--- a/graphene_django/tests/test_query.py
+++ b/graphene_django/tests/test_query.py
@@ -8,6 +8,7 @@ from py.test import raises
 import graphene
 from graphene.relay import Node
 
+from ..utils import DJANGO_FILTER_INSTALLED
 from ..compat import MissingType, RangeField
 from ..fields import DjangoConnectionField
 from ..types import DjangoObjectType
@@ -282,6 +283,9 @@ def test_should_query_connectionfields():
         }
     }
 
+
+@pytest.mark.skipif(not DJANGO_FILTER_INSTALLED,
+                    reason="django-filter should be installed")
 def test_should_query_node_filtering():
     class ReporterType(DjangoObjectType):
 
@@ -319,7 +323,6 @@ def test_should_query_node_filtering():
         editor=r,
         lang='en'
     )
-
 
     schema = graphene.Schema(query=Query)
     query = '''
--- a/graphene_django/tests/test_query.py
+++ b/graphene_django/tests/test_query.py
@@ -345,25 +345,21 @@ def test_should_query_node_filtering():
     '''
 
     expected = {
-		"allReporters": {
-			"edges": [
-				{
-					"node": {
-						"id": "UmVwb3J0ZXJUeXBlOjE=", 
-						"articles": {
-							"edges": [
-								{
-									"node": {
-										"id": "QXJ0aWNsZVR5cGU6MQ=="
-									}
-								}
-							]
-						}
-					}
-				}
-			]
-		}
-	}
+        'allReporters': {
+            'edges': [{
+                'node': {
+                    'id': 'UmVwb3J0ZXJUeXBlOjE=',
+                    'articles': {
+                        'edges': [{
+                            'node': {
+                                'id': 'QXJ0aWNsZVR5cGU6MQ=='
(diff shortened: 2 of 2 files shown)


syrusakbary:
Great PR! Could be possible to have a test too showcasing how it the filtering got fixed so we don't fall in this bug in the future?

Thanks! (and happy holidays!)

drakon:
Ah, very nice. 
On question though: What if the resolver returns a list? Wouldn't this list remain unaffected by the filter?

pchinea:
@syrusakbary I added a test for this issue. Please, check it to ensure everything is right.

Greetings.

pchinea:
@drakon The problem in #80 was in the line: `qs = filterset_class(data=filter_kwargs, queryset=qs)`, if `qs` is a list instead a queryset it fails. Sincerely I don't have a deep knowledge on this module, but I think in a real application it always will be a queryset so your modification will work too. Anyway I'm not sure about it so I made the simplest modification assuming `resolve` could return a list.

drakon:
Yeah, that I got. Thanks for the hint! :)

My point was that in that case the connection_resolver method still is broken (not an issue that is new with your PR). Assuming it returns a list, default_manager never gets considered and therefore I'd argue that a test considering this would fail.

Otherwise your solution is better. Mine originally was more of an easy workaround to make it work in case someone needs this functionality. So I didn't have to fork and fix the issue. Your fix seems like a better solution.

pchinea:
@drakon You are right, when I coded the fix I realized it won't be filtered if a list is returned by resolver and I think there are no efficient way to filter the list. But this is my question: Is there a real case in which a list is returned? I think it shouldn't do it, because you need a queryset to be able to apply a filter. Maybe using lists in tests is an error.

Anyway, I would like to know @syrusakbary 's opinion on this matter

drakon:
Yeah, exactly my thoughts! Maybe it is required by some GraphQL spec. Also interested in @syrusakbary 's thoughts on this. If it's really a non intended use case we could at least issue a warning that the filters get ignored when the resolver returns a list.

syrusakbary:
In general a `DjangoFilterConnectionField` would only work when the resolver returns a `QuerySet`.
The `DjangoConnectionField` supports a list returned in the resolver because `list/Iterables` are accepted as return values in normal `ConnectionFields`, plus the willing to filter manually a queryset and return a list.

However this PR handles this cases perfectly and also with a good test coverage... so everything looks great!

nickhudkins:
@syrusakbary thanks for getting this merged into master. Is there a planned release that will have this included in it?

imsickofmaps:
Also curious as to next pypi release... 1.2.1 was released on 2016-12-16.

syrusakbary:
Hi @nickhudkins @imsickofmaps.
The release is waiting on a new feature of graphene that will let this package to improve significantly the registry logic.
I expect to release graphene 1.2 today, so today or tomorrow will be the release of graphene-django 1.3.

Sorry for the delay!

pchinea:
@syrusakbary This PR introduces a serious efficiency problem, I'll try to push a fix today. I hope this can be included in the next release.

syrusakbary:
@pchinea I will wait few days and revert in case is not solved so we can release a new version.

antonkhelou:
Just thought I would let you guys know that there's an issue with this fix, if you do something like:

```
    def resolve_tasks(self, args, context, info):
        tasks = Task.objects.filter(
            project=context.user.project
        )[:MAX_PER_PAGE]

        return tasks
```

It will cause an error due to the slicing with `[:MAX_PER_PAGE]`:

`"Cannot combine queries once a slice has been taken."`

The message says it all: Some code does some combination of QuerySets after the resolver has finished it computation.

pchinea:
@antonkhelou If you want to use a custom resolver instead of the automatic resolver you should return a list to avoid further filtering, in this case:

```
    def resolve_tasks(self, args, context, info):
        tasks = Task.objects.filter(
            project=context.user.project
        )[:OFFSET]

        return list(tasks)
```

Greetings

antonkhelou:
@pchinea I see. But now the filtering does not work. Both of the following queries return the same thing:

```
query{
  tasks {
    edges {
      node {
        id
        name
      }
    }
  }
}
```
```
query{
  tasks(name:"asd") {
    edges {
      node {
        id
        name
      }
    }
  }
}
```

Is this what issue #84 is referring to?

pchinea:
@antonkhelou Anyway, I think there are no way to fix this, you can't filter a queryset after slicing. This isn't a graphene or django limitation, this is a SQL limitation (you can't use WHERE clause after LIMIT clause). In this case I'm afraid you'll need to find a way to get the final query in the resolver.

PR #86, opened 2017-01-05
Avoids collisions in choices names conversion.

pchinea:
When a django field with choices is converted to an enum and the choices names are normalized, two or more names could be normalized to the same name producing a collision. For example: "Etc/GMT+1" and "Etc/GMT-1" will be converted to "ETC_GMT_1". This patch avoids collisions generating a new unique name when a collision is produced.

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -34,7 +34,7 @@ def get_choices(choices):
                 yield choice
         else:
             name = convert_choice_name(value)
-            if name in converted_names:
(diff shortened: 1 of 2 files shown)


fmartins-in-loggi:
Hey buddies! 
This PR solves a problem that I am facing... 
Thank you @pchinea

jeffbr13:
Thank you @pchinea, this fixes a problem that I was having with pytz TimezoneFields as well!

PR #94, opened 2017-01-23
Return 400 status code if result has an error

dotsbb:
Is there any reason why we should return 200 status code on response with errors?


syrusakbary:
Hi @dotsbb, thanks for your work.

Based on the GraphQL spec and the express-js official implementation, the current behavior is the correct one.
You can take a look here: https://github.com/graphql/express-graphql/blob/master/src/__tests__/http-test.js#L925-L946

dotsbb:
@syrusakbary Thanks, that is exactly what i was looking for.

Issue #79, opened 2016-12-23
Permission System


pizzapanther:
I would like to add a permission system but want to some feedback on the API before I implement.

You would have two options and I'm proposing to add both:

## Option 1: Custom queryset method

This option would let you overwrite how a queryset is filtered.

```python
class UserNode(DjangoObjectType):
  class Meta:
    model = User
    interfaces = (relay.Node,)
    only_fields = ('email', 'first_name', 'last_name')

  @classmethod
  def get_queryset (cls, queryset, args, request, info):
    return queryset.filter(owner=request.user)
```

## Option 2: Permissions List

This option would setup a Meta API to use to define permissions

```python
def auth_required(queryset, args, request, info):
  if request.user.is_authenticated():
    return queryset

  return queryset.none()

class UserNode(DjangoObjectType):
  class Meta:
    model = User
    interfaces = (relay.Node,)
    only_fields = ('email', 'first_name', 'last_name')
    permissions = [auth_required]
```

If these look like good APIs then I'll implement.

pizzapanther:
Also note `DjangoConnectionField` and `get_node` for NodeFields would know to check for these APIs

fmartins:
@pizzapanther

I liked of your suggestion to implement a permission system. Please, could you this other issue? https://github.com/graphql-python/graphene/issues/385

At this moment I am interested and available to do this, what do you about we plan this architecture and works together in this?

Thank you!

nickhudkins:
@pizzapanther perhaps we could use an array of functions or strings. If listed as strings, we can check built in django permission code names and verify context.user against it. Thoughts?

fmartins-in-loggi:
@valdergallo @romulorosa

In our case, we have over a hundred objects that extends `DjangoObjectType` and already using a middleware solution to avoid unauthorized access to queries and mutations that add `user` to `context` dictionary.

Assuming that in `context.user` exists a method called `has_perm` and `has_perms`. Using the Django perms we can build an interface that sounds good for us. See the example above:

```
class MyCustomNode(DjangoObjectType):
    my_custom_field = graphene.List(String)

    def resolve_my_custom_field(self, args, context, info):
        """Here, is not necessary to inspect if `context.user` has
        permission to view this field, because the framework handle this 
        internally.
        """"
        return 1

    class Meta(object):
        model = MyCustomObjectFromDomain
        only_fields = (
            'field_a_from_model',
            'field_b_from_model',
        )
        permissons = {
[...]

PR #95, opened 2017-01-24
#79 feat(permissions) add decorator has_perms for nodes

valdergallo:
I created one decorator to control the permissions

```
Check if have user logged and permissions to see some value

Example:
------------
from graphene_django.decorators import has_pems

class CityNode(DjangoObjectType):
    class Meta(object):
        interfaces = (relay.Node,)
        model = City
        only_fields = (
           'name', 'locality', 'slug', 'state', 'active',
        )

        @has_perms(["django_city_app.can_see_location"])
        def resolve_location(self, args, context, info):
            return self.locality.pos

```


graphene_django/decorators.py
```diff
@@ -0,0 +1,51 @@
...
+            method.permissions = []
+        else:
+            method.permissions = permissions
+
+        @wraps(method)
+        def wrapper(*args, **kwargs):
+            context = kwargs.get('context', dict(zip(method.func_code.co_varnames,
+                                                     args)).get('context', None))
+            if not context:
+                return HttpResponseForbidden('Forbidden. No context, no access.')
```
syrusakbary: I think is better to use the base `Exception` or custom created ones for this case.

### Why?
The `HTTPResponse*` exception implies only one triggered per request, which is not the case as could be triggered multiple times (one per each field resolution).

graphene_django/decorators.py
```diff
@@ -0,0 +1,51 @@
...
+            try:
+                user = context.user
+            except AttributeError:
+                return HttpResponseForbidden('Forbidden. No request.')
```
syrusakbary: Same, changing this to a normal `Exception` should be fine too :)

graphene_django/decorators.py
```diff
@@ -0,0 +1,51 @@
...
+
+            if method.permissions and not user.has_perms(method.permissions):
+                return HttpResponseForbidden('Forbidden. User without access')
```
syrusakbary: Same ^_^

syrusakbary:
I really like this PR! Other than the small nit comments I think should be good to go! :)

patrick91:
This looks amazing @valdergallo, do you have time to finish you work on it? I'd be happy to help if you don't :)

firaskafri:
@jmerkow still interested to put this together?

phalt:

PR #99, opened 2017-01-27
accept connections for DjangoFilterConnectionField

### LINKED ISSUE #98: DjangoFilterConnectionField does not accept Connections as type
Standard ConnectionFields allow you to pass a custom Connection in as the type for the field. This allows for defining Connection Fields like `totalCount` etc...

DjangoFilterConnectionField does not currently support this.

nickhudkins:
This resolves #98 

@syrusakbary am having trouble getting the test suite running locally, so pardon the lack of tests.

### DIFF
--- a/graphene_django/filter/fields.py
+++ b/graphene_django/filter/fields.py
@@ -1,6 +1,7 @@
 from functools import partial
 
 from ..fields import DjangoConnectionField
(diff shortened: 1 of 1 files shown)


syrusakbary:
Great addition!

PR #104, opened 2017-02-02
Allow nodes to skip the registry

Tritlo:
I've added an option to nodes to skip the registry. This allows the user to control whether the node implementation of a model gets used. My use case is that I want to enable:

    class PrivateProfileNode(DjangoObjectType):
      email = graphene.Field(graphene.String)
      username = graphene.Field(graphene.String)

      class Meta:
        model = Profile
        interfaces = (graphene.relay.Node,)
        skip_registry = True

    class ProfileNode(DjangoObjectType):
      username = graphene.Field(graphene.String)

      class Meta:
        model = Profile
        interfaces = (graphene.relay.Node,)

without having to worry about the email leaking out via some other nodes.

### DIFF
--- a/graphene_django/registry.py
+++ b/graphene_django/registry.py
@@ -13,7 +13,8 @@ class Registry(object):
         # assert self.get_type_for_model(cls._meta.model) == cls, (
         #     'Multiple DjangoObjectTypes registered for "{}"'.format(cls._meta.model)
         # )
-        self._registry[cls._meta.model] = cls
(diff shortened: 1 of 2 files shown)


syrusakbary:
Interesting approach! Will think a little bit more about it, but I think is a good idea :)

syrusakbary:
With the PR #115, this will be done but using `skip_global_registry` instead :)

PR #105, opened 2017-02-03
Exception raises when GET request is sent to batch endpoint

femesq:
Was getting:
```
ValueError at /graphql/batch/
max() arg is an empty sequence
```


syrusakbary:
This is fixed now with #117.
Will publish a new package to PyPI soon :)

PR #123, opened 2017-03-02
Remove payload key from response and stick to original format.

### LINKED ISSUE #122: Incorrect response format for batched queries
I've been working with `graphene-django` for the last 3 weeks, and have been implementing it with `apollo-client` on the frontend. I just switched on batched queries, and determined that `graphene-django` is trying to return a list of responses in the following format:

`
response = [
{
  id,
  payload,
  status
},
...
]
`

I can't seem to find any documentation that indicates that this should be the format for batched responses. As it is, this breaks `apollo-client`. I propose just returning a list of responses to adhere to the expected format.

HorizonXP:
Fixes #122

### DIFF
--- a/graphene_django/tests/test_views.py
+++ b/graphene_django/tests/test_views.py
@@ -178,7 +178,7 @@ def test_batch_allows_post_with_json_encoding(client):
     assert response.status_code == 200
     assert response_json(response) == [{
         'id': 1,
-        'payload': { 'data': {'test': "Hello World"} },
+        'data': {'test': "Hello World"},
         'status': 200,
     }]
 
@@ -233,7 +233,7 @@ def test_batch_supports_post_json_query_with_string_variables(client):
     assert response.status_code == 200
     assert response_json(response) == [{
         'id': 1,
-        'payload': { 'data': {'test': "Hello Dolly"} },
+        'data': {'test': "Hello Dolly"},
         'status': 200,
     }]
 
@@ -260,7 +260,7 @@ def test_batch_supports_post_json_query_with_json_variables(client):
     assert response.status_code == 200
     assert response_json(response) == [{
         'id': 1,
-        'payload': { 'data': {'test': "Hello Dolly"} },
+        'data': {'test': "Hello Dolly"},
         'status': 200,
     }]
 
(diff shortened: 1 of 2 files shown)


syrusakbary:
Hi @HorizonXP, thanks for the PR!

The batching support was initially coded for supporting `react-relay-network-layer` (as in https://github.com/nodkz/react-relay-network-layer/blob/4e5edc9096a8f64adff9722adea4f41f8f8f198a/test/batch.test.js#L42-L45 )

I thought `apollo-client` was using the same strategy for batching.

@helfer @nodkz it seems the batching query strategy in `apollo-client` and `react-relay-network-layer` differs, could be possible to standardize one or other?

### Apollo Client

```json
[
  {
    "id": 1,
    "errors": [
      { "location": 1, "message": "major error" },
    ],
  },
  { "id": 2, "data": {} },
]
```
### react-relay-network-layer

```json
{
  "status": 200,
  "body": [
    {
      "id": 1,
      "payload": {
        "errors": [
          { "location": 1, "message": "major error" },
        ],
      },
    },
    { "id": 2, "payload": { "data": {} } },
  ],
}
```

helfer:
@syrusakbary It's actually even simpler for Apollo as it doesn't use an `id` field. From our perspective there's no point in adding `id` if all responses come together anyway. If you don't want to send them together, it's a different transport anyway.

I also think the nodkz batching is a bit different from what you described. Status is present on every response: `[{id, status, payload }, {id, status, payload}, ... ]`.

To harmonize the two formats, I would propose dropping the `payload` field and simply expanding its contents into the response directly. Apollo Client can safely ignore the `id` and `status` fields, and for nodkz batching it's just a very simple change. So the format I'm proposing for graphene-django would look like this:

```
[{
    id,
    status,
    data,
    errors
  },
  // etc ...
] 
```

What do you think?

PS: @HorizonXP thanks for making the PR to prompt this discussion!

helfer:
@HorizonXP make sure you also change the tests, and not just the code. Currently the tests are failing because they still expect `payload`.

nodkz:
@syrusakbary @helfer i'll look tomorrow what can i do. I'd like idea of harmonization of two formats.

HorizonXP:
I'll update the PR so that it passes the tests, but I can't test against `react-relay-network-layer`. @nodkz, I trust that you could do this yourself? It looks like it's an external change that shouldn't affect `graphene-django`.

nodkz:
Just published `react-relay-network-layer@1.5.0` now supports its batch format and apollo batch format. So @syrusakbary you may safely migrate.

syrusakbary:
Thanks for the great work! @nodkz 

Merging :)

PR #125, opened 2017-03-05
Django permissions required at Nodes, Mutations and DjangoFilterConnectionField

CarlosMart626:
Django permissions integration

### DIFF
--- a/graphene_django/tests/test_auth.py
+++ b/graphene_django/tests/test_auth.py
@@ -102,7 +102,8 @@ if DJANGO_FILTER_INSTALLED:
 
 class QueryRoot(ObjectType):
     pet = relay.Node.Field(PetNode)
-    pets = PetFilterConnection(PetNode)
+    if DJANGO_FILTER_INSTALLED:
+        pets = PetFilterConnection(PetNode)
 
 
 class MutationRoot(ObjectType):


CarlosMart626:
Hi @syrusakbary,  Django Filter 1.0.1 does not support Django 1.6 and Django 1.7 so update travis? or set Django Filter dependency to 0.11.0?

I've done some changes to use Django permissions and I would like to have some feedback 👍

CarlosMart626:
Hi!, any feedback?

BossGrand:
You should get this to pass the ci tests, I think you'll be more likely to get feedback from the core team if you do that

CarlosMart626:
Thanks for your feedback @BossGrand I'm going to work to fix the issues and create a new pull request.

CarlosMart626:
After a lot of testing I've updated and pass all tests. Created auth mixins to require permissions at nodes and mutations also created AuthDjangoFilterConnectionField to create filters with permissions required.

CarlosMart626:
Hi @syrusakbary, I would like to have your feedback about these changes. I don't know if it fits in the project or if there is something to review and do it better.

Thanks.

spockNinja:
We definitely need some sort of permissions implementation. There have been a few different approaches proposed. I will try to aggregate the ideas and get discussion for which one we should implement.

CarlosMart626:
Hi @spockNinja I've done some changes to this solution to add more flexibility. Can I help upgrading this solution to version 2.0?

PR #132, opened 2017-03-10
Add tests for DjangoObjectType only/exclude_fields

samdroid-apps:
I was planning to write a new feature to rename fields on a DjangoObjectType, because I wanted my foreign key relationships to have nice reverse names.  So I started working on that by writing these tests.

However, I eventually opened the "tests/models.py" field and saw how the "related_name" can be used to rename the relationship.  So I didn't need to write the feature any more, but tests are still cool :)


syrusakbary:
Thanks for the contribution!

PR #126, opened 2017-03-05
[WIP] Fixes DjangoFilterConnectionField filter slices being ignored with a custom resolver

NiallBrickell:
TODO:
- [ ] Fix tests

PR #151, opened 2017-04-15
Fixed FilterConnectionFields with limit in the FilterSet

syrusakbary:
Improved #126 tests, by chaining the querysets and mixing the limits.


tillkolter:
@syrusakbary I found out that `merge_queryset` (or better `QuerySet.__add__`) is cleaning extra/annotated fields. My resolver query is rather complex and need that annotated column for sorting.

PR #160, opened 2017-04-28
Add ability for model types to be placed on input fields [WIP] feedback wanted

colanconnon:
Add the ability to have a model type on the input class in mutations for more code reuse. This is my first time contributing so some feedback is welcome!

### DIFF
--- a/graphene_django/input_types.py
+++ b/graphene_django/input_types.py
@@ -1,0 +1,7 @@
+import collections
+from graphene_django import converter, utils
+
+from graphene.types.utils import merge
+from graphene.utils.is_base_type import is_base_type
+
+
--- a/graphene_django/tests/test_types.py
+++ b/graphene_django/tests/test_types.py
@@ -1,10 +1,11 @@
 from mock import patch
 
-from graphene import Interface, ObjectType, Schema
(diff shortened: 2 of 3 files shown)


picturedots:
Thanks for this -- tried and it and I get errors if my model has a field called "name".  
unhashable type: 'String' -- 
/usr/local/lib/python3.5/site-packages/graphql/type/typemap.py in reducer

59.         if type.name in map:

spockNinja:
Thanks for the idea. An easy way to generate `Input` based on models is definitely useful.

I'll point out a few similar alternatives.

1. There is already support for mutations that utilize existing DRF Serializers. [Docs here](http://docs.graphene-python.org/projects/django/en/latest/rest-framework/#mutation)
2. There is [another PR](https://github.com/graphql-python/graphene-django/pull/217) in the works to do the same with native Django Forms.

Both of those options allow for validation and such to be handled by the serializer/form. They also have a default `mutate` that does what you would expect, validating and saving the serializer/form. So I *think* those two approaches should cover this functionality.

Are there any advantages of this approach over those two solutions that I am not thinking of?

firaskafri:
@colanconnon still interested to work on this?

phalt:

PR #166, opened 2017-05-02
fix select_related and prefetch_related set in resolver getting lost

jsurloppe:
Despite prefetching relation in resolve_ method, the relations are lost and filter redo n+1 queries.

This patch may introduce a reverse bug where relations prefetched in the default_queryset are lost, i haven't tested but i think this is way less problematic.


jsurloppe:
Closing for now, i want to check something, will open another if needed.

syrusakbary:
Hey @jsurloppe, could you add some tests that showcases the issue so we assure that there is no regression in the future?

BossGrand:
I think this might be causing the bug in #173

syrusakbary:
Closing this PR in favor of #315

PR #181, opened 2017-05-22
Fix ManyToMany schema mapping

### LINKED ISSUE #155: Schema incorrectly typing ManyToMany field
Hi,

I'm new to graphene (and django, really) but have come across a stumbling block.

I have the following django models:

```python
class Person(models.Model):
    name = models.CharField(max_length=100, blank=False)

class Booking(models.Model):
    people = models.ManyToManyField(Person, related_name="bookings", blank=False)
```

One person has many bookings; one booking has many people.

Then I have the following object types in my graphene schema:

```python
class PersonType(DjangoObjectType):
    class Meta:
        model = Person

class BookingType(DjangoObjectType):
    class Meta:
        model = Booking

class Query(graphene.AbstractType):
    all_people = graphene.List(PeopleType)
    all_bookings = graphene.List(BookingType)

    def resolve_all_people(self, *args, **kwargs):
        return Person.objects.all()

    def resolve_all_bookings(self, *args, **kwargs):
        return Booking.objects.all()
```

everything else is boilerplate from the graphene/django getting started tutorials.

When I make my query with graphiql:

```graphql
{
  allBookings {
    people {
      id
    }
  }
}
```

I get the following error:

```json
{
  "errors": [
    {
      "message": "Expected value of type \"BookingType\" but got: Person.",
      "locations": [
        {
          "line": 3,
          "column": 5
        }
      ]
    }
  ],
  "data": {
    "allBookings": [
      {
        "people": [
          null
        ]
      }
    ]
  }
}
```

Any ideas what i'm doing wrong? all my other foreign key (one-to-many) relations work fine out of the box. Also; if I query for a person's bookings then it works fine too. Just when I query for a booking's people do I get the error.

### LINKED ISSUE #157: Docs for graphene-django could use development on handling m2m fields
This could relate to [issue 155](https://github.com/graphql-python/graphene-django/issues/155).

### LINKED ISSUE #172: M2M connection is broken
Hi,

I have following models:

```
class Place(models.Model):
    categories = models.ManyToManyField(PlaceCategory, related_name='places')
```
```
class PlaceCategory(models.Model):
    name = models.CharField(max_length=255, unique=True)
```
Furthermore, I have following DjangoObjectType models:

```
class PlaceNode(DjangoObjectType):
    class Meta:
        model = Place
        interfaces = (graphene.relay.Node, )
```
```
class PlaceCategoryNode(DjangoObjectType):
    class Meta:
        model = PlaceCategory
        interfaces = (graphene.relay.Node, )
```

Then when I execute this query:
```
{
  placeCategories {
    edges {
      node {
        id
        places {
          edges {
            node {
              id
            }
          }
        }
      }
    }
  }
}
```

everything works great but when I try to execute this one:
```
{
  places {
    edges {
      node {
        id
        categories {
          edges {
            node {
              id
            }
          }
        }
      }
    }
  }
}
```
I get errors: 
> Cannot combine queries on two different base models.

It seems that M2M connection is working improperly and it is unable to fetch related objects from instance where M2M is defined.

Has anyone run into the same problem ?

spockNinja:
Issues
closes #155 
closes #157 
closes #172

### The Problem

ManyToMany fields were not mapping correctly and required some less than ideal workarounds.

This was because the `get_reverse_fields` did not differentiate between the original instantiation of the `ManyToMany` and the automatic `reverse` version. There was then two copies of the field coming back from `get_model_fields`, the one at the end being incorrect.

### The Solution

Unfortunately, the `reverse` flag on the "reverse" fields has not always been around. To maintain backwards compatibility, `get_reverse_fields` is doing the best it can.

Instead, we can make sure that local fields are not overridden by reverse fields. So a few lines in `get_model_fields` limits `get_reverse_fields` to skip local fields.

### Tests
Before fix:
screen shot 2017-05-25 at 11 25 09 am

After fix:
screen shot 2017-05-25 at 11 26 01 am

### Example
Before fix:
screen shot 2017-05-25 at 11 34 12 am

After fix:
screen shot 2017-05-25 at 11 40 30 am

### DIFF
--- a/graphene_django/tests/test_utils.py
+++ b/graphene_django/tests/test_utils.py
@@ -1,19 +1,5 @@
-
-
-from ..utils import get_model_fields, get_reverse_fields
+from ..utils import get_model_fields
 from .models import Film, Reporter
-
-
-def test_get_reverse_fields_correct():
-    reporter_reverse_fields = get_reverse_fields(Reporter)
-    reporter_field_names = [field[0] for field in reporter_reverse_fields]
-    assert sorted(reporter_field_names) == [
-        'articles', 'films'
-    ]
-
-    film_reverse_fields = get_reverse_fields(Film)
-    film_field_names = [field[0] for field in film_reverse_fields]
-    assert film_field_names == ['details']
 
 
 def test_get_model_fields_no_duplication():
--- a/graphene_django/tests/test_utils.py
+++ b/graphene_django/tests/test_utils.py
@@ -7,7 +7,7 @@ from .models import Film, Reporter
 def test_get_reverse_fields_correct():
     reporter_reverse_fields = get_reverse_fields(Reporter)
     reporter_field_names = [field[0] for field in reporter_reverse_fields]
-    assert reporter_field_names == [
+    assert sorted(reporter_field_names) == [
         'articles', 'films'
     ]
 
--- a/graphene_django/tests/test_utils.py
+++ b/graphene_django/tests/test_utils.py
@@ -1,0 +1,26 @@
+
+
+from ..utils import get_model_fields, get_reverse_fields
+from .models import Film, Reporter
+
+
+def test_get_reverse_fields_correct():
+    reporter_reverse_fields = get_reverse_fields(Reporter)
+    reporter_field_names = [field[0] for field in reporter_reverse_fields]
+    assert reporter_field_names == [
+        'articles', 'films'
+    ]
+
+    film_reverse_fields = get_reverse_fields(Film)
+    film_field_names = [field[0] for field in film_reverse_fields]
+    assert film_field_names == ['details']
+
+
+def test_get_model_fields_no_duplication():
+    reporter_fields = get_model_fields(Reporter)
+    reporter_name_set = set([field[0] for field in reporter_fields])
+    assert len(reporter_fields) == len(reporter_name_set)
+
+    film_fields = get_model_fields(Film)
+    film_name_set = set([field[0] for field in film_fields])
+    assert len(film_fields) == len(film_name_set)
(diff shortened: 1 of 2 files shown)


spockNinja:
@syrusakbary 

It appears that Django 1.8 does not have the `reverse` flag. We could simply use `getattr` and this fix will then not work for <1.8. Also open to other suggestions now that the root of the problem is clear.

spockNinja:
Another option might be to add logic to `get_model_fields` that ignores any `reverse_fields` that are already in `all_fields`. I'll give that a shot and see if it is more friendly to Django 1.8.

spockNinja:
That did the trick locally, and is passing all the tests. I'll update the PR description with the new solution.

LegoStormtroopr:
a. @spockNinja you are amazing
b. I think Django <= 1.7 is officially deprecated, so I wouldn't worry about that.

Considering how bleeding edge GraphQL is, my recommendation would be if 1.8 doesn't have the fields needed the focus on 1.9 and above. LTS is a stability release, and one of the downsides to that is not having every new piece of tech. But I'm not a maintainer of this repo, so I'm not sure what weight that advice holds here.

spockNinja:
I have changed the solution to not use the "new" `reverse` flag. The solution now is to simply ignore "local" fields when building the list of reverse fields. Tests are passing and it's working for me. Any feedback would be appreciated.

spockNinja:
@syrusakbary I believe this PR is ready for review. I have tested it with our data, added automatic tests, and all tests are passing. Let me know if there is anything I need to improve to get this merged. Thanks! :smile:

syrusakbary:
Great PR, and great analysis on how to make it in a compatible way.
Merging :)

clodal:
In case anyone needs it now, I left a workaround for graphql m2m connections in #172

PR #194, opened 2017-06-05
Fix graphene 1.3 annotation bug

### LINKED ISSUE #173: graphene 1.3 annotation bug
There seems to be a bug with graphene == 1.3

I have a model that I annotate a field "total_rent"
I think graphene is stripping it away somehow

I get this exception ```Cannot resolve keyword 'total_rent' into field. Choices are ... ```

When reverting back to graphene == 1.2.1 the error goes away.
I wish I could provide a more in depth report but my deadlines are to much at the moment

dvdmgl:
Fixes #173

### DIFF
--- a/graphene_django/fields.py
+++ b/graphene_django/fields.py
@@ -57,7 +57,7 @@ class DjangoConnectionField(ConnectionField):
 
     @classmethod
     def merge_querysets(cls, default_queryset, queryset):
-        return default_queryset & queryset
+        return queryset & default_queryset
 
     @classmethod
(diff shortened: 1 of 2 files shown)


crucialfelix:
Thanks for the fix.  What is in default_query_set that it should always be merged with the query_set that we supply ? It seems to me that I supply a query_set then it should be used as is. Merging with default_query_set seems dangerous.

jole78:
is this fix released on PyPI??

crucialfelix:
I don't think so

dleitee:
Why is it not o PyPi?

PR #202, opened 2017-06-23
Remove support to django 1.6 and 1.7

### LINKED ISSUE #187: Remove support for Django 1.6 and 1.7
While working on #186 I discovered that this library is still supporting django 1.6 and 1.7.

I think we can remove support of it since even Django itself is not supporting those versions anymore.

What do you think?

patrick91:
Closes #187 and it is helpful for #186 :)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -50,10 +50,6 @@ matrix:
   fast_finish: true
   include:
   - python: '2.7'
-    env: TEST_TYPE=build DJANGO_VERSION=1.6
-  - python: '2.7'
-    env: TEST_TYPE=build DJANGO_VERSION=1.7
-  - python: '2.7'
     env: TEST_TYPE=build DJANGO_VERSION=1.8
   - python: '2.7'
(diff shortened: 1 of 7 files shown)


patrick91:
@syrusakbary can you have a look at this one as well? :)

syrusakbary:
I've analyzed the downloads for Django 1.6, 1.7 vs the newest ones in [Google BigQuery](https://bigquery.cloud.google.com) database for PyPI:

```sql
SELECT
  MONTH(timestamp) as month,
  file.version,
  COUNT(*) as total_downloads
FROM
  TABLE_DATE_RANGE(
    [the-psf:pypi.downloads],
    TIMESTAMP("20170401"),
    CURRENT_TIMESTAMP()
  )
WHERE
  file.project = 'django'
GROUP BY
 month,
 file.version
ORDER BY
file.version,
month
LIMIT 1000
```

And the downloads for Django 1.6, Django 1.7 are now considerably low for the latest month compared to the most recent versions.

I think is reasonable to deprecate support for it (will add a small note in the README)

Merging :)

Issue #67, opened 2016-11-24
Choice fields automatically converted to enums


kaglowka:
```python
class User(DjangoObjectType):
    class Meta:
        model = get_user_model()
        only_fields = ['signup_site_version']

class Query(graphene.ObjectType):
    user = graphene.Field(User)

    def resolve_user(self, args, context, info):
        return user_from_info(info)  # return django user from context
```

### User model:
```python
@python_2_unicode_compatible
class User(AbstractUser):
    (...)
    signup_site_version_choices = [(1, u'site 1'), (2, u'site 2')]
    signup_site_version = models.IntegerField(choices=signup_site_version_choices)
```

### Query
```
query { 
  user {
    signupSiteVersion
  }
}
```

### Result
```json
{
  "data": {
    "user": {
      "signupSiteVersion": "A_1"
    }
  }
}

```

A copy-paste from where the weird value is born:
```python
def convert_choice_name(name):
    name = to_const(force_text(name))
    try:
        assert_valid_name(name)
    except AssertionError:
        name = "A_%s" % name
    return name
```

We've had similar problems in the old graphene (< 1.0): https://github.com/graphql-python/graphene/issues/134
The output is "A_1", because the field converter assumes django choices are tuples (id, name) where name is a constant name (not containing spaces etc.) or is convertible to a valid constant name. For some reason it is not successful while doing that for "site 1", but that is not an issue. The problem is it assumes we would like to do that.
In our django application (I suppose it is the norm) we use the second tuple element as a human-readable label. It means that in most cases replacing space with an underscore will do the work, but
in the other names are still bound to be long and there is no character we can safely assume the name won't contain (including hyphens, all native character varieties etc.).
The Enum fields are certainly nice, as their values are schema-validated, to say the least, but I doubt automatic conversion should be the only possible behaviour.

mjtamlyn:
My personal opinion is that it should be easy to make choices an Enum, but that it shouldn't be the default. The requirements of django choices are much, much more permissive than those of an Enum (in either Python or GraphQL)

nlhkh:
I also ran into this but I thought it were intensional.

pity7736:
Hi,

How I can overwrite the "A_1" output?

Thanks

BrianChapman:
Choices are also converted to uppercase, which makes it difficult to manage during mutations, especially if the choices were mixed upper and lower case. There should be an option to just leave your choices alone.

denis-ryzhkov:
[...]

PR #209, opened 2017-07-07
Support formatting of enum choices

frensjan:
When using [django_enumfields](https://github.com/hzdg/django-enumfields) ```field.choices``` is a list of the enum 'constants' and their labels / names. ```convert_choice_name``` renders the 'values' with str. This ok for normal choice values, but the string representation of python ```Enum```s follows the TypeName.VALUE format. This results in TYPENAME_VALUE formatted values for GraphQL queries, instead of VALUE (which I think it should be). This PR fixes that.

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -1,3 +1,5 @@
+import enum
+
 from django.db import models
 from django.utils.encoding import force_text
 
@@ -17,6 +19,8 @@ singledispatch = import_single_dispatch()
 
 
 def convert_choice_name(name):
+    if isinstance(name, enum.Enum):
+        name = name.name
     name = to_const(force_text(name))
     try:
         assert_valid_name(name)


Thibaut-Fatus:
this would be nice to have, and it fixes https://github.com/graphql-python/graphene-django/issues/67

syrusakbary:
Like it! A test case for this scenario would be helpful for assuring that we don't break support in the future :)

frensjan:
I've joined a new company and am not working on GraphQL nor on python anymore. Should I close the PR?

mvanlonden:
While I think this would be nice to have, I'm worried about the dependence on django-enumfields. Comments like this lead me to worry https://github.com/hzdg/django-enumfields/issues/96

phalt:

GitRon:
I think the topic is still relevant though I'm not sure about the current PR.

I faced the problem that the default (and not changable) behavoir is returning really weird things from ChoiceFields. For example "A_1" instead of the 1 is would expect.

I monkey-patched the behavior so I could work with it but a configurable solution would be awesome.

PR #212, opened 2017-07-11
Inheritance support for GraphQLView (class attributes)

### LINKED ISSUE #211: Problem with creating CustomGraphQlView
Looks like it is impossible to create custom CBV inherited from GraphQLView 

_(Schema in example below is correct GraphQLSchema)_
**This example will not work**
```python 
# views.py
class CustomGraphQLView(GraphQLView):
    schema = Schema
    graphiql = True if settings.DEBUG else False

# urls.py
...CustomGraphQLView.as_view()...
```
For now the only way  to pass schema and graphiql attributes is using .as_view() method, that is definitely not how CBV should behave

**This one will work**
```python
# urls.py
...CustomGraphQLView.as_view(schema = Schema, graphiql = True if settings.DEBUG else False)...
```

After short investigation i problem was localized to this part of code:
https://github.com/graphql-python/graphene-django/blob/master/graphene_django/views.py#L67-L75

When using 1st example, in \_\__init\__\_ self.schema is defined (equals to Schema object) but local schema variable (attribute of \_\__init_\_\_ method) is None, so check on L69 is always is True and on L75  self.schema is being rewritten. In my case, graphene_settings.SCHEMA is not defined so self.schema after L75 is always None and check on [L84](https://github.com/graphql-python/graphene-django/blob/master/graphene_django/views.py#L84) always raises AssertionError

There is the same problem with attributes like ```graphiql``` that are being overwritten by \_\__init_\_\_.

**Possible fix:**
```self.schema = self.schema or schema  # L75```
```self.graphiql = self.graphiql or graphiql  # L81```
etc

python 3.6
graphene-django==1.3

### PR DESCRIPTION
**Pull request by maqquettex**
Proposal for solving #211 (closes #211)

schema, pretty, graphiql and batch attributes can be defined in style of CBV

```python
class MyGraphQLView(GraphQLView):
    graphiql = True
    schema = SomeSchema
    pretty = True
    batch = True  # Of course cannot define both batch and graphiql, just an example
```
There is also simple test for all attributes except batch (because all 4 implemented in the same style).

### DIFF
--- a/graphene_django/tests/test_views.py
+++ b/graphene_django/tests/test_views.py
@@ -386,6 +386,24 @@ def test_allows_post_with_get_operation_name(client):
     }
 
 
+@pytest.mark.urls('graphene_django.tests.urls_inherited')
+def test_inherited_class_with_attributes_works(client):
+    inherited_url = '/graphql/inherited/'
+    # Check schema and pretty attributes work
+    response = client.post(url_string(inherited_url, query='{test}'))
+    assert response.content.decode() == (
+        '{\n'
+        '  "data": {\n'
+        '    "test": "Hello World"\n'
+        '  }\n'
+        '}'
+    )
+
+    # Check graphiql works
+    response = client.get(url_string(inherited_url), HTTP_ACCEPT='text/html')
+    assert response.status_code == 200
+
+
 @pytest.mark.urls('graphene_django.tests.urls_pretty')
 def test_supports_pretty_printing(client):
     response = client.get(url_string(query='{test}'))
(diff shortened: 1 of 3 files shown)


spockNinja:
@syrusakbary This is a clean and simple change to improve the re-usability and inheritance of the GraphQL view. The test looks good. 👍 from me.

maqquettex:
@syrusakbary Tests are passing on Graphene 2.0 (tested locally). Should i push merge-commit to this pull request, make rebase or anything like that?

patrick91:
@maqquettex I think the best way is to do a rebase, so the history is cleaner :)

maqquettex:
@patrick91 Done!

PR #213, opened 2017-07-12
WIP Add option for including foreignkey IDs

spockNinja:
### Problem
By default, relations are only exposed as complex nested structures. So in order to get a related ID, you have to perform the join to the related field. Sometimes, only the id is needed at first, and the related object might be fetched by id later on an as-needed basis.

### Solution
Add a `include_foreignkey_ids` option in the `DjangoObjectTypeMeta` that tells graphene-django to add `graphene.ID` fields for foreign keys.

Utilize this new option in `construct_fields` to look for `ForeignKey` fields and add them along with their existing relational field.

### Current Behavior
```graphql
{
  myObjects {
    relatedObject {
      id
    }
  }
}
```

### Current Workaround
You can currently achieve the same behavior by simply adding the id attribute to your `DjangoObjectType` class explicitly.

```python
class MyObject(DjangoObjectType):
    foreign_id_one = graphene.ID()
    foreign_id_two = graphene.ID()

    class Meta:
        model = MyObjectModel
```

### Usage after this PR
```python
class MyObject(DjangoObjectType):

    class Meta:
        model = MyObjectModel
        include_foreignkey_ids = True
```

And then you can simply do:
```graphql
{
  myObjects {
    relatedObjectId
  }
}
```
And avoid making a join when you don't have to.

While the workaround isn't difficult, in a large project with lots of models and relations, it can be a lot to add. In our case, we want these IDs present on all models, so it makes sense to provide a more convenient approach.


spockNinja:
@syrusakbary 

Does this seem like a feature you'd like to include in this repo? If so, I'll add some tests and documentation. If not, I'll just implement it for our project's graphql wrappers and close this PR.

spockNinja:
I'm actually wondering if this would be nicer as a global setting? It's still a bit tedious to add that line to every DjangoObjectType Meta.

firaskafri:
@jkimbo @dopeboy @danpalmer 
What do you think?

phalt:
This is a cool feature but it feels like a use case that isn't general enough to add a whole setting for. Could this not just be done with resolver methods?

Issue #179, opened 2017-05-16
DjangoConnectionField discards select_related


crucialfelix:
graphene-django==1.3

https://github.com/graphql-python/graphene-django/blob/master/graphene_django/fields.py#L59

When you do this:

```python
# DjangoFilterConnection

    @classmethod
    def merge_querysets(cls, default_queryset, queryset):
        return default_queryset & queryset
```

you discard any `select_related` or `prefetch_related` or `annotate` that might have been applied to the query set.

When we write a `def resolve_things(self,...` method and return a customized QuerySet it is then merged into the default_queryset and this discards everything (I think) except the filtering.

```python

queryset = Deals.objects.select_related('client')

> /usr/local/lib/python2.7/dist-packages/graphene_django/fields.py(58)merge_querysets()
     56             return self.model._default_manager
     57
---> 58     @classmethod
     59     def merge_querysets(cls, default_queryset, queryset):
     60         return default_queryset & queryset

ipdb> queryset[0].client
[0.051] SELECT "clients_deal"."id",
       "clients_deal"."client_id",
       "clients_deal"."agent_id",
       "clients_deal"."type",
       "clients_deal"."open",
       "clients_deal"."stage",
       "clients_deal"."created_on",
       "clients_deal"."modified_on",
       "clients_deal"."client_last_active_on",
       "clients_client"."id",
       "clients_client"."contact_id",
       "clients_client"."email",
       "clients_client"."fname",
       "clients_client"."lname",
       "clients_client"."phone",
       "clients_client"."work",
       "clients_client"."verified",
       "clients_client"."suggested_listings",
       "clients_client"."referral_source_id",
       "clients_client"."created_on",
       "clients_client"."modified_on",
       "clients_client"."created_by_id"
FROM "clients_deal"
INNER JOIN "clients_client" ON ("clients_deal"."client_id" = "clients_client"."id")
WHERE ("clients_deal"."agent_id" = 769
       AND "clients_deal"."open" = true
       AND "clients_deal"."agent_id" = 769)
ORDER BY "clients_deal"."client_last_active_on" DESC
LIMIT 1

# no extra query required
<Client: BOb Hop>

# here is after merging using &
ipdb> merged = default_queryset & queryset
ipdb> merged[0].client
[0.006] SELECT "clients_deal"."id",
       "clients_deal"."client_id",
       "clients_deal"."agent_id",
       "clients_deal"."type",
       "clients_deal"."open",
       "clients_deal"."stage",
[...]

PR #224, opened 2017-07-21
Fix select_related with filtering

camd:
I must admit, I don't totally see the need for this union between the ``default_queryset`` and ``queryset`` but through experimentation, I have found that if the order of the union is reversed, it will preserve the ``select_related`` and ``prefetch_related``.  It has the same effect of finding the union of records returned.  The list is identical.

So I hope this change doesn't break something I'm not aware of.  :)

This is an attempt to fix https://github.com/graphql-python/graphene-django/issues/179

### DIFF
--- a/graphene_django/filter/fields.py
+++ b/graphene_django/filter/fields.py
@@ -43,8 +43,8 @@ class DjangoFilterConnectionField(DjangoConnectionField):
     def filtering_args(self):
         return get_filtering_args_from_filterset(self.filterset_class, self.node_type)
 
-    @staticmethod
-    def merge_querysets(default_queryset, queryset):
+    @classmethod
+    def merge_querysets(cls, default_queryset, queryset):
         # There could be the case where the default queryset (returned from the filterclass)
         # and the resolver queryset have some limits on it.
         # We only would be able to apply one of those, but not both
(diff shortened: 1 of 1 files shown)


spockNinja:
@syrusakbary Same fix as #194, just specific to filter fields. Looks fine to me.

jancespivo:
Hi, IMO this code violates DRY principle and should be refactorized. The DjangoFilterConnectionField is inherited from DjangoConnectionField where the code is ok (in classmethod merge_queryset). I agree with this fix (it's the same fix as mine in duplicate #272 ) but I think it is definitely not a good final solution.

camd:
Thanks for accepting the fix!  I'm not able to merge it, so I'll leave that to the project collaborators.  Thanks everyone.  :)

syrusakbary:
I agree with @jancespivo that the code needs to be refactored, but we can do it in a separate PR.

The change looks good, but I think before merging the PR should include tests ensuring that the issue is not repeated in future releases.

spockNinja:
7a110ab should address everyone's concerns. Using `super` to use `DjangoConnectionField`'s default implementation. Since the base class's implementation was tested in #194 then it is tested here.

Tests are failing, but it's unrelated. I'm looking into those failures in a separate PR.

spockNinja:
Now that the failing test in master is fixed, this PR is ✅.

@syrusakbary Ready for merge, and I can close several duplicate issues/PRs for this bug.

camd:
Thanks for the merge!  :)

urbandove:
There's an issue with this merge - if a class is subclassing `DjangoFilterConnectionField` and a queryset is returned from `resolve_schema_field` the `super(cls,cls)` refers back to `DjangoFilterConnectionField` and has an endless loop.

PR #228, opened 2017-07-24
Allow for custom instantiation of input serializer

abdulhaq-e:
@patrick91

### DIFF
--- a/docs/rest-framework.rst
+++ b/docs/rest-framework.rst
@@ -19,3 +19,29 @@ You can create a Mutation based on a serializer by using the
         class Meta:
             serializer_class = MySerializer
 
+
+Customizing the mutation
+--------
+
+By default, when the serializer is instantiated, only the input argument
+is passed. Sometimes, the Django `request` object is required in the serializer
+context. In fact, any sort of complicated serializer will probably require something
+like `request.user`. This can be performed by customizing the instantiation
+method as such:
+
+.. code:: python
+
+    from graphene_django.rest_framework.mutation import SerializerMutation
+
+    class MySecondAwesomeMutation(SerializerMutation):
+        
+        @classmethod
+        def instantiate_serializer(cls, instance, args, request, info):
+            
+            input = args.get('input')
+            
+            return cls._meta.serializer_class(data=dict(input),
+                                              context={'request': request})
+      
+        class Meta:
+            serializer_class = MySerializer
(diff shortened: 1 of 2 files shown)


patrick91:
Looks good @abdulhaq-e, I've add just a comment :) 
also let's wait for @syrusakbary feedback, since I'm not a maintainer on this project

patrick91:
also, can you add a test? :)

docs/rest-framework.rst
```diff
@@ -19,3 +19,29 @@ You can create a Mutation based on a serializer by using the
...
+method as such:
+
+.. code:: python
+
+    from graphene_django.rest_framework.mutation import SerializerMutation
+
+    class MySecondAwesomeMutation(SerializerMutation):
+        
+        @classmethod
+        def instantiate_serializer(cls, instance, args, request, info):
```
abdulhaq-e: I just had a quick look at drf source, it's indeed `get_serializer` everywhere. I'll make the change.

abdulhaq-e:
I added the tests.

@patrick91 While writing the tests, I may have discovered a bug; serializers with no fields cannot be defined. If you see the test file and look at `MyBasicSerializer`, the field `whatever` cannot be deleted or else the tests will fail with:

> AssertionError: MyBasicSerializerInput fields must be a mapping (dict / OrderedDict) with field names as keys or a function which returns such a mapping.

I know it's an edge case but it maybe something to think about.

graphene_django/rest_framework/tests/test_mutation.py
```diff
@@ -68,3 +69,47 @@ class Meta:
     model_input_type = model_input._type.of_type
     assert issubclass(model_input_type, InputObjectType)
     assert 'cool_name' in model_input_type._meta.fields
+    
+
+def test_custom_serializer(capfd):
```
patrick91: what's `capfd` here?

patrick91: oh, got it: https://docs.pytest.org/en/latest/capture.html#accessing-captured-output-from-a-test-function (TIL) :)

I think you can remove it, since it not used :)

patrick91:
@abdulhaq-e well spotted, not sure if serialiser without fields are a thing, do you know any use case for it?

graphene_django/rest_framework/tests/test_mutation.py
abdulhaq-e: Removed

abdulhaq-e:
> serialiser without fields are a thing, do you know any use case for it?

Their `create` method can be used to run something which may only use the serializer `context` for example. Even if it's an edge case, the error message is not accurate. I'll try to have a look.

firaskafri:
@abdulhaq-e still interested to work on this?

abdulhaq-e:
@firaskafri I have a lot in my hand right now so unfortunately I can't :(

phalt:

PR #232, opened 2017-07-28
Pass context object to FilterSet for request-baed filtering

### LINKED ISSUE #203: request not being passed to filterclass
given the following code:
```
class ManageQuery(AbstractType):
    group_requests = DjangoFilterConnectionField(RequestNode, filterset_class=RequestManageFilter)
```
and the filterset class:
```
class RequestManageFilter(FilterSet):
    # Do case-insensitive lookups on 'name'

    @property
    def qs(self):
        user = getattr(self.request, 'user', None)
        print('User: ', user)
        parent = super(RequestManageFilter, self).qs
        return parent # .filter(group=self.request.user.main_author)

    class Meta:
        model = Request
        exclude = set()
```

the instance property self.request is always None, because when it's instantiated, graphene-django does not pass the context as request, shown here in fields.py's connection_resolver

```
   @classmethod
    def connection_resolver(cls, resolver, connection, default_manager, max_limit,
                            enforce_first_or_last, filterset_class, filtering_args,
                            root, args, context, info):
        filter_kwargs = {k: v for k, v in args.items() if k in filtering_args}

        qs = filterset_class(
            data=filter_kwargs,
            queryset=default_manager.get_queryset(),
        ).qs
```

passing request=context would make it work:
```
        qs = filterset_class(
            data=filter_kwargs,
            queryset=default_manager.get_queryset(),
            request=context,
        ).qs
```

basilfx:
This PR adds support for passing the context object to the filter set instance (it's called a [request](http://django-filter.readthedocs.io/en/latest/guide/usage.html#request-based-filtering) object inside a `FilterSet` instance).

I've added an additional test and updated the documentation.

It should fix #203.

### DIFF
--- a/docs/filtering.rst
+++ b/docs/filtering.rst
@@ -126,3 +126,23 @@ create your own ``Filterset`` as follows:
         # We specify our custom AnimalFilter using the filterset_class param
         all_animals = DjangoFilterConnectionField(AnimalNode,
                                                   filterset_class=AnimalFilter)
+
+The context argument is passed on as the `request argument <http://django-filter.readthedocs.io/en/latest/guide/usage.html#request-based-filtering>`__
+in a ``django_filters.FilterSet`` instance. You can use this to customize your
+filters to be context-dependent. We could modify the ``AnimalFilter`` above to
+pre-filter animals owned by the authenticated user (set in ``context.user``).
+
+.. code:: python
+
+    class AnimalFilter(django_filters.FilterSet):
+        # Do case-insensitive lookups on 'name'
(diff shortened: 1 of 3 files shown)


basilfx:
Is there anything I can do to help getting this PR merged?

spockNinja:
@syrusakbary As the added docs show, this is supported by [django-filter](http://django-filter.readthedocs.io/en/develop/guide/usage.html?highlight=request#request-based-filtering). It is a simple addition that will improve the django-filter/DRF integration. The test covers the new functionality and everything looks good to me.

PR #236, opened 2017-08-07
Generate a universal wheel in addition to the source distribution

### LINKED ISSUE #219: Provide wheel distribution on PyPI
Hi!

Please could this package also by made available as a wheel on PyPI?
https://pypi.python.org/pypi/graphene-django/1.3

Wheels have the following advantages:
* Faster installation time
* Allows the wheel cache to work even when using the new pip hash-checking mode
* Allows tools to statically inspect the package metadata without having to execute untrusted code.
* ...and more - see: http://pythonwheels.com

This package is pure Python and supports both Python 2 and 3, so can be distributed as a "universal wheel":
https://packaging.python.org/tutorials/distributing-packages/#wheels

To save having to pass `--universal` to `setup.py` each time, the pref can be saved in `setup.cfg`:

```
[bdist_wheel]
universal=1
```

The Python packaging guide recommends using [Twine](https://github.com/pypa/twine) to upload packages:
https://packaging.python.org/tutorials/distributing-packages/#uploading-your-project-to-pypi

So with the `setup.cfg` as above, the steps for building/uploading a new release are then:

```bash
$ pip install -U twine setuptools wheel
$ rm -rf dist/ build/
$ python setup.py sdist bdist_wheel
$ twine upload dist/*
```

The PyPI package listing will then display two entries for the new release, one with type "Source" and the other "Python Wheel".

Many thanks :-)

edmorley:
See:
https://packaging.python.org/tutorials/distributing-packages/#universal-wheels
https://docs.travis-ci.com/user/deployment/pypi/#Uploading-different-distributions

Fixes #219.


edmorley:
Many thanks :-)

PR #247, opened 2017-08-15
fixed out of date links to django filtering docs

picturedots:


picturedots:
Well, it looks like https://github.com/graphql-python/graphene-django/pull/246 is the exact same fix.  😀   So merge 246 or 247, but not both.

syrusakbary:
Closing this in favor of merged #246.
Thanks for the fix anyway :)

PR #261, opened 2017-08-31
fix various SerializerMutation regressions

### LINKED ISSUE #259: Rest Framework SerializerMutation Error - NoneType is not callable
Hi,
I'm using the SerializerMutation class to attempt to create mutations based off of my existing Rest Framework serializers. By taking advantage of the ModelSerializer in DRF, I hoped to avoid having to create all the mutations from scratch with validation and whatnot.

However, the use of the SerializerMutation class isn't sufficiently explained in the docs. I did this
```
class ItemMutation(SerializerMutation):
    class Meta:
        serializer_class = ItemSerializer

class Mutation(AbstractType):
    #...other mutations...
    createItem = ItemMutation.Field()
```

from a simple DRF ModelSerializer

```
class ItemSerializer(serializers.ModelSerializer):
    class Meta:
        model = Item
        fields = (
            'id',
            'name',
            'description',
            'website',
        )
```

But Graphiql shows me the following error whenever I try to use the mutation
```
mutation newItem($data: ItemMutationInput!) {
  createItem(input: $data) {
    name,
    description
  }
}

{
  "data": {
    "name": "Blah",
    "description": "Blah blah blah",
    "website": "https://blah.blah.blah"
  }
}
```

```
{
  "errors": [
    {
      "message": "'NoneType' object is not callable",
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ]
    }
  ],
  "data": {
    "createItem": null
  }
}
```

Also, Rest Framework's serializers can often have multiple actions (mutations) like create, delete, update, etc and the use of SerializerMutation appears to just produce a single mutation - presumably the "create" action. How do I handle the other actions?

q3aiml:
Following the 2.x work SerializerMutation is no longer working for me. I think I found 3 issues:

* Support for input fields was removed in 72529b7 causing SerializerMutations to accept only clientMutationId as input
* _meta.serializer_class is no longer set causing mutate_and_get_payload to throw a NoneType not callable error
* On success SerializerMutation.errors is graphene.List rather than None, leading to an error because graphene.List is not a valid value for a list

Thanks for looking

Fixes #259

### DIFF
--- a/graphene_django/rest_framework/mutation.py
+++ b/graphene_django/rest_framework/mutation.py
@@ -55,6 +55,7 @@ class SerializerMutation(ClientIDMutation):
         output_fields = fields_for_serializer(serializer, only_fields, exclude_fields, is_input=False)
 
         _meta = SerializerMutationOptions(cls)
+        _meta.serializer_class = serializer_class
         _meta.fields = yank_fields_from_attrs(
             output_fields,
             _as=Field,
--- a/graphene_django/rest_framework/mutation.py
+++ b/graphene_django/rest_framework/mutation.py
@@ -84,4 +84,4 @@ class SerializerMutation(ClientIDMutation):
     @classmethod
     def perform_mutate(cls, serializer, info):
         obj = serializer.save()
-        return cls(**obj)
+        return cls(errors=None, **obj)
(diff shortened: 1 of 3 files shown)


q3aiml:
Travis failures are from master being broken which should be fixed by #260

syrusakbary:
Thanks for the fixes! Merging :)

PR #285, opened 2017-10-09
Pass context to serializer so authentication works

glamdisju:
Without passing the context when creating the serializer you can't access the user in validation. Simply add the context when creating the serializer class fixes the authentication.

### DIFF
--- a/graphene_django/rest_framework/mutation.py
+++ b/graphene_django/rest_framework/mutation.py
@@ -69,7 +69,7 @@ class SerializerMutation(ClientIDMutation):
 
     @classmethod
     def mutate_and_get_payload(cls, root, info, **input):
-        serializer = cls._meta.serializer_class(data=input, context=info.context)
+        serializer = cls._meta.serializer_class(data=input, context=getattr(info, 'context', None))
 
         if serializer.is_valid():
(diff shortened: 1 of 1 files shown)


glamdisju:
I just saw that maybe I'm duplicating issue #228 altough that PR is a more general way of instantiating serializers.

graphene_django/rest_framework/mutation.py
```diff
@@ -69,7 +69,7 @@ def __init_subclass_with_meta__(cls, serializer_class=None,
 
     @classmethod
     def mutate_and_get_payload(cls, root, info, **input):
-        serializer = cls._meta.serializer_class(data=input)
+        serializer = cls._meta.serializer_class(data=input, context=getattr(info, 'context', None))
```
spockNinja: I believe there is a chance this might break if someone has overridden their serializer class's `__init__` in such a way as to not accept `**kwargs`. Even then, if they don't pop the `context` kwarg off before the base DRF `Field` class is resolved, I think this could cause an error.

spockNinja:
Duplicate functionality with #228

spockNinja:
Duplicate functionality with #228

PR #290, opened 2017-10-11
Fix distinct bug

danyx23:
This adds a test case and fixes graphql-python/graphene-django/issues/289. Please review and let me know if anythin should change, I'm not an expert in either graphql nor django :)

### DIFF
--- a/graphene_django/fields.py
+++ b/graphene_django/fields.py
@@ -67,6 +67,10 @@ class DjangoConnectionField(ConnectionField):
 
     @classmethod
     def merge_querysets(cls, default_queryset, queryset):
+        if default_queryset.query.distinct and not queryset.query.distinct:
+            queryset = queryset.distinct()
+        elif queryset.query.distinct and not default_queryset.query.distinct:
+            default_queryset = default_queryset.distinct()
         return queryset & default_queryset
 
     @classmethod
@@ -77,10 +81,6 @@ class DjangoConnectionField(ConnectionField):
         if isinstance(iterable, QuerySet):
             if iterable is not default_manager:
                 default_queryset = maybe_queryset(default_manager)
-                if default_queryset.query.distinct and not iterable.query.distinct:
-                    iterable = iterable.distinct()
-                elif iterable.query.distinct and not default_queryset.query.distinct:
-                    default_queryset = default_queryset.distinct()
                 iterable = cls.merge_querysets(default_queryset, iterable)
             _len = iterable.count()
         else:
--- a/graphene_django/fields.py
+++ b/graphene_django/fields.py
@@ -77,6 +77,10 @@ class DjangoConnectionField(ConnectionField):
         if isinstance(iterable, QuerySet):
             if iterable is not default_manager:
                 default_queryset = maybe_queryset(default_manager)
+                if default_queryset.query.distinct and not iterable.query.distinct:
+                    iterable = iterable.distinct()
+                elif iterable.query.distinct and not default_queryset.query.distinct:
+                    default_queryset = default_queryset.distinct()
                 iterable = cls.merge_querysets(default_queryset, iterable)
             _len = iterable.count()
         else:
--- a/graphene_django/tests/test_query.py
+++ b/graphene_django/tests/test_query.py
@@ -478,7 +478,7 @@ def test_should_query_node_filtering_with_distinct_queryset():
         'films': {
             'edges': [{
                 'node': {
-                    'genre': 'ot'
+                    'genre': 'OT'
                 }
             }]
         }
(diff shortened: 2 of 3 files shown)


spockNinja:
@danyx23 

This problem should be resolved by #194 or #224, which handles more than just the `distinct` being lost.

If you can verify that those fixes solve your problem, we can close this PR and it's corresponding issue.

Thanks!

jussih:
@spockNinja #194 does not fix this issue. Distinct and non-distinct querysets cannot be combined, assertion in Django prevents this.

graphene_django/fields.py
```diff
@@ -90,6 +90,10 @@ def connection_resolver(cls, resolver, connection, default_manager, max_limit,
         if isinstance(iterable, QuerySet):
             if iterable is not default_manager:
                 default_queryset = maybe_queryset(default_manager)
+                if default_queryset.query.distinct and not iterable.query.distinct:
```
spockNinja: How about moving this logic into `merge_querysets`? Then any of the subclasses that use that will get this fix for free.

graphene_django/fields.py
```diff
@@ -90,6 +90,10 @@ def connection_resolver(cls, resolver, connection, default_manager, max_limit,
...
+                    iterable = iterable.distinct()
```
spockNinja: How does this behave when the _other_ distinct call had field args? (`queryset.distinct('field_1', 'field_2')`)

graphene_django/tests/test_query.py
```diff
@@ -366,6 +368,61 @@ class Query(graphene.ObjectType):
...
+                'node': {
+                    'genre': 'OT'
+                }
+            }]
+        }
+    }
+
+    result = schema.execute(query)
+    assert not result.errors
+    print(result.data)
```
spockNinja: Lets be sure to drop this `print` once the test is working.

graphene_django/fields.py
danyx23: Good idea, did that

graphene_django/fields.py
danyx23: I honestly have no idea. From my superficial understanding it should be fine :). Do I understand it correctly that this is only possible when using postgres backends?

graphene_django/tests/test_query.py
danyx23: Done

danyx23:
Updated the PR. The travis check seems to rebase this on top of master but my test was written against django-graphene 1.x (4 parameters to resolve functions). Is there a way to make travis test this against the 1.x branch and once everyone is fine with the PR we adapt it to 2.0 and merge an updated PR there? Or is the 1.x line no longer supported?

spockNinja:
Since this PR is merging into `master`, it will need to use the latest resolver signature. If you also want a PR to add this for an older version, you can target the PR against the tagged version instead of the master branch.

There also appears to be a merge conflict in `fields.py`.

Thanks a bunch for your work. Let's get those last things fixed up and I'll pass this on to the owner for a final review. 😄

patrys:
@danyx23 Are you still working on this? If not I'd be happy to pick it where you've left it and adapt it for the current master.

@jussih The default `qs` implementation returning `model.objects.all()` is an implementation detail of `django-filters` so unless you propose that we detect whether a queryset is filtered, I think there's no easy way to skip merging in the default case of unfiltered data.

danyx23:
Ah damn, I totally forgot about this. Yes, I'll rebase against master and send an updated PR!

danyx23:
Rebased against master, tests are green. Please let me know if you are happy for this to merge. (It's the first time I rebased a github PR, I should probably have created a new one to preserve the history of the old commits, sorry abou that)

danyx23:
Do you mind merging this or do you consider it blocked because the coverage decreased by 0.08%?

graphene_django/tests/test_query.py
patrick91: the print is still there :)

danyx23: Ah sorry, missed your comment - thanks, will fix now!

graphene_django/tests/test_query.py
```diff
@@ -431,6 +435,60 @@ class Query(graphene.ObjectType):
...
+
+        class Meta:
+            model = Film
+            interfaces = (Node, )
+            filter_fields = ('genre',)
+
+    class Query(graphene.ObjectType):
+        films = DjangoConnectionField(FilmType)
+
+        # def resolve_all_reporters_with_berlin_films(self, args, context, info):
```
patrick91: sorry last one :) Is this needed? 

I guess after resolving this @spockNinja or @syrusakbary can merge :)

danyx23: There might be a way to integrate it with the existing test suite but I didn't see a straightforward way to construct a query that would lead to one "half" of the resulting query being distinct and the other not. Doing it in this way keeps the addition localized to where it is needed. If you have a suggestion on how to integrate it with the existing model instead please let me know!

syrusakbary:
The PR looks good 👍 . Merging

PR #297, opened 2017-10-20
Fix tutorial-plain.rst

ChouUn:
Arguments are incorrect.


syrusakbary:
The docs are based on new new resolution API in Graphene 2.0: https://github.com/graphql-python/graphene/blob/master/UPGRADE-v2.0.md#simpler-resolvers

PR #298, opened 2017-10-24
Update documentation in tutorial to avoid positional argument error

harshadbhatia:


syrusakbary:
This improvement is not necessary with Graphene 2.0 (would be helpful for previous versions though).
So I will close the PR :)

PR #299, opened 2017-10-25
Use maybe_queryset in DjangoFilterConnectionField

furious-luke:
In our projects we've found a need to inherit from
DjangoFilterConnectionField in order to customize the queryset. When
passing the queryset into "connection_resolver", an error was
raised about "get_queryset" not existing. This is due to the
assumption that the "base_manager" is always a manager. Using
"maybe_queryset" resolves the issue.

If there's a better way of doing any of this please let me know.


furious-luke:
Ah, I didn't mean for the second commit to end up here. I'll close this and open a new one.

PR #308, opened 2017-10-30
Read CSRF token from DOM instead of cookie

### LINKED ISSUE #307: GraphQLView fails when using session CSRF
If `settings.CSRF_USE_SESSIONS == True` or using https://github.com/mozilla/django-session-csrf then Django rejects all AJAX POST requests made by the GraphQLView.

The GraphQLView reads the CSRF token from the cookie.

Perhaps it should be read from the DOM instead?

nealedj:
Fixes #307


syrusakbary:
Would be great to support both (start with cookie and fallback to the input) instead of using only one.

phalt:

dulmandakh:
LGTM. @nealedj could you please rebase or resolve the conflict.

Issue #304, opened 2017-10-28
Custom Connection in 2.0


BossGrand:
What is the proper way to set a ModelNode Connection in 2.0?

The connection gets created by DjangoObjectType class. But I dont see how to overrride it with my own Connection

BossGrand:
I have found the connection field on DjangoObjectType meta. How can I assign the attr without causing a circular import issue

```python

class MyConnection(Connection):
    count = graphene.Int()

    def resolve_count(self, info):
        return self.length

    class Meta:
        node = MyNode


class MyNode(DjangoObjectType):

    class Meta:
        interfaces = (Node, )
        model = MyModel

        connection = MyConnection


```

connection and node need to reference each other. But I get a not defined error trying to do it this way. I've also tried lambdas but that does not work either

BossGrand:
I propose that we allow connection to be a lamda function similar to how Fields handle circular dependencies

syrusakbary:
Sorry, just realized I closed this issue by mistake.

BossGrand:
this was solved by the pull request

helpse:
@BossGrand So what's the usage of connection_class on this scenario?

helpse:
Answered here:
https://stackoverflow.com/questions/52746119/using-djangofilterconnectionfield-with-custom-connection-in-graphene-django/52746690#52746690

PR #309, opened 2017-10-30
Allow DjangoObjectType to use an Abstract Connection Class

BossGrand:
referred to as connection_class, it will instantiate the connection from the provided class or default to ```graphene.Connection``` if not supplied

Solves #304


BossGrand:
closing inorder to use the correct branch #313

PR #314, opened 2017-11-03
Fix Bug causing DjangoFilterConnectionField to ignore annotations, .reverse() and

BossGrand:
other queryset modifications being applied in resolvers.

the actual fix is really small. 

In  ```DjangoFilterConnectionField.merge_querysets```  default_queryset was overriding queryset.

if you look at ```DjangoConnectionField.merge_querysets``` you can see it was doing this merge correctly. 

```python
@classmethod
    def merge_querysets(cls, default_queryset, queryset):
        return queryset & default_queryset 
```

So this pr simply changes DjangoFilterConnectionField merge to match DjangoConnectionField merge. Along with a couple of tests


BossGrand:
Closing because I realized I accidentally rolled in a different pull request into this

PR #315, opened 2017-11-03
Fix Bug causing DjangoFilterConnectionField to ignore annotations, .reverse()

### LINKED ISSUE #197: Is there any way to get select_related to work?
I can't seem to get ``select_related`` to be honored at all, in my experience.  Can anyone help me identify how it should work?

``Job`` has a foreign keys to ``Push`` and to ``FailureClassification`` models.
Here's my Graphene Schema:

    class PushGraph(DjangoObjectType):
        class Meta:
            model = Push
            filter_fields = ('revision', )
            interfaces = (graphene.relay.Node, )
    
        jobs = DjangoFilterConnectionField(JobGraph)
    
        def resolve_jobs(self, args, context, info):
            return (Job.objects
                .filter(push=self, **args)
                .select_related("failure_classification")
            )

This seems to have no effect on doing a join in the queries.  I get tons of repeat queries for each ``Job`` that's returned.  Is this a bug?  Or a result of using ``DjangoFilterConnectionField``? as mentioned in Issue 179?

Thanks!

### LINKED ISSUE #238: DjangoFilterConnectionField ignores .using()
class ArticuloNode(DjangoObjectType):
    descripcion = graphene.String(source='descripcion')
    class Meta:
        model = Articulo
        filter_fields = {'nombre':['icontains', 'iexact'],
                        }
        interfaces = (graphene.Node, )

    articulos = DjangoFilterConnectionField(ArticuloNode)

def resolve_articulos(self, args, context, info):
    **print(Articulo.objects.using("db").all())
    x=Articulo.objects.using("db").all()    <----- Output printer it's ok, but the def returns ignoring "db" and shows "default" db result. 
    return x**

### LINKED ISSUE #244: DjangoFilterConnectionField with coalescence in a resolver
Hi,

Stumbled upon the following case:

1) DjangoFilterConnectionField won't allow me to use Coalese in a resolver's queryset throwing "Cannot resolve keyword u'display_or_updated_date' into field."
Using DjangoConnectionField instead helps.

```python
    qs = qs.annotate(display_or_updated_date=Coalesce('display_date', 'updated_at'))
```

2) At the same time, with `DjangoConnectionField` I cannot apply a slice (aka page size) to my query_set, getting "Cannot combine queries once a slice has been taken." + implement filtering, so would be great to find a solution for coalescence.

Any thoughts are appreciated!
Thank you!

BossGrand:
other queryset modifications being applied in resolvers.

the actual fix is really small. 

In  ```DjangoFilterConnectionField.merge_querysets```  default_queryset was overriding queryset.

if you look at ```DjangoConnectionField.merge_querysets``` you can see it was doing this merge correctly. 

```python
@classmethod
    def merge_querysets(cls, default_queryset, queryset):
        return queryset & default_queryset 
```

So this pr simply changes DjangoFilterConnectionField merge to match DjangoConnectionField merge. Along with a couple of tests

This should fix #197 fix #238 fix #244

### DIFF
--- a/graphene_django/filter/fields.py
+++ b/graphene_django/filter/fields.py
@@ -61,7 +61,7 @@ class DjangoFilterConnectionField(DjangoConnectionField):
         low = default_queryset.query.low_mark or queryset.query.low_mark
         high = default_queryset.query.high_mark or queryset.query.high_mark
         default_queryset.query.clear_limits()
-        queryset = default_queryset & queryset
+        queryset = queryset & default_queryset
         queryset.query.set_limits(low, high)
         return queryset
 
--- a/graphene_django/filter/tests/test_fields.py
+++ b/graphene_django/filter/tests/test_fields.py
@@ -2,13 +2,17 @@ from datetime import datetime
 
 import pytest
 
-from graphene import Field, ObjectType, Schema, Argument, Float, Boolean
+from graphene import Field, ObjectType, Schema, Argument, Float, Boolean, String
 from graphene.relay import Node
 from graphene_django import DjangoObjectType
 from graphene_django.forms import (GlobalIDFormField,
                                    GlobalIDMultipleChoiceField)
 from graphene_django.tests.models import Article, Pet, Reporter
 from graphene_django.utils import DJANGO_FILTER_INSTALLED
+
+# for annotation test
+from django.db.models import TextField, Value
+from django.db.models.functions import Concat
 
 pytestmark = []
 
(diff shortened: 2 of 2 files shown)


BossGrand:
The coverage looks really unrelated to what I was doing. Can someone look at it and help me figure out how to make it match what it was before

urbandove:
I just noticed that it seems this same pull request has been suggested multiple times.
The same change is in PR #166 (minus test) and it seems to me the same issue is being fixed with PR #224

PR #325, opened 2017-11-13
Fix infinite recursion when sublassing DjangoFilterConnectionField

urbandove:
Issue was caused by #224 
if a class is subclassing `DjangoFilterConnectionField` and a queryset is returned from `resolve_schema_field` the `super(cls,cls)` refers back to `DjangoFilterConnectionField` and has an infinite recursion.
I've subclassed `DjangoFilterConnectionField` to add orderby capabilities and it started erroring out because of infinite recursions
possibly a better fix would be
`queryset = DjangoFilterConnectionField.merge_querysets(default_queryset, queryset)`


urbandove:
build is failing pending the merger of #324

PR #326, opened 2017-11-15
DRF Serializer update

pizzapanther:
Currently you can't use serializers for updates only add operations unless you customize `mutate_and_get_payload`. This PR adds functionality to support most updates without any customization needed by the user. I also added support for Partial updates.

### DIFF
--- a/graphene_django/rest_framework/mutation.py
+++ b/graphene_django/rest_framework/mutation.py
@@ -1,5 +1,7 @@
 from collections import OrderedDict
 
+from django.shortcuts import get_object_or_404
+
 import graphene
 from graphene.types import Field, InputField
 from graphene.types.mutation import MutationOptions
@@ -15,6 +17,9 @@ from .types import ErrorType
 
 
 class SerializerMutationOptions(MutationOptions):
+    lookup_field = None
+    model_class = None
+    model_operations = ['create', 'update']
     serializer_class = None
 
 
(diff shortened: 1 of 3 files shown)


urbandove:
Looks great!

I see you decided to go with making the add/update decision based on if there was some sort of id specified. Much more django like. ;)

I'm wondering about raising the 404 on update if the instance isn't found - is GraphQL meant to be raising HTTP status codes? From https://github.com/graphql-python/graphene/issues/142 it seems not.
Wondering if there is any consensus in the GraphQL community on how such issues are meant to be dealt with.

pizzapanther:
I decided to go with the add/update and leave it up to user if they want to split out. The other way, the implementation would be very much the same but with more enforcement. So I went the flexible route.

As for the 404, that is something I am wishing for :-)

Graphene eats a lot of errors and returns 200 codes.  I see the reasoning for not with the link you posted but I still think that's probably bad behavior to continue executing after a failure. I'll update if needed but one can always dream.

graphene_django/rest_framework/mutation.py
```diff
@@ -67,9 +90,35 @@ def __init_subclass_with_meta__(cls, serializer_class=None,
         )
         super(SerializerMutation, cls).__init_subclass_with_meta__(_meta=_meta, input_fields=input_fields, **options)
 
+    @classmethod
+    def resolve_serializer_inputs(cls, root, info, **input):
```
spockNinja: What do you think of `get_serializer_kwargs` for this method? "resolve" and "inputs" both have special meanings in the graphene world.

graphene_django/rest_framework/mutation.py
```diff
@@ -67,9 +90,35 @@ def __init_subclass_with_meta__(cls, serializer_class=None,
...
+            else:
+                raise Exception(
+                    'Invalid update operation. Input parameter "{}" required.'.format(
+                        lookup_field
+                    ))
+
+            return {
+                'instance': instance,
+                'data': input,
+                'partial': cls._meta.partial
```
spockNinja: `instance` and `data` definitely make sense here, as those are very dynamic and dependent on the query. 

However, I think kwargs like `partial` (`many`, `context`, etc.) should use a different mechanism instead of living on `SerializerMutationOptions`. If we go with renaming this `get_serializer_kwargs`, those would be easy enough to add with a call to `super` on an overridden version.

There is also the compromise of adding a `serializer_kwargs` option to the `SerializerMutationOptions` that would be combined with this dict.

graphene_django/rest_framework/mutation.py
pizzapanther: I like it 👍

graphene_django/rest_framework/mutation.py
pizzapanther: Sounds good, I wasn't too happy with `partial` there

pizzapanther:
@spockNinja updated with feedback, also changed the operations to `'create', 'update'` instead of `'add', 'update'` so it mirrors DRF serializer language.

spockNinja:
It looks like the build is breaking due to an issue with py.test. I'll put in a separate PR to get that fixed up, and rebuild this one once it's merged.

pizzapanther:
@colanconnon added your lines but getting a weird test failure.  I'll see if i can debug, it seems unrelated

patrick91:
@pizzapanther if you rebase from master the CI check will pass :)

pizzapanther:
@patrick91 rebased

syrusakbary:
@patrick91 @spockNinja is this PR ready to be merged?

syrusakbary:
I re-reviewed and everything looks perfect. Sorry for taking so long to merge.

PR #334, opened 2017-12-03
Docs: Update Cookbook tutorial

fusion44:
Update the tutorial with a few fixes and add the models to the admin panel.


docs/tutorial-plain.rst
```diff
@@ -80,9 +81,21 @@ Add ingredients as INSTALLED_APPS:
     INSTALLED_APPS = [
         ...
         # Install the ingredients app
-        'ingredients',
+        'cookbook.ingredients',
     ]
 
+Register models with admin panel:
```
spockNinja: I think that since this part is optional to get to the graphene part of things, that we should move it down to the "Load some test data" section if they decide not to import the fixtures. https://github.com/graphql-python/graphene-django/pull/334/files#diff-6d6af2e0d0bec6b0abdc90fe84f077a5R125

PR #335, opened 2017-12-05
models.DateField => graphene Date Scalar

urbandove:
Currently models.DateField is processed as a DateTime Scalar - this changes it to the new Date Scalar added to graphene

Please note - that the test pass on Django 1.11 - but i was having erros crop up on django 2.0 (1 related to pytest dependancy and the rest was an issue of foreignkeys missing the rel attribute - which i dont think is caused by this)

### DIFF
--- a/examples/starwars/models.py
+++ b/examples/starwars/models.py
@@ -5,7 +5,7 @@ from django.db import models
 
 class Character(models.Model):
     name = models.CharField(max_length=50)
-    ship = models.ForeignKey('Ship', blank=True, null=True, related_name='characters')
+    ship = models.ForeignKey('Ship', on_delete=models.CASCADE, blank=True, null=True, related_name='characters')
 
     def __str__(self):
         return self.name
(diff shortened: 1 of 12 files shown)


urbandove:
Closing and reopening to get travis to run again

urbandove:
Hi @spockNinja 
Thanks for looking this over.
I've removed the code for the django 2 updates to keep this pull request cleaner - it was left in my branch from #336 by mistake
I've also added a version bump to 2.0.1 - same as the version of graphene which supports Date Scalars

spockNinja:
Thanks @urbandove!

@syrusakbary This update looks good to me. I'm not too sure how you want to handle version bumps when something breaks existing functionality (but is still a general improvement). @urbandove has bumped the version in this PR, but if that's not the version workflow you follow, we can undo that. In my review, I was just making a note of the *need* for a bump in general (before it goes to pypi), not necessarily in this PR.

PR #336, opened 2017-12-06
support Django 2 + in tests

urbandove:
Foreign keys in Django 2 require an on_delete attribute
Also the field.rel attribute has been deprecated as of Django 2 and is now field.remote_field
I have added a fallback for 1.8

[The django docs specifying the changes to field.rel](https://docs.djangoproject.com/en/1.9/releases/1.9/#field-rel-changes)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,6 +38,12 @@ env:
 matrix:
   fast_finish: true
   include:
+  - python: '3.4'
+    env: TEST_TYPE=build DJANGO_VERSION=2.0
+  - python: '3.5'
+    env: TEST_TYPE=build DJANGO_VERSION=2.0
+  - python: '3.6'
+    env: TEST_TYPE=build DJANGO_VERSION=2.0
   - python: '2.7'
     env: TEST_TYPE=build DJANGO_VERSION=1.8
   - python: '2.7'
--- a/examples/starwars/models.py
+++ b/examples/starwars/models.py
@@ -5,7 +5,7 @@ from django.db import models
 
 class Character(models.Model):
     name = models.CharField(max_length=50)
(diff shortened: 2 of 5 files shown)


urbandove:
I should note that in my previous pull request #335 i added on_delete attribute to all related fields - as this is a required field starting in 2.0 (but supported before)

urbandove:
Have updated python version in setup.py so that if its python 2 it doesnt try install django 2
this makes build pass - which hasnt been happening till now

urbandove:
Now the 2.0 tests are failing because pytest-django has not released a version which supports 2.0 (Coming in version 3.1.3)
(Also -the current pytest-django version is pinned at an old version anyways)
We can wait for the new version to be push or pip install directly from the github master branch

spockNinja:
Thanks for your work here! It's looking great so far.

It looks like pytest-django 3.1.3 is coming pretty soon, so my vote would be to wait for that and then we can update the `test_requires` to unpin/update it.

If you want to try out their master in the meantime to make sure it doesn't break anything else, that might be a good next move. There's always a chance that updating to their master will cause older python/django versions to break.

spockNinja:
For now, I'm going to steal this bit so we can at least get new PRs passing their builds.
```
django_version = 'Django>=1.8.0,<2' if sys.version_info[0] < 3 else 'Django>=1.8.0'
```

spockNinja:
I cherry-picked your 489d287 commit into #338 so we can get other builds passing again.

urbandove:
Hi @spockNinja 
pytest-django finally updated their pypi version to 3.2.1 which supports django2. So now this patch is ready to go.
There is the field.remote field change from an if-else to a try-catch - which I think is the accepted way for dealing with deprecated fields.

PR #338, opened 2017-12-09
Fixing the test build

spockNinja:
Several PRs are failing for unrelated reasons. Figuring out those reasons and getting the build passing again...

It looks like there is just one issue:

Django 2.0 is released and we aren't ready for that yet. #336 will fix it completely, but for now, we just need to not install 2.0 on python 2.7

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,6 +24,7 @@ tests_require = [
     'pytest-django==2.9.1',
 ] + rest_framework_require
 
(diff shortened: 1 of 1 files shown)


spockNinja:
@syrusakbary Our test suites for 2.7 are failing across several outstanding PRs.

#336 is going to get us Django 2.0 supported in full, but it's waiting on an upstream push to pytest. For now, we can get existing PRs fixed back up with just this change.

PR #339, opened 2017-12-10
Remove Django1.7 shims

urbandove:
Currently this library says its only compatible for django > 1.8 - this is just removing some shims that were there for earlier version


spockNinja:
Thanks for the cleanup!

We've got an outstanding issue with the Travis build, waiting on #338 so we can get your PR a clean build.

After that, this looks fine to me, and I'll pass it on to the owner. 👍

PR #340, opened 2017-12-12
Fix qfactor rankings for HTTP-ACCEPT

urbandove:
Currently qfactor is being parsed - and they are ordered in order of preference - but when i comes time to check if to show html vs json it just checks if either of them exist and then show html if html exists in the list of accepted headers and json doesn't - instead of figuring out what's the first in the list.
Also - when splitting multiple accepted headers it leaves a space in front of the second one and on - so I've added a ```.strip()```

Also added tests

### DIFF
--- a/graphene_django/views.py
+++ b/graphene_django/views.py
@@ -281,8 +281,8 @@ class GraphQLView(View):
     def request_wants_html(cls, request):
         accepted = get_accepted_content_types(request)
         accepted_length = len(accepted)
-        #the list will be ordered in preferred first - so we have to make
-        #sure the most preferred gets the highest number
+        # the list will be ordered in preferred first - so we have to make
+        # sure the most preferred gets the highest number
         html_priority = accepted_length - accepted.index('text/html') if 'text/html' in accepted else 0
         json_priority = accepted_length - accepted.index('application/json') if 'application/json' in accepted else 0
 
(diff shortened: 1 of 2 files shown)


graphene_django/views.py
```diff
@@ -280,8 +280,11 @@ def can_display_graphiql(cls, request, data):
     @classmethod
     def request_wants_html(cls, request):
         accepted = get_accepted_content_types(request)
-        html_index = accepted.count('text/html')
-        json_index = accepted.count('application/json')
+        accepted_length = len(accepted)
+        #the list will be ordered in preferred first - so we have to make
+        #sure the most preferred gets the highest number
+        html_index = accepted_length - accepted.index('text/html') if 'text/html' in accepted else 0
```
spockNinja: Yeah, `count` is definitely not accounting for priority here. Thanks for the fix.

Can we rename `html_index` and `json_index` to `html_priority` and `json_priority`? I was a little confused at first because their index wouldn't need `accepted_length`. But I now see why we need the `accepted_length - accepted.index(...` part to make the `html_priority > json_priority` condition work.

Thanks!

urbandove: Good Point - the names were leftovers from before the fixes - I have updated the  names

spockNinja:
@syrusakbary This PR looks good to me, and is ready for your approval/merge.

PR #356, opened 2017-12-28
Upgrade graphiql and react versions

jkimbo:
jsdelivr seems to have adopted a new url scheme. I also couldn't find a browser version of the `fetch` package so I switched to the `whatwg-fetch` package.


jkimbo:
No idea why the tests are failing...

jkimbo:
Tests are passing now

PR #365, opened 2018-01-05
updated path to graphene favicon in readme

mislavcimpersak:
In both README.md and README.rst path to graphene's favicon is a dead end.

I've linked the favicon straight from the main graphene-python.org repo and now it displays properly.
 
Before:
screen shot 2018-01-05 at 10 56 21

After:
screen shot 2018-01-05 at 10 56 08

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@ Please read [UPGRADE-v2.0.md](https://github.com/graphql-python/graphene/blob/ma
 
 ---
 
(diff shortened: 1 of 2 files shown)


spockNinja:
@syrusakbary The link is indeed broken. However, you might want to fix it for http://graphene-python.org/ anyway, which would fix this link.

mislavcimpersak:
I'm closing this pull request since there isn't even a simple logo image in the graphene-python.org repo anymore so I can't fix it even there by just linking to the image.

PR #368, opened 2018-01-08
Convert Date/Time/DateTime form fields to appropriate Graphene types

sciyoshi:


patrick91:
I've rebased from master and this is ready to be merged! /cc @syrusakbary

urbandove:
The graphene type Date was only added in graphene 2.0.1 - so before this gets merged it needs to make graphene-django have a requirement of graphene >= 2.0.1

In my pull request #335 I've updated that dependancy - so maybe the 2 pull request should be merged together?

Issue #319, opened 2017-11-06
Support for Proxy Models


jm2242:
Is there support for querying proxy models? I'm trying to query a model that has a bunch of proxy models but am getting an error. Here's an abbreviated look at my setup:
```
class Column:
...

class FieldColumn(Column):
  proxy=True
```

```
class Query(graphene.ObjecType):
    column = graphene.Field(ColumnType)
    all_columns = DjangoFilterConnectionField(ColumnType)
```

and 
```
class ColumnType(DjangoObjectType):
    class Meta:
        model = Column
        filter_fields = ['sheet__name', 'name']
        interfaces = (relay.Node, )
        exclude_fields = [  
            'legislation_type', 'project_stat_type', ... # weirdly have to exclude these fields because the 'type' in the variable name makes the field introspector sad
        ]
```

my query is:
```
query {
	allColumns(sheet_Name:"Simple Sheet") {
		edges {
      node{
        id, name
      }
    }
  }
}
```

Here's the trace:
```
Traceback (most recent call last):
  File "/Users/jonathanmares/.virtualenvs/quorum/lib/python2.7/site-packages/promise/promise.py", line 65, in try_catch
    return (handler(*args, **kwargs), None)
  File "/Users/jonathanmares/.virtualenvs/quorum/lib/python2.7/site-packages/graphql/execution/executor.py", line 375, in <lambda>
    resolved
  File "/Users/jonathanmares/.virtualenvs/quorum/lib/python2.7/site-packages/graphql/execution/executor.py", line 405, in complete_value
    return complete_object_value(exe_context, return_type, field_asts, info, result)
  File "/Users/jonathanmares/.virtualenvs/quorum/lib/python2.7/site-packages/graphql/execution/executor.py", line 499, in complete_object_value
    field_asts
graphql.error.base.GraphQLError: Expected value of type "ColumnType" but got: FieldColumn.
```

I know this PR https://github.com/graphql-python/graphene-django/pull/156/files is waiting to be merged, but I tried running my query with the proposed changes and still get the same error.

spockNinja:
Following that traceback, I think the root of the problem is [here](https://github.com/graphql-python/graphene-django/blob/master/graphene_django/types.py#L111-L112).

We are checking for `_meta.model` to match between the actual returned object and the model assigned in the DjangoObjectType meta.

[...]

PR #373, opened 2018-01-15
Basic Proxy model support

jm2242:
This PR, in reference to #319 :
-  adds support for querying a table that has proxy model representations in Django, as long as the type is declared to be the superclass (and not the proxy model class). While you could still query for all objects of a particular type, if those objects were cast to a proxy model at runtime, the `is_type_of()` for the object would fail.
~- adds `on_delete` where it was breaking tests due to Django 2 requirement~
- fixes tests associated with the Reporter model, to accommodate an additional field `reporter_type`

In the future, a more intentional way of dealing with Proxy models can be considered.

~There are 4 failing tests that are unrelated to this PR~

### DIFF
--- a/graphene_django/tests/models.py
+++ b/graphene_django/tests/models.py
@@ -38,7 +38,8 @@ class Reporter(models.Model):
         'Reporter Type',
         null=True,
         blank=True,
-        choices=[(1, u'Regular'), (2, u'CNN Reporter')] )
+        choices=[(1, u'Regular'), (2, u'CNN Reporter')]
+    )
 
     def __str__(self):              # __unicode__ on Python 2
         return "%s %s" % (self.first_name, self.last_name)
--- a/graphene_django/tests/models.py
+++ b/graphene_django/tests/models.py
@@ -15,7 +15,11 @@ class Pet(models.Model):
 
 class FilmDetails(models.Model):
     location = models.CharField(max_length=30)
-    film = models.OneToOneField('Film', related_name='details')
+    film = models.OneToOneField(
+        'Film',
+        on_delete=models.CASCADE,
+        related_name='details'
+    )
 
 
 class Film(models.Model):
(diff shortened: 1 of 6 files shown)


graphene_django/tests/models.py
```diff
@@ -30,15 +34,49 @@ class Reporter(models.Model):
     pets = models.ManyToManyField('self')
     a_choice = models.CharField(max_length=30, choices=CHOICES)
 
+    reporter_type = models.IntegerField(
+        'Reporter Type',
+        null=True,
+        blank=True,
+        choices=[(1, u'Regular'), (2, u'CNN Reporter')])
```
patrick91: nit: can you add a line break before the closing `)`? just to be consistent with the rest of the changes

graphene_django/tests/models.py
```diff
@@ -30,15 +34,49 @@ class Reporter(models.Model):
...
+        if self.reporter_type == 2:  # quick and dirty way without enums
+            self.__class__ = CNNReporter
+
+class CNNReporter(Reporter):
+    """
+    This class is a proxy model for Reporter, used for testing
+    proxy model support
+    """
+    class Meta:
+        proxy=True
```
patrick91: nit: can you add a space before and after the `=`?

jm2242:
@patrick91 rebased and made the style changes (hopefully I understood the first one)

patrick91:
@jm2242 I mean a line break, like this:

```python
...
    choices=[(1, u'Regular'), (2, u'CNN Reporter')]
)
```

jm2242:
that was silly, fixed

jm2242:
no problem @spockNinja, done!

mjtamlyn:
To be honest I know next to nothing about proxy models so probably can't help much here :)

syrusakbary:
The PR looks good, merging :)
Thanks for the reviews @spockNinja @patrick91  @mjtamlyn

serates:
Is graphene-django looking to release a new version and release this to pypi, the last build released was in october. I would like to have access to this fix so that i no longer have to use chain and lists for some of my queries that need proxy support

PR #388, opened 2018-01-31
Fix Django Rest Framework version when testing

patrick91:
Version 3.7 has removed support for anything less than 1.10, see here: http://www.django-rest-framework.org/topics/3.7-announcement/


patrick91:
@spockNinja I can't merge this directly, can you have a look at it? Should fix all the PR that are failing right now :)

PR #392, opened 2018-02-04
Convert GeometryField (fixes #390)

### LINKED ISSUE #390: Support for GIS field types
Right now, using a DjangoObjectType on a model that contains GIS fields (like `django.contrib.gis.db.models.fields.PointField`) will trigger an exception: `Exception: Don't know how to convert the Django field myapp.MyModel.point (<class 'django.contrib.gis.db.models.fields.PointField'>)`.

It would be great to add support for this. Otherwise, is there any docs on how to create a custom Field with custom serialization/deserialization like in django-rest-framework?

mongkok:
the PR convert Geometry fields to well-known text (WKT):

```json
{
    "id": "1",
    "location": "SRID=4326;POINT (0 1)"
}
```


zbyte64:
In my own project using graphql I prefer having GeometryField to not serialize as a string but a generic scalar because we're using geojson and the client doesn't bother (de)serializing json but gets direct objects.

Aside from that, we probably shouldn't be importing `django.contrib.gis` by default. 

Here is what I have for reference:
```
class GeoJSON(GenericScalar):
    @staticmethod
    def geos_to_json(value):
        return json.loads(GEOSGeometry(value).geojson)

    @staticmethod
    def json_to_geos(value):
        return GEOSGeometry(value)

    serialize = geos_to_json
    deserialize = json_to_geos


@convert_django_field.register(models.PolygonField)
@convert_django_field.register(models.MultiPolygonField)
def convert_geofield_to_string(field, registry=None):
     return GeoJSON(description=field.help_text, required=not field.null)


@convert_form_field.register(fields.PolygonField)
@convert_form_field.register(fields.MultiPolygonField)
def convert_geofield_to_string(field, registry=None):
     return GeoJSON(description=field.help_text, required=field.required)
```

jkimbo:
I don't know enough about GeometryField's in Django to be able to contribute here but I would say that (as @danpalmer mentions) generally creating customer scalar fields that are typed to match the underlying data model is a good thing. I'm not sure whether it is up to graphene-django to define what that type should look like this case but I don't see why not since GeometryField is part of Django.

PR #439, opened 2018-05-17
Allow graphql schema export to use a canonical representation

garyd203:
When we use the `graphql_schema` management command, the output can vary from run to run depending on arbitrary factors (because there is no guarantee made about the order used to output JSON dictionary keys). This makes it difficult to compare two schema's at different points in time.

We address this by including a new `canonical` flag to the command, which uses standard `json.dump` funcitonality to sort dictionary keys and force pretty-printed output.

### DIFF
--- a/docs/introspection.rst
+++ b/docs/introspection.rst
@@ -11,7 +11,7 @@ data to ``schema.json`` that is compatible with babel-relay-plugin.
 Usage
 -----
 
-Include ``graphene_django`` to ``INSTALLED_APPS`` in you project
+Include ``graphene_django`` to ``INSTALLED_APPS`` in your project
 settings:
 
 .. code:: python
@@ -28,6 +28,8 @@ the command:
 It dumps your full introspection schema to ``schema.json`` inside your
 project root directory. Point ``babel-relay-plugin`` to this file and
 you're ready to use Relay with Graphene GraphQL implementation.
+
+The schema file is sorted to create a reproducible canonical representation.
 
 Advanced Usage
 --------------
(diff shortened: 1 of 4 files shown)


firaskafri:
@garyd203 Interested to update this pull request?

garyd203:
Yes @firaskafri, hopefully in the next day or two

garyd203:
I have rebased this now @firaskafri

firaskafri:
Thank you @garyd203!

docs/introspection.rst
```diff
@@ -35,6 +35,10 @@ Advanced Usage
 The ``--indent`` option can be used to specify the number of indentation spaces to
 be used in the output. Defaults to `None` which displays all data on a single line.
 
+The ``--canonical`` option can be used to force the output to be sorted. This is
+useful if you want a reproducible canonical representation of your schema.
+Defaults to `False`, which displays keys in an arbitrary order.
```
danpalmer: Is there any need to make this an option? It feels like this should be the standard behaviour, I'm not sure I'd ever want the keys in an undefined order.

garyd203: Good question. My initial instinct was to maintain backward compatibility - if you change behaviour it should default to the original behaviour.

But like you say, it's hard to imagine a use case where someone would want arbitrary order and specifically doesn't want sorted order.

If we're all happy with that, I'll drop the command option and the documentation, and just add the field to the JSON export.

danpalmer:
Given that the current behaviour is essentially defined by dict ordering which until recently has been explicitly something that shouldn’t be depended on, I don’t think this is really breaking anything. 

I’m strongly in favour of remove this as it simplifies the code significantly. 

It may be worth keeping a brief note in the documentation to say that keys are sorted as it’s a nice thing to be able to rely on now!

jkimbo:
Completely agree with @danpalmer - this behaviour should be the default.

PR #454, opened 2018-06-13
Make examples diff better against each other

ericfrederich:
I was trying to see what the differences were between these examples.  There were a bunch of whitespace change and other changes which made no semantic difference.  This cleans them up so they diff better against each other.


ericfrederich:
After this patch the only differences are

- `ingredients/schema.py` and `recipes/schema.py` have changes
- cookbook has `django-filter` in `requirements.txt`
- cookbook has `dummy_data.json`

I used `meld` against the two directories and they are cleaned up now.

syrusakbary:
Looks good! Merging

PR #467, opened 2018-07-19
Support converters for custom fields with choices

danpalmer:
Currently if there's a field with choices, it cannot use a custom converter. We use `EnumField`s a lot, which have choices, but we want to customise how enums are exposed quite heavily, so would prefer custom converters to take precedence over the enum approximation based on the choices.

I introduced a new Exception class so that we can catch only the specific exceptions that we want to.

Open to review of the approach. I have only tested this on our codebase where we essentially won't use the choice-based enums at all, so I realise this might not be the best general approach.

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -38,13 +38,25 @@ def get_choices(choices):
             yield name, value, description
 
 
+class UnknownDjangoFieldException(Exception):
+    pass
+
+
 def convert_django_field_with_choices(field, registry=None):
     if registry is not None:
         converted = registry.get_converted_field(field)
         if converted:
             return converted
-    choices = getattr(field, 'choices', None)
-    if choices:
+
+    try:
+        converted = convert_django_field(field, registry)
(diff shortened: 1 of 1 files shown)


syrusakbary:
I like this PR. In order to avoid this issue in the future, would be great to have unit tests demonstrating the fix.
Once tests are added, it should be good to merge :)

danpalmer:
On further investigation, this change would break backwards compatibility – as things like a `CharField(choices=(...))` currently converts to an enum.

I'm not sure whether this is worth breaking backwards compatibility for. It's a shame that custom fields conversion can't be done for fields with choices, especially when those fields are provided by a library, but I'm not sure what the best option is here.

Do you have any thoughts? Should we make this breaking change?

zbyte64:
Another backward incompatible change to consider: Make enum conversions opt-in for non-charfields. #649 Is an example of using enums with IntegerField, resulting in non-obvious usage.

PR #477, opened 2018-07-25
Ensure code example is fully functional

whirish:
Simple change, it took me a while to figure out why the documentation's code was throwing an AssertionError :)


wsantos:
Are you sure? I'm using `Object` without a problem.

whirish:
I'm pretty sure it's tied to this, it was a fresh install on my end and everything. You're on the latest version and py3?

zbyte64:
I think having `Query` inherit `ObjectType` is proper usage.

jkimbo:
Thanks @whirish !

PR #482, opened 2018-07-30
one to one ModelSerializer mutations

srtucker22:
Hey there,

First of all, this project is fantastic!

I've been attempting to slowly apply `graphene-django` to some of the models/endpoints we use at work, and struggled with writing mutations over `ModelSerializer`s.

If you consider the Star Wars example, where every `Character` has a `Ship`, it would be pretty normal to write a `CharacterSerializer` like this:
```py
class CharacterSerializer(serializers.ModelSerializer):
    class Meta:
        model = Character
        fields = "__all__"
```
This will include a field `ship` of class `PrimaryKeyRelatedField`

To use this serializer to power a GQL mutation, you'd like to write something really clean like this:
```py
class CreateCharacter(SerializerMutation):
    class Meta:
        serializer_class = CharacterSerializer
```

Currently, this mutation will run, but the response won't match the GQL we expect:

```gql
# what we would expect
mutation createCharacter(name: String, ship: ID) {
    id
    name
    ship {
        id
        name
    }
    errors {
        messages	
    }
}

# what it creates
mutation createCharacter(name: String, ship: ID) {
    id
    name
    ship # a String :/
    errors {
        messages	
    }
}
```

You could roll your own mutation like in the star wars example:
https://github.com/graphql-python/graphene-django/blob/52d14f3b8ace2a3de0e3d07a597539390277ed2b/examples/starwars/schema.py#L41
but it would be great to support DRF `ModelSerializer` out of the box (we use DRF all over our codebase, often with complex validation)

This is a first crack at solving this problem. More work would be needed to support M2M relationships, but I wanted to get this in front of y'all first and get some feedback first.

Let me know what you think!

### DIFF
--- a/graphene_django/rest_framework/serializer_converter.py
+++ b/graphene_django/rest_framework/serializer_converter.py
@@ -58,7 +58,6 @@ def convert_serializer_field(field, is_input=True):
             field_model = field.Meta.model
             args = [global_registry.get_type_for_model(field_model)]
 
-    print('graphql_type', graphql_type, args)
     return graphql_type(*args, **kwargs)
 
 
--- a/graphene_django/rest_framework/serializer_converter.py
+++ b/graphene_django/rest_framework/serializer_converter.py
@@ -11,7 +11,7 @@ singledispatch = import_single_dispatch()
 
 
 @singledispatch
-def get_graphene_type_from_serializer_field(field):
+def get_graphene_type_from_serializer_field(field, **kwargs):
     raise ImproperlyConfigured(
         "Don't know how to convert the serializer field %s (%s) "
         "to Graphene type" % (field, field.__class__)
@@ -25,7 +25,7 @@ def convert_serializer_field(field, is_input=True):
     and the field itself is required
     """
 
-    graphql_type = get_graphene_type_from_serializer_field(field)
+    graphql_type = get_graphene_type_from_serializer_field(field, is_input=is_input)
 
     args = []
     kwargs = {"description": field.help_text, "required": is_input and field.required}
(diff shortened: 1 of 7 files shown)


danpalmer:
I'm afraid I don't have any experience with Django Rest Framework, so I'm going to opt-out of reviewing this one.

graphene_django/rest_framework/serializer_converter.py
```diff
@@ -72,49 +77,56 @@ def convert_serializer_to_input_type(serializer_class):
...
+@get_graphene_type_from_serializer_field.register(serializers.PrimaryKeyRelatedField)
+def convert_serializer_key_to_field(field, is_input=True):
+    if is_input:
+        return graphene.String
     return graphene.Field
 
 
 @get_graphene_type_from_serializer_field.register(serializers.ListSerializer)
-def convert_list_serializer_to_field(field):
+def convert_list_serializer_to_field(field, **kwargs):
```
charlex: Why start accepting `**kwargs` in these functions?

srtucker22: Hey @charlex, thanks for reviewing!

IIUC, for any `convert_x_serializer_to_y`, we run these converters on both inputs and resolvers. However, we need to treat inputs differently than resolvers for certain fields like `PrimaryKeyRelatedField`. In a mutation, an InputType for a `PrimaryKeyRelatedField` should be an `ID` (`String`), but should resolve the entire `Field` object.

In `convert_serializer_field`, we immediately check for the graphene type from the serializer field, but we need to know whether or not the field is an input for `PrimaryKeyRelatedField`. So now `convert_serializer_key_to_field` needs a second argument about whether we are converting for an input or a resolver. 

We could conditionally pass extra args into `get_graphene_type_from_serializer_field` e.g.:

```py
if isinstance(field, serializers.PrimaryKeyRelatedField):
  get_graphene_type_from_serializer_field(field, is_input=is_input)
else:
  get_graphene_type_from_serializer_field(field)
```
but, I thought it would make sense to let the other `convert_x_serializer_to_y` converters flexibly accept kwargs:

```py
get_graphene_type_from_serializer_field(
  field,
  is_input=is_input,
  other_arg=something_else
)

...

convert_x_serializer_to_y(field, **kwargs):
   # conversion logic here
```

If we're cool with this approach, I'd like to also apply this logic to 1:N mutations -- `PrimaryKeyRelatedField(many=True)`

Happy to adjust the code if there's a better approach!

PR #489, opened 2018-08-03
Add Introspection Schema Link in Documentation

CharlesBradshaw:
By the end of the Graphene and Django Tutorial using Relay, one might think they have finished everything needed server side for Relay, but six sections later, in a section that doesn't mention Relay in the title, the final required step for Relay is documented.


jkimbo:
Thanks @CharlesBradshaw !

PR #498, opened 2018-08-17
Always use HTTPS for CDN files

adamchainz:
There's no point using insecure, deprecated HTTP even if the current page is on HTTP.

### DIFF
--- a/graphene_django/templates/graphene/graphiql.html
+++ b/graphene_django/templates/graphene/graphiql.html
@@ -16,11 +16,11 @@ add "&raw" to the end of the URL within a browser.
       width: 100%;
     }
   </style>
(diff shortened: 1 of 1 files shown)


Amoki:
What about going one step further and adding the SHA hash of these js ?

adamchainz:
Forgot that's a thing now. I don't have the time to do it here. But it's not possible for two files as they are templated by version with `{{graphiql_version}}`

danpalmer:
@adamchainz I reckon it's worth doing for the other files, as it reduces the potential attack surface area.

adamchainz:
@danpalmer you're right

adamchainz:
Thanks!

PR #499, opened 2018-08-19
Fix code errors in form-mutations.rst

toolness:
This fixes what appear to be some code errors/typos:

* The `FormMutation` class was renamed to `DjangoFormMutation`, and `ModelFormMutation` to `DjangoModelFormMutation`, in 40610c64a3be003719d88db439e442085ed18072.
* `form_valid` was renamed to `perform_mutate` in 463ce68b16b070c0a49637dc71755a33acdf9d49.

It also clarifies a few things that I found confusing:

* It explicitly mentions that `perform_mutate` is a class method.
* The code samples now import the form classes from their packages, so readers know where to import them from too.


jkimbo:
Thanks for the improvements @toolness !

PR #506, opened 2018-08-29
Update authorization docs to Graphene 2.0

ccsv:
*  Re-write some language in "Limiting Field Access"
* Added code to "Queryset Filtering On Lists" section to handle queries that return nothing
* fix code to Filtering ID-based node access to work based on question [here](https://stackoverflow.com/questions/51057784/django-graphene-with-relay-restricting-queries-access-based-on-id/51958088#51958088)
* Rewrote Adding Login Requirements to be Django 2.0 compatible 
Fixed login requirements

### DIFF
--- a/docs/authorization.rst
+++ b/docs/authorization.rst
@@ -20,7 +20,7 @@ Let's use a simple example model.
 Limiting Field Access
 ---------------------
 
-This is easy, simply use the ``only_fields`` meta attribute.
+To limit fields in a GraphQL query simply use the ``only_fields`` meta attribute.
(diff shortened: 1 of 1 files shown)


japrogramer:
uhm, seeing as this is limiting the access after it has resolved is there a strategy to limit access not in the resolve but before the query is executed

ccsv:
@japrogramer  feel free to change the text I only change up some of the code because it bothered me that my GraphQL stuff did not work and asking around no one really resolved the issues.

PR #512, opened 2018-09-08
Provide a way to dump schema to stdout.

kalekseev:
It can be useful if you need `schema.graphql` instead of json eg:
`graphql_schema --schema project.schema --out=- | node bin/json-to-gql.js --src - --out project/schema.graphql`

### DIFF
--- a/graphene_django/management/commands/graphql_schema.py
+++ b/graphene_django/management/commands/graphql_schema.py
@@ -21,7 +21,7 @@ class CommandArguments(BaseCommand):
             type=str,
             dest="out",
             default=graphene_settings.SCHEMA_OUTPUT,
-            help="Output file (default: schema.json)",
+            help="Output file, --out=- prints to stdout (default: schema.json)",
         )
 
         parser.add_argument(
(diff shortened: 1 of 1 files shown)


andykenward:
@kalekseev where does the code in `json-to-gql.js` come from? Is it a custom script?

I need to convert the output to `.graphql`

kalekseev:
@andykenward 

```sh
django-admin graphql_schema --out=/tmp/schema.json
node json-to-gql.js --src /tmp/schema.json --out schema.graphql
```

json-to-gql.js
```js
const { buildClientSchema, printSchema } = require('graphql');
const fs = require('fs');

let src = null;
let out = null;

process.argv.forEach(function(val, index, array) {
  if (val === 'src') {
    src = array[index + 1];
  }
  switch (val) {
    case '--src':
      src = array[index + 1];
      break;
    case '--out':
      out = array[index + 1];
      break;
  }
});

function writeSchema(fileContent, output) {
  if (!fileContent) {
    console.error('no schema found');
    return;
  }
  const { data } = JSON.parse(fileContent);
  const clientSchema = buildClientSchema(data);
  const graphqlSchemaString = printSchema(clientSchema);
  fs.writeFileSync(output, graphqlSchemaString);
}

if (!src) {
  console.error('must provide src path');
} else if (!out) {
  console.error('must provide out path');
} else {
  if (src === '-') {
    let data = '';
    process.stdin
      .on('data', function(d) {
        data += d;
      })
      .on('end', function() {
        writeSchema(data, out);
      })
      .setEncoding('utf8');
  } else {
    const fileContent = fs.readFileSync(src, 'utf-8');
    writeSchema(fileContent, out);
  }
}

```

PR #513, opened 2018-09-09
Document Django Debug types

danpalmer:
This adds documentation for the debug fields/types in Graphene-Django.

Included in this is more strictness in whether some of the fields are required, as that is an important form of documentation in itself. These were determined from reading the source for `NormalCursorWrapper._record`, and some of the Django functions/methods that back that, but please do review as I might have missed something.


danpalmer:
@syrusakbary I think this is all ready to go. Hopefully an uncontroversial change.

PR #514, opened 2018-09-10
Apply camel case converter to field names in DRF errors

kalekseev:
This change eliminates inconsistency between schema field names and serializer errors field names

### DIFF
--- a/graphene_django/utils/utils.py
+++ b/graphene_django/utils/utils.py
@@ -2,7 +2,11 @@ import inspect
 
 from django.db import models
 from django.db.models.manager import Manager
+from django.utils import six
+from django.utils.encoding import force_text
+from django.utils.functional import Promise
 
+from graphene.utils.str_converters import to_camel_case
 
 try:
     import django_filters  # noqa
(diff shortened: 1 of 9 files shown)


kalekseev:
Though TypeError has type string[], I think nested serializers should be supported too? like if `errors = {entity: { option_a: ["This Field is required"] }}` the option_a should be converted to optionA, what do you think?

patrick91:
I quite like this, even thought I think I'll rewrite the implementation of the errors for the serializer mutation, but I'll probably keep the current implementation for backwards compatibility.


@syrusakbary what do think of this pull request? It breaks the compatibility with user using errors and having fields with _ in them, but it is definitely nicer than before I guess

artofhuman:
Should form errors convert to camel case too? https://github.com/graphql-python/graphene-django/blob/master/graphene_django/forms/mutation.py#L49

kalekseev:
> Should form errors convert to camel case too? https://github.com/graphql-python/graphene-django/blob/master/graphene_django/forms/mutation.py#L49

I can implement that in this pr but both are breaking changes so need decision from author first.

artofhuman:
Maybe for easy release we need another option in the schema. Like:

```
schema = Schema(Query, auto_camelcase=True, errors_camelcase=True)
```
Or if it related only for this packed use some global config like `DJANGO_GRAPHINE_CAMELCASE_ERRORS=True`

And check this options here with deprecation message. In major release check only `auto_camelcase`

And what if `auto_camelcase=False`, i think we should not use camelcase for errors. WDYT?

jkimbo:
This looks like it's still a useful change so don't close this stalebot!

@kalekseev so that we can get this merged without it being a breaking change could you add a flag like `DJANGO_GRAPHENE_CAMELCASE_ERRORS=True` that @artofhuman suggested?

kalekseev:
Hi guys, updated pr with recursive conversion to camel case for nested serializers support, added flag to enable this change.

artofhuman:
👍

graphene_django/utils/utils.py
```diff
@@ -12,6 +16,28 @@
...
+    return True
+
+
+def _camelize_django_str(s):
+    if isinstance(s, Promise):
+        s = force_text(s)
+    return to_camel_case(s) if isinstance(s, six.string_types) else s
+
+
+def camelize(data):
```
zbyte64: Is this function the same as `graphene.utils.str_converters.to_camel_case` ?
https://github.com/graphql-python/graphene/blob/ac6714e8fa61d3cb6d5f73b2614ce65b3cb13941/graphene/utils/str_converters.py#L6

zbyte64: Nvm, I see it now.

Issue #89, opened 2017-01-13
Add support to disable/enable reverse fields


fmartins:
I would suggest adding an option of disable/enable reverse fields in _graphene django_, for this, I did a quick draft that is available to analyses [1].

The motivation to this is that I don't believe that is safe to expose by default all reverse fields available in domain.

Please, could you express your point of view about this?

Thank you!

[1] https://github.com/fmartins-in-loggi/graphene-django-1.0

zbyte64:
I think pull request #435 touches on this

phalt:
Closing as would be resolved by #435

PR #519, opened 2018-09-20
feat(graphene_django): Add support to turn on/off reverse fields.

fmartins:
Hi, 

I would like receive a review with instructions about what I need to do to `merge` this changes in *master*. 

In the past, I created a issue that demonstrate the interest of another people to the subject - https://github.com/graphql-python/graphene-django/issues/89


fmartins:
This already was fixed here! https://github.com/graphql-python/graphene-django/commit/bb0b4fa76651ec0dd809bceadfb463596049458c

Issue #503, opened 2018-08-23
DjangoObjectType interprets string choices fields as required in schema even with blank=True attribute


picturedots:
To reproduce this problem, create a Django model with a choices field with `blank=True`:

    class MyModel(models.Model):
        PERIODIC_INTERVAL_CHOICES = (('Weekly', 'Weekly'),
             ('Bi-Weekly', 'Bi-Weekly'),
              ('Monthly', 'Monthly'),
              ('Quarterly', 'Quarterly'),
              ('Semi-Annually', 'Semi-Annually'),
              'Annually', 'Annually'))
        
        payment_frequency = models.CharField(
            blank=True,
            choices=PERIODIC_INTERVAL_CHOICES,
            max_length=13)

The schema generated by DjangoObjectType will show incorrectly show that this field `payment_frequency` is required.  This is incorrect -- the expected behavior is that it NOT required.

I have been able to fix this issue by patching how the value of required is determined in `graphene.converter.convert_django_field_with_choices` (shown for graphene-django 2.0.0)  :

    def convert_django_field_with_choices(field, registry=None):
        # Modified from graphene_django.converter import convert_django_field_with_choices
        # to adjust "required"
        choices = getattr(field, 'choices', None)
        if choices:
            meta = field.model._meta
            name = to_camel_case('{}_{}'.format(meta.object_name, field.name))
            choices = list(get_choices(choices))
            named_choices = [(c[0], c[1]) for c in choices]
            named_choices_descriptions = {c[0]: c[2] for c in choices}
    
            class EnumWithDescriptionsType(object):
    
                @property
                def description(self):
                    return named_choices_descriptions[self.name]
    
            enum = Enum(name, list(named_choices), type=EnumWithDescriptionsType)
            required = not (field.blank or field.null or field.default)  # MODIFIED FROM ORIGINAL
            return enum(description=field.help_text, required=required)
        return convert_django_field(field, registry)

eamigo86:
Hi @picturedots, I have tried your solution but still the error reported

picturedots:
The code above is for iillustrative purposes of where in the code the issue lies.  To actually use my code you need to replace every single one of your string enum fields, using a method like

    def create_enum_field(django_model, field_name):
        """ 
        Creates a graphene Enum field from a Django choices field that is
        suitable for inputs and outputs.
        """
        django_field = django_model._meta.get_field(field_name)
        graphene_field = convert_django_field_with_choices(django_field)
        return graphene_field

[...]

PR #524, opened 2018-09-25
Charfield choices with blank=True fix

legios89:
Hi,
I just found this problem during my work, and also found the solution the following issue: https://github.com/graphql-python/graphene-django/issues/503


legios89:
Sorry I still need to fix the tests

Issue #474, opened 2018-07-23
Django choice fields cannot be blank, even if field is not required -- graphene-django 2.1.0 / graphql-core 2.1 regression


picturedots:
This seems to be a regression in the upgrade of graphql-core to 2.1 together withj graphene-django to 2.1.0.

Prior to the upgrade, I could create an optional  choice field in Django, and could query the value if the value of the choice field was blank.  

For example, assume that I have a Django model `MyModel` with an optional choice field `my_choices_field`, and a query handler that would accept a query like

```
{
  myModel(id: 1) {
    id
    myChoicesField
  }
}
```

With the updated versions, if the value of `my_choices_field` is blank, this query throws  a handled exception like:

```
 Traceback (most recent call last):
   File "/usr/local/lib/python3.7/site-packages/promise/promise.py", line 65, in try_catch
     return (handler(*args, **kwargs), None)
   File "/usr/local/lib/python3.7/site-packages/graphql/execution/executor.py", line 527, in <lambda>
     exe_context, return_type, field_asts, info, path, resolved
   File "/usr/local/lib/python3.7/site-packages/graphql/execution/executor.py", line 556, in complete_value
     return complete_leaf_value(return_type, path, result)
   File "/usr/local/lib/python3.7/site-packages/graphql/execution/executor.py", line 622, in complete_leaf_value
     path=path,
 graphql.error.base.GraphQLError: Expected a value of type "MyModelMyChoicesField" but received:
```

And the output of the query includes errors when the value of the field is blank:

```
{
  "errors": [
    {
      "message": "Expected a value of type \"MyModelMyChoicesField\" but received: ",
      "path": [
        "myModel",
        "myChoicesField"
      ]
    }
  ],
  "data": {
    "myModel": {
      "id": "1",
      "myChoicesField": null
    }
  }
}

```

I imagine the issue is related to the work in graphql-core to related to enum fields 

> Improved GraphQLEnumType serialization allowing python Enum values 0cef6fb 24687b6

atomboulian:
I am experiencing the same thing - only started noticing it Monday morning when we got in - Friday it was working perfectly. Thank you for opening this @picturedots.

picturedots:
It seems like the real issue is with graphql-core 2.1 -- I've tried to track down the issue a bit and added an issue there.

atomboulian:
Thank you @picturedots. I'll chime in.

sciyoshi:
[...]

PR #525, opened 2018-09-25
Charfield choice with blank=True fix

legios89:
Hi,
I just found this problem during my work, and also found the solution in the following issue: #503
I also modified tests, what I forget to do in the previous pull request.

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -61,6 +61,14 @@ def convert_django_field_with_choices(field, registry=None):
     if choices:
         meta = field.model._meta
         name = to_camel_case("{}_{}".format(meta.object_name, field.name))
+        # Not null fields with choices and blank=True and without default value
+        if (
+            field.blank and
+            not field.null and
+            field.default == models.fields.NOT_PROVIDED
+        ):
+            choices = list(choices)
+            choices.append(('', ''))
         choices = list(get_choices(choices))
         named_choices = [(c[0], c[1]) for c in choices]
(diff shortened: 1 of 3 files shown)


legios89:
I am a little bit confused about what this error means? For me, it looks like everything went fine.

patrick91:
Hi @legios89 why did you close this? I think it is quite good, my only doubt would be that we are returning an empty string where it might make more sense to return None (just thinking from a frontend perspective).

Anyway, I've attached a test case that you can put in the PR, might be worth merging this and closing #474 :)

@spockNinja @syrusakbary do you have any additional feedback on this? 

```py
# https://github.com/graphql-python/graphene-django/issues/474

import graphene
from django.db import models

class Car(models.Model):
    class Meta:
        app_label = 'app'

    choice = models.CharField(
        choices=[('blue', 'Blue'), ('green', 'Green')],
        max_length=20,
        blank=True
    )

class CarNode(DjangoObjectType):
    class Meta:
        model = Car

def test_blank_choice():

        car = graphene.Field(CarNode)

        def resolve_car(self, info):
            return Car(choice="")

    result = schema.execute('{ car { choice } }')

    print(result.data, result.errors)

    assert not result.errors
```

zbyte64:
I too am confused as to why this is closed. This is still an issue and the PR seemed reasonable. The only thing I would change is the choice label for the null/empty value. With this PR it will generate "A_" for the enum label when it should likely be "BLANK" or "NULL"

I want to reopen this but only if the @legios89 agrees. Otherwise we need to make our own PR or merge with #654

PR #528, opened 2018-10-05
Get queryset

zbyte64:
My attempt at adding support for `DjangoObjectType.get_queryset` as described as Option 1 for #79 ; Tried to match up the method call as much as possible but dropped `request` and `args` as I wanted `get_node` to also be able to call `get_queryset` .

### DIFF
--- a/docs/authorization.rst
+++ b/docs/authorization.rst
@@ -96,6 +96,29 @@ schema is simple.
 
     result = schema.execute(query, context_value=request)
 
+
+Global Filtering
+----------------
+
+If you are using ``DjangoObjectType`` you can define a custom `get_queryset`.
+
+.. code:: python
+
+    from graphene import relay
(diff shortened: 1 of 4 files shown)


zbyte64:
After reviewing https://github.com/graphql-python/graphene-django/pull/160 I think this would be better served by a permission class. In the future a DjangoConnectionField could consult it's permissions for more specific permission actions like filtering objects on a connection.

maarcingebala:
This is something that I wanted to implement in my project and then I found this PR. There is a lot of cases when I'd like to be able to limit the queryset used underneath when resolving nodes so I think it's a must-have for `graphene-django`.

I took a look at the permissions PR you linked above and I think the solution you proposed is fine for now. Overriding a function that returns the queryset is a common pattern used in class-based views or DRF; it's also simple to use and explicit.

Issue #534, opened 2018-10-15
DjangoFilterConnectionField is silently bypassing FilterSet errors


helpse:
`DjangoFilterConnectionField` is not handling ValidationError from `FilterSet` at connection_resolver class method.

helpse:
@syrusakbary @spockNinja please review https://github.com/graphql-python/graphene-django/pull/535

PR #535, opened 2018-10-15
#534 add filterset validation to DjangoFilterConnectionField

helpse:


helpse:
@zbyte64 thanks! It's fixed now.

helpse:
@zbyte64 want to review it?

graphene_django/filter/fields.py
```diff
@@ -89,11 +90,22 @@ def connection_resolver(
...
-        ).qs
+        )
+
+        if not (filterset.is_bound and filterset.form.is_valid()):
+            exc = {
+                str(key): [str(e.message) for e in error_list]
+                for key, error_list in filterset.form.errors.as_data().items()
+            }
+
+            raise GraphQLError(exc)
```
danpalmer: I think we're passing a dictionary here, but [GraphQLError](https://github.com/graphql-python/graphql-core/blob/master/graphql/error/base.py#L26) appears to take a `str`. Should we format this into a string? I'm not sure that relying on the stringifying behaviour of `GraphQLError` is the right thing to do.

One alternative would be to wrap up the `dict -> str` processing in a `GraphQLDjangoFilterError` (better name needed). This would:

a) make the code here simpler
b) make it easier to customise the behaviour in the view the API that lets the Django view intercept errors.

graphene_django/filter/fields.py
```diff
@@ -89,11 +90,22 @@ def connection_resolver(
...
+        filterset = filterset_class(
             data=filter_kwargs,
             queryset=default_manager.get_queryset(),
             request=info.context,
-        ).qs
+        )
+
+        if not (filterset.is_bound and filterset.form.is_valid()):
+            exc = {
+                str(key): [str(e.message) for e in error_list]
```
danpalmer: The key here is in Python/Django source casing I believe. I think we could improve the error reporting here anyway (see other comment), but if we _do_ want to send these to the client, they should use the field names as in the API – i.e. probably `camelCase` names, although I believe this can be configured per schema, and overridden per field.

PR #537, opened 2018-10-18
convert DRF ChoiceField to Enum

### LINKED ISSUE #517: Reuse enum type of choices in serializer mutation
Having a field model with choices like the following:

```python
class Question(models.Model):
    TYPE_CHECKBOX = "checkbox"
    TYPE_INTEGER = "integer"

    TYPE_CHOICES = (
        TYPE_CHECKBOX,
        TYPE_INTEGER,
        TYPE_FLOAT,
        TYPE_RADIO,
        TYPE_TEXTAREA,
        TYPE_TEXT,
    )
    TYPE_CHOICES_TUPLE = ((type_choice, type_choice) for type_choice in TYPE_CHOICES)
    type = models.CharField(choices=TYPE_CHOICES_TUPLE, max_length=10)
```

would lead to enum type with uppercased options. Rest Framework `SerializerMutation` will expose this as `graphene.String` and expects value to be lower case.

Adding following serializer converter solves this problem, so a choices field will expose as correct enum type also in mutations:

```python
@serializer_converter.get_graphene_type_from_serializer_field.register(
    serializers.ChoiceField
)
def convert_serializer_field_to_enum(field):
    model_class = None
    serializer_meta = getattr(field.parent, "Meta", None)
    if serializer_meta:
        model_class = getattr(serializer_meta, "model", None)

    if model_class:
        registry = get_global_registry()
        model_field = model_class._meta.get_field(field.source)
        return type(convert_django_field_with_choices(model_field, registry))

    return graphene.String

```

Restriction is that when someone overwrites `ChoicesField` in serializer with a restricted set of options it will still use the same enum.

If the fix looks OK in general and there is a interest in this fix, I am happy to provide a PR. Let me know.

zbyte64:
This is to resolve #517 so that we use choices defined from DRF.

This also impacts `FilePathField` as it is technically a `ChoiceField`.


zbyte64:
Tests do pass, but can't directly trigger a rebuild. For some reason the CI broke (seems rarely intermittent):

```
$ curl -sSf -o python-3.5.tar.bz2 ${archive_url}
curl: (56) SSL read: error:00000000:lib(0):func(0):reason(0), errno 104
Unable to download 3.5 archive. The archive may not exist. Please consider a different version.
```

cowlicks:
I hit this issue today. Is there a workaround you would recommend that will allow me to have a schema where enums are used for Django choice fields when using the serializer mutation?

cowlicks:
Ideally, there would be a way to override the defaults for the inputs and outputs.

editkid:
@zbyte64 any chance this PR can be resurrected?

I'm using the workaround from https://github.com/graphql-python/graphene-django/issues/517. It would be great to have a built-in solution.

graphene_django/rest_framework/serializer_converter.py
```diff
@@ -138,3 +138,10 @@ def convert_serializer_field_to_jsonstring(field):
 @get_graphene_type_from_serializer_field.register(serializers.MultipleChoiceField)
 def convert_serializer_field_to_list_of_string(field):
     return (graphene.List, graphene.String)
```
zbyte64: If serializers.ChoiceField is an Enum,
serializers.MultipleChoiceField would logically become (graphene.List, Enum)

zbyte64:
@cowlicks you can always redefine the field where you define your own `DjangoObjectType`

zbyte64:
PR now converts multiple choices fields to list of enums.

Now the remaining question is whether it makes sense for FilePathField to be an enum. Choices may not be static as new directories could be added.

jkimbo:
@zbyte64 is this still a PR that you would like to see merged?

zbyte64:
This is still an issue and I still believe this is a valid way to address the issue. So yes, we should merge it...

jkimbo:
Merging this PR because it's been waiting around for ages.

overide:
@zbyte64 Can you direct me to the example where I see this change in action? I tried it but didn't work, maybe I didn't understand it correctly. I'm using v2.5.0

PR #538, opened 2018-10-18
Cookbook plain 2.1.2 compatible

alqinae:
Hello,

I am Khaled Alqenaei from DjangoCon sprint and I made the example compatible with the latest version of Django which is 2.1.2.


patrick91:
@syrusakbary @spockNinja this is ready to be squashed :) it's also @alqinae first contribution!

PR #539, opened 2018-10-18
Fix duplicated ErrorType declaration

### LINKED ISSUE #520: Duplicate defined ErrorType(s) causes an error
When using both form mutations, and DRF mutations there arises an error:
`AssertionError: Found different types with the same name in the schema: ErrorType, ErrorType.`

It seems we have two classes `ErrorType` defined by each integration.

patrick91:
Closes #520


anudeepsamaiya:
has this been released? because I am still getting this error.

PR #541, opened 2018-10-19
add setting to disable enum conversion

### LINKED ISSUE #67: Choice fields automatically converted to enums

zbyte64:
Needs unit test but not entirely sure about patching settings.

Resolves #67

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -24,6 +24,7 @@ from graphql import assert_valid_name
 from .compat import ArrayField, HStoreField, JSONField, RangeField
 from .fields import DjangoListField, DjangoConnectionField
 from .utils import import_single_dispatch
+from .settings import graphene_settings
 
 singledispatch = import_single_dispatch()
 
@@ -58,7 +59,7 @@ def convert_django_field_with_choices(field, registry=None):
         if converted:
             return converted
     choices = getattr(field, "choices", None)
-    if choices:
+    if choices and graphene_settings.GRAPHENE_CONVERT_CHOICES_TO_ENUMS:
         meta = field.model._meta
         name = to_camel_case("{}_{}".format(meta.object_name, field.name))
         choices = list(get_choices(choices))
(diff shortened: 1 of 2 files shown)


firaskafri:
@zbyte64 whats the alternative to enum to guarantee correct queries/mutations?

zbyte64:
Way I see it there are 3 options:

* Create one converter to rule them all. This is what we're doing right now, I don't think it will ever work.
* Option to disable enum globally. More like a stopgap solution, should be avoided
* Option to explicitly set the enum value. This is probably the right way to go. Check out #209

PR #546, opened 2018-10-28
[WIP] Improve form and serializer mutations

patrick91:
This PR improves both FormMutation and SerializerMutation to add support for better error types and wrapping the return type for the SerializerMutation.

I've also removed the ModelFormMutation since I think it is easier to use one class, and I'm not sure why it was possible to customize the model class there too.

# TODO:

- [ ] Use registry for return fields 
- [ ] Add support for return field to SerializerMutation
- [ ] Add support for adding more inputs or overriding them
- [ ] Add support for adding additional arguments
- [ ] Non opinionated errors[1]
- [ ] Update doc
- [ ] Allow to create Form Mutation for models that don't have graphql types
- [ ] more stuff (potentially)

Feel free to add comments and notes even if this is still a WIP.

[1] My idea is to add support for customizable errors via a setting property

NOTE: this is going to have some breaking changes (errors and return types have changed)


firaskafri:
@patrick91 any plans to update this pull request?

patrick91:
@firaskafri I'd love to, but I don't have much time right now, I might try to work on it during DjangoCon Europe sprints :)

firaskafri:
@CharlesBradshaw can you update @patrick91 todo list?

zbyte64:
My current attempt at supporting Global IDs as part of form mutation lookup does a hard type check:

```
@classmethod
    def get_form_kwargs(cls, root, info, **input):
        kwargs = {"data": input}

        pk = input.pop("id", None)
        if pk:
            pk_type = cls.Input._meta.fields['id'].type
            if isinstance(pk, str) and pk_type == ID:
                _, pk = from_global_id(pk)
            instance = cls._meta.model._default_manager.get(pk=pk)
            kwargs["instance"] = instance

        return kwargs
```

For an `InputField` I could not find a duck typed method that would parse global IDs. Alternatively we could try using the form field to parse the supplied value, I think that will be my next attempt.

zbyte64:
I think adding a "resolve_instance" class method to `BaseDjangoFormMutation` might be a good idea. Users may have their own custom lookup logic, ie UUID, and that would make it easier for them.

PR #563, opened 2018-12-18
 Support use `first` and `last` at same time as `offset`

NateScarlet:
Current implementation of max limit not correct when use `first` and `last` at same time
This operation is only valid when `graphene_settings.RELAY_CONNECTION_MAX_LIMIT` greater than 9999:
```gql
{
  users(first: 9999, last: 1) {
    edges {
      node {
        id
      }
    }
  }
}
```
Fixed it by using `count = min(i for i in (first, last) if i)`


zbyte64:
According to coveralls there are 3 new conditions not tested:

* negative `first`
* negative `last`
*  0 `max_limit`

We should have tests for those.

NateScarlet:
> 
> 
> According to coveralls there are 3 new conditions not tested:
> 
>     * negative `first`
> 
>     * negative `last`
> 
>     * 0 `max_limit`
> 
> 
> We should have tests for those.

Added

graphene_django/fields.py
```diff
@@ -103,39 +102,44 @@ def resolve_connection(cls, connection, default_manager, args, iterable):
...
+            cls,
+            resolver,
+            connection,
+            default_manager,
+            max_limit,
+            enforce_first_or_last,
+            root,
+            info,
+            **kwargs):
+        # pylint: disable=R0913,W0221
```
danpalmer: I don't believe we use pylint in this codebase. Could we remove this please?

graphene_django/fields.py
```diff
@@ -103,39 +102,44 @@ def resolve_connection(cls, connection, default_manager, args, iterable):
...
+
+        first = kwargs.get("first")
+        last = kwargs.get("last")
+        if not (first is None or first > 0):
+            raise ValueError(
+                "`first` argument must be positive, got `{first}`".format(**locals()))
```
danpalmer: The call to locals and constructing the function call to format is fairly expensive here I think. Could we instead pass the kwarg we need? (Same in the other instances in this function)

graphene_django/fields.py
```diff
@@ -103,39 +102,44 @@ def resolve_connection(cls, connection, default_manager, args, iterable):
...
+            max_limit,
+            enforce_first_or_last,
+            root,
+            info,
+            **kwargs):
+        # pylint: disable=R0913,W0221
+
+        first = kwargs.get("first")
+        last = kwargs.get("last")
+        if not (first is None or first > 0):
```
danpalmer: What do you think of wording this as `if first is not None and first < 1:` or even `if first and first < 1:`? I think it might be clearer, but I'm not sure.

graphene_django/fields.py
```diff
@@ -103,39 +102,44 @@ def resolve_connection(cls, connection, default_manager, args, iterable):
...
+            raise ValueError(
+                "You must provide a `first` or `last` value "
+                "to properly paginate the `{info.field_name}` connection.".format(**locals()))
+
+        if not max_limit:
+            pass
+        elif first is None and last is None:
+            kwargs['first'] = max_limit
+        else:
+            count = min(i for i in (first, last) if i)
```
danpalmer: Is it even valid to have a first and a last? I'm not sure I fully understand what behaviour providing `first` and `last` should result in. Could you give an example?

NateScarlet: By current `graphql_relay` implementation,
`last` means last N items in the selection,
and when use `last` with `first`,  `last` means last N items in first N items.

graphene_django/fields.py
NateScarlet: Changed to `if first is not None and first <= 0`.
0 is also a invalid value so `if first and first < 1` is not correct.

NateScarlet:
ping @danpalmer

NateScarlet:
I am closing this PR because my project has been rewritten in golang.

thomas-arun:
@NateScarlet, Sorry to bring this thread back to life. But would you help me with the reason why you re wrote your project in Go Lang? 
Did you use [graphql-go](https://github.com/graphql-go/graphql) ?

NateScarlet:
> @NateScarlet, Sorry to bring this thread back to life. But would you help me with the reason why you re wrote your project in Go Lang? Did you use [graphql-go](https://github.com/graphql-go/graphql) ?

I switch to golang because it has better performance on same hardware.

I am using [gqlgen](https://github.com/99designs/gqlgen)

PR #565, opened 2018-12-29
Improve Error Message if error object is not type of DjangoObjectType

bastiW:
See also: https://stackoverflow.com/a/49834443/2440049

If you inherit from a wrong type then the error you get an error which is confusing. 
Currently you get **...type object ‘User’ has no attribute ‘name’**
now you get ** [object_name] must inherit from DjangoObjectType. **

```
        if "has no attribute 'name'" in str(e):
            object_name = str(e).split("'")[1]

```
Those two lines I don't like so much but I didn't know how to get something like an error id. 

Actually the current confusing part of the error is raised by Django. Do you think it is a better Idea to create a pull request in the django repository?


firaskafri:
@bastiW This looks nice, can you please squash your commits?

bastiW:
Interesting that you merged it without any comment. Code I wrote looked quiete hacky to me. 

Do you think there is another way of solving it?

bastiW:
Is it already solved in a better way?

Issue #460, opened 2018-06-28
BaseDjangoFormMutation should use from_global_id when fetching existing instance


nuarhu:
We are using DjangoFormMutations (#217), and what I'm not sure about is how Django's ID/PKs are resolved in the process. They still appear in base64 encoding in the GraphQL output but in `BaseDjangoFormMutation` method they are processed as is.

The error returned from GraphQL shows that the instance is fetched using the string with the base64 encoded ID instead of the decoded PK integer value.

    "errors": [
      {
        "message": "invalid literal for int() with base 10: 'UHJvamVjdE5vZGU6MTA1'",

I think the following code from /graphene_django/forms/mutation.py should use `from_global_id` instead of the value as is:

    @classmethod
    def get_form_kwargs(cls, root, info, **input):
        kwargs = {'data': input}

        pk = input.pop('id', None)
        if pk:
            instance = cls._meta.model._default_manager.get(pk=pk)
            kwargs['instance'] = instance

        return kwargs

I'm not 100% sure about this. We are using UUIDs next to Django's IDs so I had to extend `get_form_kwargs` anyway, and it works fine that way. But it's not working for ID. If the above is the correct location for the fix I can create a pull request. Just wanted to check back if I'm understanding this correctly.

Murthy10:
I'm facing the same issue is there any progress?

nuarhu:
We developed our own layer instead of using graphene-django. Unfortunately, we are still stuck with the staticmethods and classmethods and init_with_subclass instead of a decent way of extending the base classes. It's a pain and performance is an issue.

Murthy10:
On the client side I'm using React and Apollo GraphQL to do the mutation.
After some fiddling around I have a hacky solution for my problem.
I decode the base 64 Id and use the global Id.

`let id = window.atob(id).split(':').pop();`

Maybe this could help someone until the issue is fixed.

alexnitta:
I found the basis of my solution in this comment: https://github.com/graphql-python/graphene/issues/243#issuecomment-245827012

It looks like the idea is to extend the `graphene.relay.node` class to change the default behavior of base-64 encoding the primary key.

Here's an example where I'm using a UUID for the primary key, but I expect this would work with an integer primary key as well.

```python
# models.py

import uuid

from django.db import models
from graphene.relay import node


class UUIDModel(models.Model):
    id = models.UUIDField(primary_key=True, default=uuid.uuid4, editable=False)

    
class Person(UUIDModel):
[...]

PR #568, opened 2019-01-05
Fix: BaseDjangoFormMutation should use from_global_id when fetching e…

r3stm4n:
…xisting instance #460


firaskafri:
@zbyte64 the tests unexpectedly seem to be failing though

zbyte64:
You're right...

I see that #546 attempts to do the same thing but includes updated tests,

PR #585, opened 2019-02-20
Make CSRF cookie and header names configuration based on django settings

BobReid:
**Issue:**

Django allows the the CSRF cookie and header to be configured through the use of settings variables. graphene-django's implementation if GraphIQL does not respect these settings which causes a CSRF failure. 

**Fix:**
This PR fixes the issue by reading the values from the django settings values and uses them in the graphiql template. 

**Repro:**
1. create graphene-django project
2. set django setting CSRF_COOKIE_NAME and CSRF_HEADER_NAME to non standard values
3. Launch graphiql

**Expected:** 
GraphIQL works as expected (request / response pass and scheme can be inspected)

**Actual:**
403 status' are returned due to a CSRF failure

### DIFF
--- a/graphene_django/templates/graphene/graphiql.html
+++ b/graphene_django/templates/graphene/graphiql.html
@@ -26,7 +26,7 @@ add "&raw" to the end of the URL within a browser.
   <script>
     // Parse the cookie value for a CSRF token
     var csrftoken;
-    var cookies = ('; ' + document.cookie).split('; csrftoken=');
+    var cookies = ('; ' + document.cookie).split('; {{ csrf_cookie }}=');
     if (cookies.length == 2)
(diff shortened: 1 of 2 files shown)


dulmandakh:
@BobReid hey, could you please rebase or resolve the conflicts. I would like to test and review your changes. Thank you

BobReid:
@dulmandakh 

I was going to rebase these changes but it looks like the GraphIQL javascript has been extracted from the html template and placed into a static file. This refactor is incompatible with my approach because I was relying on running the javascript through the templating engine in order to set the header name dynamically. 

I could stuff the header name in a global JS variable in the html template and read it in the GraphIQL script. It is ugly but would work. 

Thoughts?

PR #598, opened 2019-03-21
Add default value to generated enum if not included in choices

mvanlonden:
Closes https://github.com/graphql-python/graphene-django/issues/474

Django fields with `default="something"` may not be included in choices but the value is a valid initial value for the field. This adds an `DEFAULT` option to the enum generated if the default value is not present in the choices.

https://simpleisbetterthancomplex.com/tips/2016/07/25/django-tip-8-blank-or-null.html

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -78,3 +78,7 @@ Session.vim
 *~
 # auto-generated tag files
 tags
+
+# Pipenv
+Pipfile
+Pipfile.lock
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -62,6 +62,8 @@ def convert_django_field_with_choices(field, registry=None):
         meta = field.model._meta
         name = to_camel_case("{}_{}".format(meta.object_name, field.name))
         choices = list(get_choices(choices))
+        if field.blank:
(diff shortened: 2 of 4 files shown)


zbyte64:
I think a person should be able to define the label for the blank option. For django forms the pattern is that you explicitly define a blank choice or pass `empty_label`: https://stackoverflow.com/questions/14541074/empty-label-choicefield-django

Another concern is how this gets squared with the proposal of letting people explicitly define enums.

mvanlonden:
Ah didn't realize defining an empty option is more idiomatic. That makes sense

mvanlonden:
@zbyte64 @firaskafri I've taken an alternative approach by inspecting the default value for a field. Often times a CharField will define `""` as the default value and it is not included in the choices. This adds a `DEFAULT` value to the generated enum.

mvanlonden:
Yes @danpalmer after further investigation it looks like you're right. Previously wasn't able to find examples where an empty value was included in choices. https://github.com/django/django/blob/1123f4551158b7fc65d3bd88c375a4517dcd0720/tests/forms_tests/models.py#L37

PR #600, opened 2019-03-25
Add support for filterset_class meta parameter

sierreis:
This would enable filtering on Django relationships that create a default DjangoFilterConnectionField based on the model type and its meta parameters.

* Allow for use of either filter_fields or filterset_class
* Add tests to check that the behavior is similar to filter_fields
* Add documentation to show how to make use of the parameter

### DIFF
--- a/docs/filtering.rst
+++ b/docs/filtering.rst
@@ -100,7 +100,7 @@ features of ``django-filter``. This is done by transparently creating a
 ``filter_fields``.
 
 However, you may find this to be insufficient. In these cases you can
-create your own ``Filterset`` as follows:
+create your own ``FilterSet``. You can pass it directly as follows:
 
 .. code:: python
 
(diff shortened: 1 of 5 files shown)


zbyte64:
Is there a particular reason why we don't pass the provided filterset class through `get_filterset_class` / `setup_filterset` which injects `GrapheneFilterSetMixin` ?

sierreis:
Nope, I just pushed a fix for that part so that it has the same behavior as passing in the FilterSet as an argument to DjangoFilterConnectionField. I think there's some important overriding behavior in the mixin to work with Relay node IDs rather than the database ID.

zbyte64:
Test coverage decreased because a new scenario is not tested:

* Failure when both filter_fields & filter_class are defined

sierreis:
@mvanlonden, the latest commit should have the previously requested test case (when filter_fields and filterset_class are both set), and the coverage is now up 0.2%. Let me know if you had something else in mind.

PR #601, opened 2019-03-25
Add static files to MANIFEST.in

sierreis:
At the moment, static files are not included in the package data
when installing using setuptools. This is necessary for the
GraphiQL view.


danpalmer:
@sierreis nice one, thanks for the contribution!

PR #603, opened 2019-03-27
Adds enhanced support for proxy models.

abettke:
- Allows for proxy models to be used directly in DjangoObjectType definitions.
- Updates tests for proxy model support.
- Removes test showing proxied DjangoObjectType definitions fail.

### DIFF
--- a/graphene_django/tests/test_query.py
+++ b/graphene_django/tests/test_query.py
@@ -1,3 +1,4 @@
+import base64
 import datetime
 
 import pytest
@@ -895,8 +896,7 @@ def test_should_handle_inherited_choices():
 
 def test_proxy_model_support():
     """
-    This test asserts that we can query for all Reporters,
-    even if some are of a proxy model type at runtime.
+    This test asserts that we can query for all Reporters and proxied Reporters.
     """
 
     class ReporterType(DjangoObjectType):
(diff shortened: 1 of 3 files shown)


graphene_django/tests/test_query.py
```diff
@@ -996,14 +947,23 @@ class Query(graphene.ObjectType):
         }
     """
 
+    def str_to_node_id(val):
+        return base64.b64encode(val.encode()).decode()
```
zbyte64: I think we should use `graphql_relay.to_global_id` here.

abettke: :thumbsup: done

abettke:
> Does this need any documentation to go along with it?

None that I can think of. This just adds, what seems to me, expected behavior. Beforehand, using a proxy model in a DjangoObjectType threw a GraphQLError. Now, it fully supports using proxy models the same way that base models are used.

threewordphrase:
I'm waiting with bated breath for this :)

abettke:
For anyone needing this immediately, my fork contains theses changes, and as far as I know is currently up to date with master.

IAmNatch:
Note to anyone searching: Upgrading from 2.2.0 to 2.4.0 resolved my issue with serializing a taggit proxy model. Thanks all who worked on this!

PR #604, opened 2019-03-27
Replaced a copy-paste error causing one test case not to run

GitRon:
Two functions were named like an existing function so the test would not run. Fixed this.


GitRon:
Who's gonna merge it? :)

PR #605, opened 2019-03-27
Missing LOC in django model form documentation (fixes #602)

### LINKED ISSUE #602: Adding DjangoModelFormMutation to schema raises import error
Hi there!

I try to use the awesome `DjangoModelFormMutation` but when I add the mutation to the schema I always get this error:

`Could not import 'project.config.schema.schema' for Graphene setting 'SCHEMA'. AttributeError: 'NoneType' object has no attribute '_meta'.`

Here's my setup:

```
class Pet(models.Model):
    name = models.CharField()

class PetForm(forms.ModelForm):
    class Meta:
        model = Pet
        fields = ('name',)

class PetMutation(DjangoModelFormMutation):
    class Meta:
        form_class = PetForm

class PetMutations(graphene.ObjectType):
    pet = PetMutation.Field(description="Pets!")


class Mutation(PetMutations,
               graphene.ObjectType):
    pass

schema = graphene.Schema(query=Query, mutation=Mutation)
```

Any ideas what I might be doing wrong?

Thanks!
Ron

GitRon:
If you don't put this in your code you'll end up with an import error deep in the source code. Took me about 6h work to figure this out 🙈

### DIFF
--- a/docs/form-mutations.rst
+++ b/docs/form-mutations.rst
@@ -43,6 +43,8 @@ DjangoModelFormMutation
             model = Pet
 
     class PetMutation(DjangoModelFormMutation):
+        pet = Field(PetType)
+
         class Meta:
             form_class = PetForm
 


zbyte64:
I think `PetMutation` is suppose to have field `name`  and not `pet`. What is not shown in the documentation is how to include the `PetMutation`  in the root `Mutation` class, the LOC here is probably also valid but I don't think it is idiomatic.

GitRon:
As I said if you don't put this line you get a vague import error. I really think that this needs to be documented!

GitRon:
After giving it a second thought I really would be happy if we did not need this line at all. So maybe there is another bug in the background? I did not manage to find out where the problem comes from...

PR #608, opened 2019-03-29
Added test class for django api unittests and documentation

GitRon:
When I set up my django project with graphene I was missing a base class and docs how to build unittests. I found a nice example here: https://www.sam.today/blog/testing-graphql-with-graphene-django/

Adjusted it a little to work more smoothly.

What do you think? I'd be happy to have this in the package.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -14,3 +14,4 @@ Contents:
    rest-framework
    form-mutations
    introspection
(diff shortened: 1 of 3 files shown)


GitRon:
Hm, I wrote an abstract class. How would you like that I write a test for this?

PR #609, opened 2019-03-30
Fix passing required=True to DjangoConnectionField

alexkirsz:
Passing `required=True` to an instance of `DjangoConnectionField` currently raises the following exception:

```
  ...
  File "/env/lib/python3.6/site-packages/graphql/type/typemap.py", line 106, in reducer
    field_map = type.fields
  File "/env/lib/python3.6/site-packages/graphql/pyutils/cached_property.py", line 22, in __get__
    value = obj.__dict__[self.func.__name__] = self.func(obj)
  File "/env/lib/python3.6/site-packages/graphql/type/definition.py", line 221, in fields
    return define_field_map(self, self._fields)
  File "/env/lib/python3.6/site-packages/graphql/type/definition.py", line 235, in define_field_map
    field_map = field_map()
  File "/env/lib/python3.6/site-packages/graphene/types/typemap.py", line 274, in construct_fields_for_type
    map = self.reducer(map, field.type)
  File "/env/lib/python3.6/site-packages/graphene-django/graphene_django/fields.py", line 61, in type
    _type, DjangoObjectType
TypeError: issubclass() arg 1 must be a class
```

The reason for the exception is that DjangoConnectionField does not expect its `type` to be wrapped in a `NonNull` instance. This PR ensures that it handles that case properly.

### DIFF
--- a/graphene_django/fields.py
+++ b/graphene_django/fields.py
@@ -1,6 +1,7 @@
 from functools import partial
 
 from django.db.models.query import QuerySet
+from graphene import NonNull
 
 from promise import Promise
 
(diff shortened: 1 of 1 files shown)


graphene_django/fields.py
```diff
@@ -103,15 +118,15 @@ def resolve_connection(cls, connection, default_manager, args, iterable):
...
-        cls,
-        resolver,
-        connection,
-        default_manager,
-        max_limit,
-        enforce_first_or_last,
-        root,
-        info,
-        **args
+            cls,
```
phalt: 8 spaces should be 4!

alexkirsz:
@phalt Done!

alexkirsz:
@phalt The previous behaviour was to raise an exception, so I don't believe there's any way in which this could break non-already broken code :)

phalt:
@alexkirsz approval from me then!

PR #611, opened 2019-04-03
Convert MultipleChoiceField to List of type String

kimbriancanavan:
The django_filters.MultipleChoiceFilter is currently being converted to a String (due to MultipleChoiceFilter eventually inheriting from django.forms.fields.ChoiceField). 

This PR changes that behavior so that it instead gets converted into a List of type String.

### DIFF
--- a/docs/filtering.rst
+++ b/docs/filtering.rst
@@ -124,10 +124,14 @@ create your own ``FilterSet``. You can pass it directly as follows:
         # Do case-insensitive lookups on 'name'
         name = django_filters.CharFilter(lookup_expr=['iexact'])
         # Allow multiple genera to be selected at once
-        genera = django_filters.MultipleChoiceFilter(field_name='genus',
-                                                     choices=(('Canis', 'Canis'),
-                                                              ('Panthera', 'Panthera'),
-                                                              ('Seahorse', 'Seahorse')))
+        genera = django_filters.MultipleChoiceFilter(
+            field_name='genus',
+            choices=(
+                ('Canis', 'Canis'),
+                ('Panthera', 'Panthera'),
+                ('Seahorse', 'Seahorse')
+            )
+        )
 
         class Meta:
             model = Animal
--- a/docs/filtering.rst
+++ b/docs/filtering.rst
@@ -123,6 +123,11 @@ create your own ``FilterSet``. You can pass it directly as follows:
     class AnimalFilter(django_filters.FilterSet):
         # Do case-insensitive lookups on 'name'
         name = django_filters.CharFilter(lookup_expr=['iexact'])
+        # Allow multiple genera to be selected at once
+        genera = django_filters.MultipleChoiceFilter(field_name='genus',
+                                                     choices=(('Canis', 'Canis'),
+                                                              ('Panthera', 'Panthera'),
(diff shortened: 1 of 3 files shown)


kimbriancanavan:
> Is this a breaking change?

@phalt yes, it does change the way MultipleChoiceFilter is resolved in graphql (from simple string to list of strings), however resolving to string is a bug. It is possible that by fixing this bug, some existing queries  in the wild will change from simply not working as intended to returning a proper error (because those queries would now be in violation of the schema)

e.g. Let's say you have a filter defined like so:
```
class ExampleFilter(django_filters.FilterSet):
    my_choice = django_filters.MultipleChoiceFilter(field_name='example', choices=(('A', 'A'), ('B', 'B')))
```
And you wanted to use this filter in a query.  Since the MultipleChoiceField is resolved to string type, schema expects something like the following:

```
query {
    blabla (my_choice: "A") {
        [rest of query here]
    }
}
```

However this will not work, because the MultipleChoiceField expects a list or tuple:

```
class MultipleChoiceField(ChoiceField):
    hidden_widget = MultipleHiddenInput
    widget = SelectMultiple
    default_error_messages = {
        'invalid_choice': _('Select a valid choice. %(value)s is not one of the available choices.'),
        'invalid_list': _('Enter a list of values.'),
    }

    def to_python(self, value):
        if not value:
            return []
        elif not isinstance(value, (list, tuple)):
            raise ValidationError(self.error_messages['invalid_list'], code='invalid_list')
        return [str(val) for val in value]
```

If we make MultipleChoiceField resolve to list of strings then our original query can change to:
```
query {
    blabla (my_choice: ["A"]) {
        [rest of query here]
    }
}
```
or if we want to choose multiple options:
```
query {
    blabla (my_choice: ["A", "B"]) {
        [rest of query here]
    }
}
```
This will now work with the underlying MultipleChoiceField

phalt:
I guess we should either document the change as part of release notes, or implement a deprecation warning so we don't break projects that are using the current behaviour without realizing it is wrong?

kimbriancanavan:
I think I would prefer the documentation option. Any ideas how to proceed with that?

phalt:
@kimbriancanavan looking at previous releases we have updated the documentation or written "migration paths" - code examples of before / after to help people update.

zbyte64:
What is our versioning strategy with backwards incompatible changes?

phalt:
@zbyte64 semver.org I guess. We've done major upgrades in the past with breaking changes, and we've written migration documents to handle it.

ProjectCheshire:
I came across this bug just now!

mvanlonden:
Since there's a breaking change here, let's hold off until we're ready for v3. I think it'd be best to do this once graphene hits v3 which should be by the end of the month.

zbyte64:
In other places we are converting choices into enums, but here we are not. I also wonder if `django_filters.MultipleChoiceFilter` support enums.

dmitrikuksik:
@phalt Stacked with that bug now. Will it be finally merged or appeared with a new graphene version?

jkimbo:
Thanks for the contribution @kimbriancanavan !

topbloc-beiswenger:
Is there a workaround for this in the meantime, or do we have to wait for v3 to be released?

PR #615, opened 2019-04-06
Don't force-capitalize choice fields

### LINKED ISSUE #133: Choice string converter should not force upper case
The convert_choice_name function in converters.py forces all string choices to upper case. These are the values taken from the choices field of a Django model field. This is less than ideal because:

- It forces the developer to use upper case in GraphQL methods, even if lower or mixed case is used everywhere else in Django apps.
- If makes all choices silently case-insensitive. This silently imposes a constraint that Django does not impose, introducing very difficult-to-detect bugs.

It would be preferable for graphene_django to respect the developers' decisions on the case of choice elements, rather than force upper case. The fix is to replace:

        name = to_const(force_text(name))
        try:
            assert_valid_name(name)
        except AssertionError:
            name = "A_%s" % name
        return name

with:

        name = force_text(name)
        try:
            assert_valid_name(name)
        except AssertionError:
            name = "A_%s" % name
        return name

reverie:
Fixes #133. Thanks to Waitak for the patch.

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -29,7 +29,7 @@ singledispatch = import_single_dispatch()
 
 
 def convert_choice_name(name):
-    name = to_const(force_text(name))
+    name = force_text(name)
     try:
         assert_valid_name(name)
     except AssertionError:


phalt:
Related to #67

clarkmoody:
@reverie Can you add a quick test or two to show expected behavior?

phalt:
This is a breaking change and so we are going to implement a better solution. Closing this for now.

reverie:
@phalt Where can I follow along with the better solution?

phalt:
@reverie we're discussing a few options in #67 

I'm trying out a solution similar to #630 to see if that helps. It doesn't stop force-capitalisation but it removes any manual management of choices/enums from the developer.

PR #624, opened 2019-04-30
Docs improvements

phalt:
Rewrite of the documentation, splitting up the docs into the following structure:

* installation - new
* tutorial-plain - currently untouched
* tutorial-relay - currently untouched
* schema - new
* queries - new 
* mutations - new
* filtering
* authorization 
* debug
* introspection
* testing

Django REST Framework and Django forms will move under "mutations".

Relay will move under queries and mutations (where it makes sense).

I am NOT rewriting the tutorials yet as they're huge!

TODO:
- [ ] Document why relay and mutations is important


docs/installation.rst
```diff
@@ -0,0 +1,69 @@
...
+============
+
+Graphene-Django takes a few seconds to install and set up.
+
+Requirements
+------------
+
+Graphene-Django currently supports the following versions of Django:
+
+* Django 2.0 >
```
dopeboy: Django 2.X

docs/installation.rst
```diff
@@ -0,0 +1,69 @@
...
+
+Installation
+------------
+
+.. code:: bash
+
+    pip install graphene-django
+
+**We strongly recommend pinning against a specific version of Graphene-Django as new versions could introduce breaking changes to your project.**
```
dopeboy: I like this suggestion. I'd change "as" to "because"

docs/installation.rst
```diff
@@ -0,0 +1,69 @@
...
+
+    from django.conf.urls import url
+    from graphene_django.views import GraphQLView
+
+    urlpatterns = [
+        # ...
+        url(r'^graphql', GraphQLView.as_view(graphiql=True)),
+    ]
+
+(Changed ``graphiql=True`` to ``graphiql=False`` if you do not want to use the graphiQL API browser.)
```
dopeboy: "Changed" => "Change",
"graphiQL" => "GraphiQL"

Newcomers to GraphQL may not know what GraphiQL is. Might be worth linking to "GraphiQL" to it with a `target="_blank` attribute to https://github.com/graphql/graphiql

docs/tutorial-plain.rst
```diff
@@ -1,12 +1,11 @@
...
+Basic Tutorial
 ===========================================
 
 Graphene has a number of additional features that are designed to make
-working with Django *really simple*.
-
-Our primary focus here is to give a good understanding of how to connect models from Django ORM to graphene object types.
-
-A good idea is to check the `graphene <http://docs.graphene-python.org/en/latest/>`__ documentation first.
+working with Django *really simple*. Our primary focus here is to give
```
dopeboy: Thee emphasis on "really simple" is unneeded (comes off as too boastful IMO).

docs/tutorial-plain.rst
```diff
@@ -1,12 +1,11 @@
...
+a good understanding of how to connect models from Django ORM to graphene
+object types. A good idea is to check the `graphene <http://docs.graphene-python.org/en/latest/>`__
+documentation first, but it is not essential to understand and use graphene-django in your project.
```
dopeboy: I think we can drop this whole paragraph - it's not adding anything to a beginner who's starting out.

docs/installation.rst
```diff
@@ -0,0 +1,69 @@
...
+Graphene-Django takes a few seconds to install and set up.
+
+Requirements
+------------
+
+Graphene-Django currently supports the following versions of Django:
+
+* Django 2.0 >
+
+Installation
```
dopeboy: General comment on the newly added material here - it duplicates some of the work in `tutorial-plain`. Is there a plan to harmonize the both together. If not, happy to volunteer for that.

docs/schema.rst
```diff
@@ -0,0 +1,50 @@
+Schema
+======
+
+The ``graphene.Schema`` object describes your data model, and provides a GraphQL server with an associated set of resolve methods that know how to fetch data. The most basic schema you can create looks like this:
```
dopeboy: Remove the "," after "model"

docs/schema.rst
```diff
@@ -0,0 +1,50 @@
...
+    class Query(graphene.ObjectType):
+        pass
+
+    class Mutation(graphene.ObjectType):
+        pass
+
+    schema = graphene.Schema(query=Query, mutation=Mutation)
+
+
+This schema does nothing like this, but it is ready to accept new Query or Mutation fields that you create.
```
dopeboy: "does nothing like this" is awkward phrasing and is better suited for a situation where you've been talking about "this" and want to draw a comparison. I might do:

"This schema doesn't do anything yet, but it is ready to accept new Query and Mutation fields."

docs/schema.rst
```diff
@@ -0,0 +1,50 @@
...
+
+
+Adding to the schema
+--------------------
+
+If you have defined a ``Query`` or ``Mutation`` you can register them with the schema:
```
dopeboy: "," after `Mutation`

docs/queries.rst
```diff
@@ -0,0 +1,232 @@
+Queries & ObjectTypes
+=====================
+
+Introduction
+------------
+
+The core functionality of GraphQL is to make queries in order to get data.
```
dopeboy: I don't think this sentence is adding much. Maybe "Graphene offers a host of features when performing GraphQL queries."

docs/queries.rst
```diff
@@ -0,0 +1,232 @@
...
+
+Graphene-Django ships with a special ``DjangoObjectType`` that automagically transforms a Django Model
```
dopeboy: As much as I love the word "automagically", I don't think we can use it in our docs. I'd omit it.

docs/mutations.rst
```diff
@@ -0,0 +1,61 @@
+Mutations
+=========
+
+Introduction
+------------
+
+Mutations allow you to `mutate` data.
```
dopeboy: "Graphene makes it easy to perform mutations too."

dopeboy:
This is a *huge* leap forward @phalt, thank you. Most of my feedback above is stylistic and grammatical in nature. The broad strokes of your work are spot on. My only high level concerns:

* There's some duplication of content between the installation & basic tutorial guides. I'm worried our reader might get confused.
* To me, there's two logical groupings of the content that I don't think is obvious. The first three are squarely "get up and going" sort of guides whereas everything that proceeds that is a "let's dig in deeper" sort of flavor. I'm wondering whether it makes sense to distinguish tha.t
* We're treating Relay as an afterthought in a lot of these guides. I'm wondering whether it makes sense to factor it out into its own section?

docs/mutations.rst
```diff
@@ -0,0 +1,61 @@
...
+
+        def mutate(self info, text, id):
+            question = Question.objects.get(pk=id)
+            question.text = text
+            question.save()
+            # Notice we return an instance of this mutation
+            return QuestionMutation(question=question)
+
+
+    class Mutation:
```
subwindow:
```suggestion
    class Mutation(graphene.ObjectType):
```

phalt:
Thanks for the feedback @dopeboy and @subwindow 

Yes, my grammar isn't great so that feedback is useful for me!

Regarding your points:

> There's some duplication of content between the installation & basic tutorial guides. I'm worried our reader might get confused.

I've not yet updated the basic tutorial guide but I want to reduce some of the duplication. I'm also aware the current docs are hard to read because it's all spread out. Does it make sense to duplicate a few things just to make it easier to read through a page in one go? Suggestions welcome.

> To me, there's two logical groupings of the content that I don't think is obvious. The first three are squarely "get up and going" sort of guides whereas everything that proceeds that is a "let's dig in deeper" sort of flavor. I'm wondering whether it makes sense to distinguish that

Yeah I wanted to make this clearer and I think it'll be better as I finish up. I based the docs around how I see the project from the point of view of a user - what does a user want to do versus what does this project provide. For example - currently Django users come to these docs, see a bare minimum of content, and don't know they have to visit the core graphene docs to find what they need.

> We're treating Relay as an afterthought in a lot of these guides. I'm wondering whether it makes sense to factor it out into its own section?

Again, I don't really see Relay as a separate topic but rather an extension on top of Queries / Mutations. This is my opinion, and I'm happy to change it if the consensus is that it should be separate. But I feel like it flows better to talk about Relay for Queries when talking about Queries, etc.

docs/installation.rst
phalt: The tutorials are quite big, I want to generalize things on the other pages, then you or I can go and rewrite the tutorials to conform with it better.

docs/mutations.rst
```diff
@@ -0,0 +1,219 @@
...
+
+
+    class QuestionMutation(graphene.Mutation):
+        class Input:
+            # The input arguments for this mutation
+            text = graphene.String(required=True)
+            id = graphene.ID()
+
+        # The class attributes define the response of the mutation
+        question = graphene.Field(QuestionType)
```
mvanlonden: Can also do `Output = QuestionType` and then just return the question instance in the `mutate` method

mvanlonden:
This is a vast improvement on what we currently have. Thank you 🙏

docs/mutations.rst
phalt: Well today I learned! See, this is why we need the better docs 😆 I'll add this

clarkmoody:
Just wanted to pop in and say thanks for updating the docs 🎉

phalt:
@clarkmoody no problems, wait until you see what I do next!

jkimbo:
@phalt great job! So much better!

phalt:
@jkimbo thanks!

Issue #577, opened 2019-01-25
CharField choices mapped incorrectly to GraphQL


mjmare:
I have this field defined: 
```
    CATEGORY_CHOICES = (('auditief', 'Auditief'),
                        ('beleving', 'Beleving'),
                        ('apparatuur', 'Apparatuur'))
    category = models.CharField('Categorie',
                                max_length=100,
                                choices=CATEGORY_CHOICES,
                                null=True,
                                blank=True,
                                )
```
graphene creates a AspectCategory:
```
AUDITIEF
Auditief
BELEVING
Beleving
APPARATUUR
Apparatuur
```
which already seems wrong to me (shouldn't it be the key instead of the value?).
When querying I get the error:
```
Expected a value of type \"AspectCategory\" but received: Auditief"
```
So for some reason graphene passes the value (not the key), which obviously does not match the GraphQL enum (I think).

phalt:
Discussion is focusing on #67 so closing this to avoid duplicate conversations.

Issue #505, opened 2018-08-27
graphene-django expose form field choices as enum


reinierpd:
Actually, the choice form field is converted to a graphene string when the best solution is an enum field. 
I have a posible solution for this. Implementing a converter function like `convert_django_field_with_choices` with a few changes we can extract the choices an export  enum field.
  ```python 
def convert_django_form_field_with_choices(field):
    """
    Helper method to convert a field to graphene Field type.
    :param field: form field to convert to
    :return: graphene Field
    """
    choices = getattr(field, 'choices', None)

    if choices:
        name = to_camel_case(field.label)  # use the field label as identifier 
        choices = list(get_choices(choices))
        named_choices = [(c[0], c[1]) for c in choices]
        named_choices_descriptions = {c[0]: c[2] for c in choices}

        class EnumWithDescriptionsType(object):
            """Enum type definition"""

            @property
            def description(self):
                """Return field description"""

                return named_choices_descriptions[self.name]

        enum = Enum(name, list(named_choices), type=EnumWithDescriptionsType)
        return enum(description=field.help_text, required=field.required)  
    return convert_form_field(field)

```
Then a few changes on the function `field_for_fom`.

```python 
def fields_for_form(form, only_fields, exclude_fields):
    """
    Helper method for extract form fields and convert it to graphene types Fields
    :param form: form instance
    :param only_fields: form only_fields
    :param exclude_fields: form exclude_fields
    :return: OrderedDict with converted fields
    """
    fields = OrderedDict()
    for name, field in form.fields.items():
        is_not_in_only = only_fields and name not in only_fields
        is_excluded = (
            name
            in exclude_fields  # or
            # name in already_created_fields
        )

        if is_not_in_only or is_excluded:
            continue

        if field.label is None:   # ensure the label is given or set the field name 
            field.label = name

        fields[name] = convert_django_form_field_with_choices(field) # call the implemented function
    return fields
```

Let's take for example the form.
```python
class MyForm(forms.Form):
    STATUS_CHOICES = (
        (PENDING, _('Pending status')),
        (COMPLETED, _('Completed status'))
    )
    status = forms.ChoiceField(choices=STATUS_CHOICES)
.....
```
The input type of the status field on the mutation is now an enum
![image](https://user-images.githubusercontent.com/16706366/44685386-56aeca00-aa19-11e8-9c86-04ebbb982b83.png)

I will be glad to create a pull request if it needed.

kendallroth:
[...]

Issue #471, opened 2018-07-19
How to create a field to expose Model Fields' choices


ZuluPro:
I have a model field with choices: 

```
CHOICES = [
  ('foo', 'Foo'),
  ('bar', 'Bar'),
]

class MyModel(models.Model):
    ham = models.CharField(max_length=2, choices=CHOICES)
```

So, in the schema I have the model and its choices. But can I expose the choices ?
I tried this but without success:
```
MyModelHamType = registry.get_converted_field(models.MyModel._meta.get_field('ham')).get_type()

class Query(object):
    choices = graphene.Field(MyModelHamType)

    def resolve_choices(self, info, **kwargs):
        # What to do here
        return MyModelHamType
```

The code above gives the following thing: 
Request : 
```
query {
  hams
}
```

Response: 
```
{
  "errors": [
    {
      "message": "Expected a value of type \"MyModelHam\" but received: MyModelHam",
      "path": [
        "hams"
      ]
    }
  ],
  "data": {
    "hams": null
  }
}
```

It seems that as the object aren't the of the same type (Pythonicly saying), Graphene doesn't want to handle.

----

I could recreate myself a `MyModelHamType` with `Enum`, but it will be duplicate with the existing one autocreated and already in schema.

ZuluPro:
@sciyoshi Any news ?

dopeboy:
Let's take this over to #67. Closing this down.

OlafHaalstra:
@ZuluPro were you able to fix this issue? Could you explain how?

PR #630, opened 2019-05-08
Manage Django choices & Enums with DjangoChoicesEnum

phalt:
This PR aims to resolve some of the issues seen with choices and enums:

-  #133
- #280
- #577
- #193 
- #240 
- #471 
- #505
- #225
- #517 

They all want different things, so there isn't going to be a solution that makes everyone happy. This PR proposes a solution that covers most of the complaints.

### Managing the difference between an enum and a choice list.

Django's choices accept a list of tuples. Graphene's enums don't support this. How can we find a happy medium?

Solution: Implement a helper class that manages the conversion for us.

```python
from graphene_django import DjangoChoicesEnum

class MyChoices(DjangoChoicesEnum):
    FOO = 'foo'
    BAR = 'bar'

# Get a Django choices list
MyChoices.choices()

# Get a Graphene enum
MyChoices.enum()
```

### How do I see my enums?

We've documented a way of accessing enums through the API, in the documentation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ dev-setup:
 	pip install -e ".[dev]"
 
 tests:
-	py.test graphene_django --cov=graphene_django -vv
+	py.test graphene_django --cov=graphene_django -vv -x
 
 format:
 	black graphene_django
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -26,6 +26,7 @@ For more advanced use, check out the Relay tutorial.
    schema
    queries
    mutations
+   types
    filtering
    authorization
    debug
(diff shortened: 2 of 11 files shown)


ProjectCheshire:
Hah! I had this same issue with neomodel choices

zbyte64:
I wonder if it makes more sense to have `DjangoChoicesEnum` directly inherit from graphene's `Enum` , not sure if all the metamagic is in the way.

zbyte64:
The user still needs to define the enum field at the `DjangoObjectType` for this to work, which I think is the proper way forward. We might want to document that use case.

zbyte64:
Came across another project's attempt to mediate enums and choices, I like the usage pattern here: https://github.com/jazzband/django-model-utils/blob/master/docs/utilities.rst#choices

PR #631, opened 2019-05-08
Fix importing error for GraphQLTestCase

### LINKED ISSUE #628: Cannot import graphene_django.tests.base_test.GraphQLTestCase
when I pip install or download it directly here https://pypi.org/project/graphene-django/2.2.0/ 
base_test.py is missing.

phalt:
As mentioned in #628 you cannot import `GraphQLTestCase`, this is because the directory is [excluded when published](https://github.com/graphql-python/graphene-django/blob/master/setup.py#L55).

The best solution is to move this under `utils` and change the import. It's broken anyway.

Fixes #628

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -9,7 +9,7 @@ Usage:
 
     import json
 
-    from graphene_django.tests.base_test import GraphQLTestCase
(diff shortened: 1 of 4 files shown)


phalt:
Going to ignore the coveralls on this as we're just shuffling around some code.

PR #635, opened 2019-05-13
Update install docs for Django 2.x

davidjb:
This uses the new URL routing syntax with `path()` introduced in Django 2.0 (https://docs.djangoproject.com/en/2.2/releases/2.0/#simplified-url-routing-syntax).  The older `url()` syntax will deprecated at [some point in future](https://docs.djangoproject.com/en/2.2/ref/urls/#url).


phalt:
Thanks @davidjb

PR #647, opened 2019-05-20
bump react to 16.8.6

dulmandakh:
Bump React version to 16.8.6. I checked changes in React and GraphQL-JS, GraphiQL for breaking changes, and didn't found anything.

It works for me.


phalt:
Why do we need to bump this?

dulmandakh:
React 16.2 is so old, let's use latest version if possible.

dulmandakh:
@phalt  we'll get a lot of fixes, improvements. and no breaking changes.

phalt:
All good.

PR #652, opened 2019-05-30
django 1.11.19 is not available, probably for security reasons

### LINKED ISSUE #651: Django 1.11.19 used in cookbook example not available in pypi anymore
Trying to run the cookbook example fails because django 1.11.19 is not available in pypi:

https://pypi.org/project/Django/#history

Perhaps upgrading this exaple's requirements to match the ones of the cookbook-plain would be a good idea? I could do that...

categulario:
Fix #651

### DIFF
--- a/examples/cookbook/requirements.txt
+++ b/examples/cookbook/requirements.txt
@@ -1,5 +1,5 @@
 graphene
 graphene-django
 graphql-core>=2.1rc1
-django==1.11.19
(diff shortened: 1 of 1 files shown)


categulario:
That simply bumps django version by one patch, but if it's necesary I could upgrade django to 2.2 and make the appropriate changes to the project

Issue #141, opened 2017-04-03
_names_ are reserved for future Enum use


Thibaut-Fatus:
Hi,

We run into this problem when using a CharField with choices, but only on the first choice, and the error disappears when we surround the value of the charfield by double underscores:

```
   # this does not work, whatever you choose to set in place of percentage ...
    PERCENTAGE = "percentage",
    AMOUNT = "amount"
    DEAL_TYPES = (
        (PERCENTAGE, "percentage"),
        (AMOUNT, "amount"),
    )

   # ... unless you use '__':
    PERCENTAGE = "__percentage__",
    AMOUNT = "amount"
    DEAL_TYPES = (
        (PERCENTAGE, "__percentage__"),
        (AMOUNT, "amount"),
    )

   # [...]
    type_of_deal = CharField(max_length=14,
                             choices=DEAL_TYPES)
```

Tested on 2 different machines (both MacOS)

graphene==1.2
graphene-django==1.2.1
graphql-core==1.0.1
graphql-relay==0.4.5

Thibaut-Fatus:
the code leading to this error lives in graphene:
https://github.com/graphql-python/graphene/blob/6c7cd4e4fa6e29f55ebf02f173281d46d9f8fc1c/graphene/pyutils/enum.py

but it seems more likely to be an issue with graphene-django

Thibaut-Fatus:
and switching constants declaration and use in the tuple allows us to use 'percentage' in combination of whatever we want as long as it's surrounded by double underscores.

```
    AMOUNT = "__amount__",
    PERCENTAGE = "percentage"
    DEAL_TYPES = (
        (AMOUNT, "__amount__"),
        (PERCENTAGE, "percentage"),
    )
```

tamhv:
i got this issue, i can't use double underscores with ChoiceField from rest framework, always invalid_choice for double underscores choice
```
    def to_internal_value(self, data):
        if data == '' and self.allow_blank:
            return ''

        try:
            return self.choice_strings_to_values[six.text_type(data)]
        except KeyError:
            self.fail('invalid_choice', input=data)
```

Issue #649, opened 2019-05-24
Filtering using Django Integer Choices


robertpro:
Hey guys, I am a real noob with GraphQL, but I think I found a bug, I am trying to filter using an integer choice, below you will see the definitions:

*models.py*
```python
class Event(models.Model):
    KIND_XXX = 1
    KIND_CHOICES = (
        (KIND_XXX, "Dummy description"),
    )
    name = models.CharField(max_length=35)
    kind = models.IntegerField(choices=KIND_CHOICES)
```

*schema.py*
```python
class EventNode(DjangoObjectType):
    class Meta:
        model = Event
        filter_fields = {
            'name': ['exact', 'in', 'icontains', 'istartswith'],
            'kind': ['exact', 'in', 'icontains', 'istartswith'],
        }
        interfaces = (graphene.relay.Node,)

class Query(graphene.ObjectType):
    event = graphene.relay.Node.Field(EventNode)
    all_events = DjangoFilterConnectionField(EventNode)
```

If I query using `name`, `name__in`, etc it works, but when I try to filter by `kind__in` it says I need an `Float` which is wrong.

Please note the Float there:
![image](https://user-images.githubusercontent.com/8595170/58301460-23347c80-7dac-11e9-99ab-196a491047cb.png)

Result:
```json
{
  "errors": [
    {
      "message": "Argument \"kind_In\" has invalid value \"1\".\nExpected type \"Float\", found \"1\".",
      "locations": [
        {
          "line": 2,
          "column": 22
        }
      ]
    }
  ]
}
```

As a second help request, I found other issue (but I am not sure if I am misunderstanding something) if I query for a not existent `kind` it returns the information:

*Query for all records...*

```graphql
query{
  allEvents{
    edges{
      node{
        kind
        name
      }
    }
  }
}
```

*Result for all records*
```json
{
  "data": {
    "allEvents": {
      "edges": [
        {
          "node": {
            "kind": "A_1",
            "name": "Event X"
          }
        }
      ]
    }
  }
}
```

*Query for not existent kind choice*
```graphql
query{
  allEvents(kind: "2"){
    edges{
      node{
        kind
        name
      }
    }
  }
}
```

*Result for not existent kind choice*

```json
{
  "data": {
    "allEvents": {
      "edges": [
        {
          "node": {
            "kind": "A_1",
            "name": "Event X"
          }
        }
      ]
    }
  }
}
```

[...]

PR #654, opened 2019-05-30
A Slightly smarter enum conversion

zbyte64:
Tweaks the enum generator for the django model field converter. When generating an enum name, if the choice value is not a string then use the choice label instead. The result is smarter enum labels for integer fields. Check out the changes in unit testing as an example of better usage.

Not backwards compatible.

Partially addresses #649

### DIFF
--- a/graphene_django/__init__.py
+++ b/graphene_django/__init__.py
@@ -1,6 +1,6 @@
 from .types import DjangoObjectType
 from .fields import DjangoConnectionField
 
-__version__ = "2.2.0"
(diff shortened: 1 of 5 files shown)


zbyte64:
The enum converter now strips non-ascii characters as the graphene does not support non-ascii enum names.

zbyte64:
Will now prefix choices with "A" if it starts with an underscore. That is to fix #141

zbyte64:
Realized that #23 specifically opted to use values for enum generation as they are less likely to change.

PR #657, opened 2019-06-06
Update mutations.rst

sweenr:
I believe the `[1]` was ommitted from the `from_global_id` call as that method returns a tuple of type and id, of which we're only interested in the id here. Took me half a day to figure out why this code wasn't working today. See function def here: https://github.com/graphql-python/graphql-relay-py/blob/master/graphql_relay/node/node.py#L67

### DIFF
--- a/docs/mutations.rst
+++ b/docs/mutations.rst
@@ -214,7 +214,7 @@ You can use relay with mutations. A Relay mutation must inherit from
 
         @classmethod
         def mutate_and_get_payload(cls, root, info, text, id):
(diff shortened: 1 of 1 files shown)


clarkmoody:
While we're here, the same code example starts with

```python
import graphene import relay, DjangoObjectType
from graphql_relay import from_global_id
```

This should probably be expanded to 

```python
import graphene
from graphene import relay
from graphql_relay import from_global_id
```

sweenr:
@clarkmoody That's a good point, thanks for catching that. That was something else I dealt with yesterday. I can make that fix as well.

PR #658, opened 2019-06-06
Stop enforcing csrf checks in GraphQLTestCase

emilgoldsmith:
Not sure what the purpose of that argument to the class instantiation was, maybe it was an artifact from when a wrapper class was called around the Client? But it acts as the `enforce_csrf_checks` argument which was botching up my tests (I copy pasted the class into my own project until you publish 2.3.0)

### DIFF
--- a/graphene_django/utils/testing.py
+++ b/graphene_django/utils/testing.py
@@ -22,7 +22,7 @@ class GraphQLTestCase(TestCase):
                 "Variable GRAPHQL_SCHEMA not defined in GraphQLTestCase."
             )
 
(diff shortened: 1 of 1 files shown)


emilgoldsmith:
I have actually developed quite a bit on my copy paste version of this test case, was thinking of contributing back a lot of my changes, should I just make this PR bigger or do you think I should create a new one_

jkimbo:
@emilgoldsmith could you create a new PR with any further improvements to the test class?

emilgoldsmith:
@jkimbo will do!

PR #665, opened 2019-06-10
Enum conversion fixes

### LINKED ISSUE #141: _names_ are reserved for future Enum use

zbyte64:
Collected backwards compatible fixes from #654 

Fixes #141 and allows using non-ascii characters as choice values.

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -28,10 +28,18 @@ from .utils import import_single_dispatch
 singledispatch = import_single_dispatch()
 
 
+def _is_dunder(name):
+    """Returns True if a __dunder__ name, False otherwise."""
+    return (len(name) > 4 and
+            name[:2] == name[-2:] == '__' and
+            name[2:3] != '_' and
+            name[-3:-2] != '_')
+
+
 def convert_choice_name(name):
     name = force_text(name).encode('utf8').decode('ascii', 'ignore')
     name = to_const(name)
-    if name.startswith('_'):
+    if _is_dunder(name):
         name = "A%s" % name
     try:
         assert_valid_name(name)
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -31,6 +31,8 @@ singledispatch = import_single_dispatch()
 def convert_choice_name(name):
     name = force_text(name).encode('utf8').decode('ascii', 'ignore')
     name = to_const(name)
+    if name.startswith('_'):
+        name = "A%s" % name
     try:
         assert_valid_name(name)
     except AssertionError:
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -30,22 +30,26 @@ singledispatch = import_single_dispatch()
 
 def _is_dunder(name):
     """Returns True if a __dunder__ name, False otherwise."""
-    return (len(name) > 4 and
-            name[:2] == name[-2:] == '__' and
-            name[2:3] != '_' and
-            name[-3:-2] != '_')
+    return (
+        len(name) > 4
+        and name[:2] == name[-2:] == "__"
+        and name[2:3] != "_"
+        and name[-3:-2] != "_"
+    )
 
 
 def _is_sunder(name):
     """Returns True if a _sunder_ name, False otherwise."""
-    return (len(name) > 2 and
-            name[0] == name[-1] == '_' and
-            name[1:2] != '_' and
-            name[-2:-1] != '_')
+    return (
+        len(name) > 2
+        and name[0] == name[-1] == "_"
+        and name[1:2] != "_"
+        and name[-2:-1] != "_"
+    )
 
 
 def convert_choice_name(name):
-    name = force_text(name).encode('utf8').decode('ascii', 'ignore')
+    name = force_text(name).encode("utf8").decode("ascii", "ignore")
     name = to_const(name)
     if _is_sunder(name) or _is_dunder(name):
         name = "A%s" % name
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -36,10 +36,18 @@ def _is_dunder(name):
             name[-3:-2] != '_')
 
 
+def _is_sunder(name):
+    """Returns True if a _sunder_ name, False otherwise."""
+    return (len(name) > 2 and
+            name[0] == name[-1] == '_' and
+            name[1:2] != '_' and
+            name[-2:-1] != '_')
+
+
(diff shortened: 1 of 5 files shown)


graphene_django/converter.py
```diff
@@ -29,7 +29,10 @@
 
 
 def convert_choice_name(name):
-    name = to_const(force_text(name))
+    name = force_text(name).encode('utf8').decode('ascii', 'ignore')
+    name = to_const(name)
+    if name.startswith('_'):
+        name = "A%s" % name
```
jkimbo: @zbyte64 I'm a bit confused with what is happening here. If the name starts with `_` you're prefixing it with a `A`? Why is that?

graphene_django/tests/test_converter.py
```diff
@@ -196,6 +196,24 @@ class Meta:
...
+    )
+
+    class UnderscoreChoicesModel(models.Model):
+        ourfield = field
+
+        class Meta:
+            app_label = "test"
+
+    graphene_type = convert_django_field_with_choices(field)
+    assert len(graphene_type._meta.enum.__members__) == 2
```
jkimbo: Can you assert the actual enum values here as well? So that it's clear what you expect to happen.

graphene_django/converter.py
zbyte64: `assert_valid_name` requires the name not to start with an `_`. Without this, a name like `_amount` would be caught and `A_` would be prefixed to become `A__amount`. With this change, the name becomes `A_amount`, removing the extra `_`

graphene_django/tests/test_converter.py
zbyte64: Done.

graphene_django/converter.py
jkimbo: @zbyte64 I don't think that is right. Running `assert_valid_name` on a values with both `_` and `__` work fine:

```python
In [1]: from graphql import assert_valid_name

In [2]: assert_valid_name('foo')

In [3]: assert_valid_name('_foo')

In [4]: assert_valid_name('__foo')
```

zbyte64: Erm, I mispoke. `assert_valid_name`  will accept names that start in `_` but graphene's Enum will skip over them when populate members. Thus we need to prefix this special case ourselves.

zbyte64: Specifically in the code where this happens: https://github.com/graphql-python/graphene/blob/6c7cd4e4fa6e29f55ebf02f173281d46d9f8fc1c/graphene/pyutils/enum.py#L109

jkimbo: Oh I see. Reading through the code it looks like the logic is a bit more subtle than just ignoring values that begin with `_` though. It looks like it ignores values that are "sunder" or "dunder" only (i.e. `_amount_` or `__amount__`). For example if I construct the following Enum:

```python
In [4]: enum = Enum('foo', [('amount', 'Amount'), ('_amount', 'Amount'), ('__amount', 'Amount'), ('__amount__', 'Amount')])

# It looks like this:
In [5]: enum._meta.enum.__members__
Out[5]:
mappingproxy({'amount': <foo.amount: 'Amount'>,
              '_amount': <foo.amount: 'Amount'>,
              '__amount': <foo.amount: 'Amount'>})
```

(If I try and construct an enum with `('_amount_', 'Amount')` it just fails completely)

From a GraphQL spec point of view names that begin with `__` are not allowed because they are reserved for introspection but names that begin with `_` are allowed. So based on that I would recommend that the logic here be something like:

```python
if _is_sunder(name) or _is_dunder(name):
	name = "A%s" % name
if name.startswith('__'):
	name = "A%s" % name
```

What do you think @zbyte64 ?

graphene_django/tests/test_types.py
```diff
@@ -174,6 +174,7 @@ def test_schema_representation():
 enum ReporterAChoice {
   A_1
   A_2
+  A_3
```
jkimbo: I can't see the corresponding code change for this schema update. Why has this been added?

graphene_django/converter.py
zbyte64: If that's the case then sunder should pass through and the prepend should only apply to dunder.

graphene_django/tests/test_types.py
zbyte64: A third option was added in the test models.py, it starts with an underscore and has a nonascii code.

jkimbo: 👍

graphene_django/converter.py
jkimbo: No I don't think that's right. If a "sunder" is passed to `Enum` then it raises an exception (https://github.com/graphql-python/graphene/blob/6c7cd4e4fa6e29f55ebf02f173281d46d9f8fc1c/graphene/pyutils/enum.py#L131) which is not great. So the code should prefix "sunder" as well as "dunder". Also the code needs to catch names that just start with `__` but don't end with it.

zbyte64: I think I see now. I expanded the test case to cover the various underscore mixtures.

phalt: How does a normal Python enum handle these sunder and dunder values?

zbyte64: Creating a python enum with a sunder causes:
`ValueError: _names_ are reserved for future Enum use`

Dunder is allowed but is ignored (doesn't generate a member).

phalt: Then does it make sense to have the same limitations in the Graphene Enum type instead of tacking `A_` on the front?

zbyte64: We're already converting choices that wouldn't be allowed, like those that start with a number.

PR #667, opened 2019-06-10
Update travis and tox

jkimbo:
* Clean up travis and tox config so it's more inline with the rest of the graphql-python ecosystem
* Add flake8 tox command

Note: I added a black command to tox but I'm not including it in the travis build because there are quite a few formatting changes and I didn't want to include it as part of this PR. Will create a separate PR with the formatting changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: python
 cache: pip
+dist: xenial
 
 install:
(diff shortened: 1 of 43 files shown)


jkimbo:
Add black formatting PR: https://github.com/graphql-python/graphene-django/pull/668

Issue #675, opened 2019-06-15
Field for Postgres RangeField not work on Django 2.2.x( at least)


sprhawk:
I'm using Django 2.2.x, graphene-django 2.3.0

I use a Postgres DateTimeRangeField, but GraphQL reports error ""User Error: expected iterable, but did not find one for field XXX"

I digged in, and found in `graphql`: `execution/executor.py`
```
586     assert isinstance(result, Iterable), (
587         "User Error: expected iterable, but did not find one " + "for field {}.{}."
588     ).format(info.parent_type, info.field_name)
```

where result is NonNull, inside is DateTimeTZRange. which is not declared List(DateTime) declared inside graphene-django converter.py
```
 233 @convert_django_field.register(RangeField)
 234 def convert_posgres_range_to_string(field, registry=None):
 235     inner_type = convert_django_field(field.base_field)
 236     if not isinstance(inner_type, (List, NonNull)):
 237         inner_type = type(inner_type)
 238     return List(inner_type, description=field.help_text, required=not field.null)
```

I want to define a data type to show structure for Range like: 
```
{
    lower
    upper
    bounds
}
```
which will introduce break changes to current API.

is it OK to do this. or what else better solution to this?

elcolie:
This problem still persist with Django3. Here is simple code to get the same error
```
from django.contrib.postgres.fields import DecimalRangeField
from django.db import models


class Premium(models.Model):
    percentage = models.DecimalField(max_digits=10, decimal_places=2)
    sum_insured = DecimalRangeField()   # Insurer accepts this range

    def __str__(self):
        return f"{self.percentage} | {self.sum_insured}"
```

```
from graphene import relay
from graphene_django import DjangoObjectType

from experiment_django3.insurances.models import Premium


class PremiumNode(DjangoObjectType):
    class Meta:
        model = Premium
        filter_fields = [
            'percentage',
            'sum_insured',
        ]
        interfaces = (relay.Node,)
```

cr0nex:
Still a problem

aradkdj:
Sorry to bump this, but it's been almost a year since the last comment. Today I hit this andgot as far as OP, also willing to make a PR if it would be welcomed

kiendang:
@aradkdj Yes a PR would be very much welcomed.

PR #677, opened 2019-06-16
resolve subtypes of postgres RangeField to new RangeType

sprhawk:
Original implementation to resolve Django.contrib.postgres Range series fields to a List type,
which is not working for DateTimeTZRange ( at least for Django 2.2.x )

#675 

What I did is to resolve Range fields to specific RangeType with upper and lower fields inside it, like
```
my_range_filed {
upper
lower
}

I don't whether this is a good method, wish anyone can give me suggestion.

### DIFF
--- a/graphene_django/__init__.py
+++ b/graphene_django/__init__.py
@@ -1,6 +1,6 @@
 from .types import DjangoObjectType
 from .fields import DjangoConnectionField
 
-__version__ = "2.3.0"
+__version__ = "2.3.2"
 
 __all__ = ["__version__", "DjangoObjectType", "DjangoConnectionField"]
--- a/graphene_django/range_types.py
+++ b/graphene_django/range_types.py
@@ -1,0 +1,29 @@
+from graphene import ObjectType
+from graphene import (
+    Float,
+    Int,
+    DateTime,
+    Date
+)
+
+class RangeResolver:
+    def resolve_lower(parent, info):
+        return parent.lower
+    def resolve_upper(parent, info):
+        return parent.upper
+
+class DateTimeRangeType(RangeResolver, ObjectType):
+    lower = DateTime()
+    upper = DateTime()
+
+class DateRangeType(RangeResolver, ObjectType):
+    lower = Date()
+    upper = Date()
+
+class IntRangeType(RangeResolver, ObjectType):
+    lower = Int()
+    upper = Int()
+
+class FloatRangeType(RangeResolver, ObjectType):
+    lower = Float()
+    upper = Float()
--- a/graphene_django/compat.py
+++ b/graphene_django/compat.py
@@ -18,4 +18,6 @@ try:
         DecimalRangeField
     )
 except ImportError:
-    ArrayField, HStoreField, JSONField, RangeField = (MissingType,) * 4
+    ArrayField, HStoreField, JSONField, RangeField, \
+    DateTimeRangeField, DateRangeField, IntegerRangeField, \
+    BigIntegerRangeField, FloatRangeField, DecimalRangeField = (MissingType,) * 10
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Comment by sprhawk** [comment]
I have seen the tests failed for IntegerRangeField, it was working as List type.
So what if I only made DateTimeRangeField working in that way (List type) ?

**Comment by zbyte64** [comment]
I think it would be good to find examples of ranges being used currently in graphene. Unfortunately I couldn't find any.

Here's another graphql project looking to support ranges, but no specifics have been defined: https://github.com/hasura/graphql-engine/issues/1082

File: graphene_django/__init__.py [hunk 1]
```diff
@@ -1,6 +1,6 @@
 from .types import DjangoObjectType
 from .fields import DjangoConnectionField
 
-__version__ = "2.3.0"
+__version__ = "2.3.2"
```
jkimbo: Can you revert this? The version number will be incremented when we create a new release.

graphene_django/range_types.py
```diff
@@ -0,0 +1,29 @@
...
+
+class RangeResolver:
+    def resolve_lower(parent, info):
+        return parent.lower
+    def resolve_upper(parent, info):
+        return parent.upper
+
+class DateTimeRangeType(RangeResolver, ObjectType):
+    lower = DateTime()
+    upper = DateTime()
```
jkimbo: Are these fields `required`?

graphene_django/range_types.py
```diff
@@ -0,0 +1,29 @@
...
+    Date
+)
+
+class RangeResolver:
+    def resolve_lower(parent, info):
+        return parent.lower
+    def resolve_upper(parent, info):
+        return parent.upper
+
+class DateTimeRangeType(RangeResolver, ObjectType):
```
jkimbo: According to http://initd.org/psycopg/docs/extras.html#adapt-range `bounds` and `empty` fields are missing.

graphene_django/range_types.py
```diff
@@ -0,0 +1,29 @@
...
+    Int,
+    DateTime,
+    Date
+)
+
+class RangeResolver:
+    def resolve_lower(parent, info):
+        return parent.lower
+    def resolve_upper(parent, info):
+        return parent.upper
```
jkimbo: If the resolvers just pick attributes on the parent object that are the same name as the field then you don't need these resolvers.

graphene_django/__init__.py
sprhawk: I'll check. I didn't remember I changed version, could it be due to merge from some branch?

graphene_django/range_types.py
sprhawk: Not quite, I think it could be optional.

One problem is the original method which returns a tuple for (lower, upper), it is working for IntegerRange which is tested in test cases, but it is not working for DateTimeRange.

graphene_django/range_types.py
sprhawk: bounds is not accessible. you said missing is referring to mutation?

sprhawk: @jkimbo here is attributes:
```
Object attributes:

isempty
True if the range is empty.

lower
The lower bound of the range. None if empty or unbound.

upper
The upper bound of the range. None if empty or unbound.

lower_inc
True if the lower bound is included in the range.

upper_inc
True if the upper bound is included in the range.

lower_inf
True if the range doesn’t have a lower bound.

upper_inf
True if the range doesn’t have an upper bound.
```

I see original method either not implement it, so I didn't too.

graphene_django/range_types.py
sprhawk: I'll try

PR #680, opened 2019-06-18
Make cookbook example working on django 2

artofhuman:
This PR updates Django version for cookbook example


artofhuman:
Yes, tested using graphiql

Issue #672, opened 2019-06-13
Release 2.3.1 not bump the version


hiroaki-yamamoto:
Hi,

graphene-django 2.3.1 doesn't seem to bump the version (i.e. it's still 2.3.0 on setup.py).
Can you re-make a new release with the bumped version on [graphene_django/__init__.py](https://github.com/graphql-python/graphene-django/blob/ab551f4a15e787fafed8e050cb066b311daae5ed/graphene_django/__init__.py#L4)?

phalt:
I don't believe 2.3.1 is an official release, I think it was done to fix our publishing mechanism. @mvanlonden might know more.

hiroaki-yamamoto:
Ok, I see, but (as fyi,) that release has wrong version that should be fixed.

Close this issue if it's not needed...

phalt:
I'll wait for confirmation from @mvanlonden

imsickofmaps:
2.3.1 fixes real production breaking issues: https://github.com/graphql-python/graphene-django/releases/tag/v2.3.1

At the moment pypi (https://pypi.org/project/graphene-django/#history) is behind as a result of https://github.com/graphql-python/graphene-django/blob/v2.3.1/graphene_django/__init__.py#L4 not being `2.3.1`.

phalt:
okay cool, @mvanlonden can you help us get this version out?

phalt:
We will be doing a `2.3.2` release to fix this issue see #686

jkimbo:
v2.3.2 (exactly the same as v2.3.1) has now been released to PyPi

PR #686, opened 2019-06-20
2.3.2 release

phalt:
2.3.1 did not work, so we are releasing 2.3.2 to resolve this as we can't remove a version from pypi.

@mvanlonden @jkimbo - I am sure I have to do more than this to make it work?

### DIFF
--- a/graphene_django/__init__.py
+++ b/graphene_django/__init__.py
@@ -1,6 +1,6 @@
 from .types import DjangoObjectType
 from .fields import DjangoConnectionField
 
-__version__ = "2.3.0"
+__version__ = "2.3.2"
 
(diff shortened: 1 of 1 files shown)


phalt:
closes #672

jkimbo:
Oh actually hang on. You need to make sure this branch is branched off the v2.3.1 tag not master. Otherwise it will include all the code that’s in master which we don’t want to include until v2.4.0

phalt:
@jkimbo there ain't a branch for it, but there is a tag.

phalt:
@jkimbo Okay I think I figured it out. Checked out the 2.3.1 tag commit, pushed up a `2-3-1` branch, then made this PR merge into `2-3-1`.

PR #688, opened 2019-06-24
Update flake8

jkimbo:
Updated flake8 config to catch simple errors and help with development.

* Add some more flake8 plugins ([flake8-bugbear](https://pypi.org/project/flake8-bugbear/) and [flake8-black](https://pypi.org/project/flake8-black/))
* Run black on `setup.py`
* Fix some issues raised by new flake8 config

### DIFF
--- a/graphene_django/filter/tests/test_fields.py
+++ b/graphene_django/filter/tests/test_fields.py
@@ -321,12 +321,14 @@ def test_filter_filterset_related_results():
         pub_date=datetime.now(),
         pub_date_time=datetime.now(),
         reporter=r1,
+        editor=r1,
     )
     Article.objects.create(
         headline="a2",
         pub_date=datetime.now(),
         pub_date_time=datetime.now(),
         reporter=r2,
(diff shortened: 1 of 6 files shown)


graphene_django/filter/tests/test_fields.py
```diff
@@ -321,12 +321,14 @@ class Query(ObjectType):
         pub_date=datetime.now(),
         pub_date_time=datetime.now(),
         reporter=r1,
+        editor=r1,
```
zbyte64: This seems out of place, but I am also not sure how Articles were being created with a null editor....

jkimbo: @zbyte64 flake8 raised the fact that this test name was the same as another one further down the file and so it was never being run. So when I renamed that test (https://github.com/graphql-python/graphene-django/pull/688/files#diff-42247536fb4398767212e74574263a3bR455) this one started failing because of the missing `editor` param.

PR #689, opened 2019-06-25
Adds documentation to `CAMELCASE_ERRORS` setting

jkimbo:
Following on from https://github.com/graphql-python/graphene-django/pull/514 this PR renames the setting to `CAMELCASE_ERRORS` (the `DJANGO_GRAPHENE` part was unnecessary since it's in the `GRAPHENE` setting already) and adds documentation for the setting.

FYI @kalekseev

### DIFF
--- a/graphene_django/types.py
+++ b/graphene_django/types.py
@@ -192,4 +192,4 @@ class ErrorType(ObjectType):
     @classmethod
     def from_errors(cls, errors):
         data = camelize(errors) if graphene_settings.CAMELCASE_ERRORS else errors
-        return [ErrorType(field=key, messages=value) for key, value in data.items()]
+        return [cls(field=key, messages=value) for key, value in data.items()]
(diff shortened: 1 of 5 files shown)


graphene_django/types.py
```diff
@@ -191,9 +191,5 @@ class ErrorType(ObjectType):
 
     @classmethod
     def from_errors(cls, errors):
-        data = (
-            camelize(errors)
-            if graphene_settings.DJANGO_GRAPHENE_CAMELCASE_ERRORS
-            else errors
-        )
+        data = camelize(errors) if graphene_settings.CAMELCASE_ERRORS else errors
         return [ErrorType(field=key, messages=value) for key, value in data.items()]
```
kalekseev: BTW I forgot to change `ErrorType` to `cls` here.

jkimbo: Thanks @kalekseev . Updated.

kalekseev:
@phalt do you see a way how it can break? We were trying to prevent that with that new option, by default it's off so it shouldn't affect old code.

jkimbo:
@phalt the original PR #514 was just merged and hasn't been released yet. Also because the functionality is setup to be an option (that defaults to the previous behaviour) this isn't a breaking change.

PR #681, opened 2019-06-18
Enforce NonNull for returned related Sets and their content. 

dperetti:
As discussed in https://github.com/graphql-python/graphene-django/issues/44 : when you retrieve the related Set of an object, 
1. You never get None : when no related items are returned, you get an empty Set.
2. You never, ever get a Set of None.
This PR enforces the fact that the Set is NonNull, and so is each object in the Set.

Issue #448, opened 2018-06-02
NonNull missing in conversions to DjangoListField


dperetti:
I think `NonNull` should be enforced at two locations:
1) [Here](https://github.com/graphql-python/graphene-django/blob/master/graphene_django/converter.py#L190), `required=True` should be added:
```python
 return DjangoListField(_type, required=True)
```
2) And [here](https://github.com/graphql-python/graphene-django/blob/master/graphene_django/fields.py#L17), `NonNull()` should be added:
```python
 super(DjangoListField, self).__init__(List(NonNull(_type)), *args, **kwargs)
```
This is because:
1) When you retrieve the related Set of an object, you **never** get `None` : when no related items are returned, you get an empty Set.
2) You never, ever get a Set of `None`.

Current implementation causes issues when using third party tools that introspect the schema, like apollo-codegen. (One gets very annoying nested optionals of nested optionals that make no sense and require additional checking).

Hope that makes sense.
(I can do a PR)

dperetti:
I beg to differ with the stale bot. This is highly important.

jkimbo:
Sorry for the bot @dperetti - we're just trying to get through a very long backlog of issues and it's useful to remove irrelevant old issues and surface the ones that are still a problem (like this one!). I'll reopen this issue since your suggestions sound sensible. Do you think you'll be able to create a PR for this @dperetti ?

dperetti:
Cool, I will !

PR #690, opened 2019-06-25
Enforce NonNull for returned related Sets and their content

jkimbo:
As discussed in https://github.com/graphql-python/graphene-django/issues/44 : when you retrieve the related Set of an object, 
1. You never get None : when no related items are returned, you get an empty Set.
2. You never, ever get a Set of None.
This PR enforces the fact that the Set is NonNull, and so is each object in the Set.

(PR recreated from https://github.com/graphql-python/graphene-django/pull/681 by @dperetti)

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -198,7 +198,9 @@ def convert_field_to_list_or_connection(field, registry=None):
 
             return DjangoConnectionField(_type, description=description)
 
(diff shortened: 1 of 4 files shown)


dperetti:
For reference, the discussion was actually https://github.com/graphql-python/graphene-django/issues/448

PR #691, opened 2019-06-25
Alias `only_fields` as `fields` and `exclude_fields` as `exclude`

jkimbo:
To make it easier for developers who have used Django REST Framework before, change `only_fields` to `fields` and `exclude_fields` to `exclude` so that it matches the `ModelSerializer`: https://www.django-rest-framework.org/api-guide/serializers/#specifying-which-fields-to-include

`only_fields` and `exclude_fields` will still work and we can drop them in a future release.

This PR also includes type checking on the options and support for the `__all__` value to select all `fields` (which mirrors DRF).

### Questions

In a future I think we should enforce that either `fields` or `exclude` should be required on all DjangoObjectTypes. What you think @phalt @zbyte64 @patrick91 @mvanlonden @dopeboy ?

Also do you think we should start logging `PendingDeprecationWarning` for `only_fields` and `exclude_fields`?

### DIFF
--- a/docs/queries.rst
+++ b/docs/queries.rst
@@ -85,7 +85,7 @@ Show all fields **except** those in ``exclude``:
     class QuestionType(DjangoObjectType):
         class Meta:
             model = Question
-            exclude = ('question_text')
+            exclude = ('question_text', )
 
 
 Customising fields
--- a/docs/queries.rst
+++ b/docs/queries.rst
@@ -85,7 +85,7 @@ Show all fields **except** those in ``exclude``:
     class QuestionType(DjangoObjectType):
         class Meta:
             model = Question
-            exclude = ('question_text', )
+            exclude = ('question_text',)
 
 
 Customising fields
@@ -99,7 +99,7 @@ You can completely overwrite a field, or add new fields, to a ``DjangoObjectType
 
         class Meta:
             model = Question
-            exclude_fields = ('question_text')
+            fields = ('id', 'question_text')
 
         extra_field = graphene.String()
 
--- a/docs/queries.rst
+++ b/docs/queries.rst
@@ -63,6 +63,17 @@ Show **only** these fields on the model:
             model = Question
             fields = ('id', 'question_text')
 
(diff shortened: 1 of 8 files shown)


docs/queries.rst
```diff
@@ -57,24 +61,35 @@ Show **only** these fields on the model:
...
+
+Show all fields **except** those in ``exclude``:
 
 .. code:: python
 
     class QuestionType(DjangoObjectType):
         class Meta:
             model = Question
-            exclude_fields = ('question_text')
+            exclude = ('question_text')
```
artofhuman: should be here tuple?
```suggestion
            exclude = ('question_text', )
```

jkimbo: Yes you're right it should be. Will fix the other examples as well. Thanks!

graphene_django/types.py
```diff
@@ -109,10 +114,38 @@ def __init_subclass_with_meta__(
...
+
+        # Alias only_fields -> fields
+        if only_fields and fields:
+            raise Exception("Can't set both only_fields and fields")
+        if only_fields:
+            fields = only_fields
+        if fields and fields != ALL_FIELDS and not isinstance(fields, (list, tuple)):
+            raise TypeError(
+                'The `fields` option must be a list or tuple or "__all__". '
+                "Got %s." % type(fields).__name__
```
phalt: Minor note - we could just use f-strings now that we're only supporting Py3 on the main branch?

jkimbo: @phalt I think it would be good to get this feature out before v3 and then for v3 we can go through and use f strings everywhere.

phalt: Let's leave this for V3 then 👍

jkimbo:
@zbyte64 I agree. I've updated the PR to start raising `PendingDeprecationWarnings` if `only_fields` and `exclude_fields` are used. They can be increased to `DeprecationWarnings` in v3.

jtratner:
+1 to requiring that you specify either fields or exclude!

PR #699, opened 2019-07-06
Adds variables arg to GraphQLTestCase.query

tlambert03:
I was looking for a way to test queries with variables, and formatting the query string directly was a little messy, and I didn't need anything as complex as a custom `input_type`.  If you would prefer I accomplish this by subclassing `GraphQLTestCase`, I understand... but here's a PR just in case you find it useful.  Also added an example to the docs.

### DIFF
--- a/graphene_django/utils/testing.py
+++ b/graphene_django/utils/testing.py
@@ -24,7 +24,7 @@ class GraphQLTestCase(TestCase):
 
         cls._client = Client()
 
-    def query(self, query, op_name=None, input_data=None):
+    def query(self, query, op_name=None, input_data=None, variables=None):
         """
         Args:
             query (string)    - GraphQL query to run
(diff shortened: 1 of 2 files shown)


graphene_django/utils/testing.py
```diff
@@ -24,7 +24,7 @@ def setUpClass(cls):
 
         cls._client = Client()
 
-    def query(self, query, op_name=None, input_data=None):
+    def query(self, query, op_name=None, input_data=None, variables=None):
```
jkimbo: Could you update the docstring below with the new argument?

graphene_django/utils/testing.py
```diff
@@ -40,8 +40,13 @@ def query(self, query, op_name=None, input_data=None):
         body = {"query": query}
         if op_name:
             body["operation_name"] = op_name
+        if variables:
+            body["variables"] = variables
         if input_data:
-            body["variables"] = {"input": input_data}
+            if variables in body and isinstance(body, dict):
```
jkimbo: Why do you need the `isinstance` check here? `body` will always be a dict because it's defined that way above.

graphene_django/utils/testing.py
tlambert03: thanks, updated.

graphene_django/utils/testing.py
tlambert03: I actually meant to check `isinstance(body['variables'], dict)` (just to make sure that the user actually provided a dict to that argument).  but I removed it, because it's a) already specified in the docs, b) not checked for `input_data`  and c) there isn't any fall-back error handling anyway.

McPo:
Would be good to get this merged in. Seems odd that input_type is an arg but variables isn't.

Issue #726, opened 2019-08-01
blank=False on updates


dan-klasson:
When setting `blank` on the model field to `False`. Why does Graphene throw a `Expected "Foo!"` error on an update? Shouldn't this only happen on a create? Because having to pass in unnecessary data seem to go against the whole purpose of GraphQL.

dan-klasson:
Did some digging and setting required to `False` in `convert_serializer_field` solves this problem. Then in your mutation that extends from `SerializerMutation` you can set `partial` to `True`:

```py
    @classmethod
    def get_serializer_kwargs(cls, root, info, **input):
        kwargs = super(PrivateMessageMutation, cls).get_serializer_kwargs(root, info, **input)
        if 'id' in input:
            kwargs['partial'] = True
        return kwargs
```
With the added benefit that you get a clear error message `This field is required.` instead of the `Expected "Foo!"` one.

I can create a PR if you want this change.

jkimbo:
@dan-klasson this is just related to the rest_framework integration correct? I don't have a huge amount of experience with that part of the code but the code snippet you pasted seems like a reasonable change. Can you make a PR for it?

(Also this issue is not related to this: https://github.com/graphql-python/graphene-django/issues/725 is it?)

dan-klasson:
@jkimbo I started looking into this when I noticed my model forms overwriting data that wasn't passed in. I then also noticed the validation problem on updates (i.e no support for partial updates). That's why I decided to switch to model serializers instead. So this PR would only be DRF related.

jkimbo:
Ok thanks for clarifying @dan-klasson

oaosman84:
Is anyone else interested in seeing this fixed? Is there someone who's specifically maintaining the DRF-integration code (under `graphene_django/rest_framework`), since this might end up being a breaking change here where we might need to discuss tradeoffs?

jkimbo:
@oaosman84 unfortunately there isn't anyone in particular looking after the DRF integration and I don't have much experience with it. If you are able to help finish @dan-klasson 's PR that would be great!

ulgens:
I'm removing the wontfix the label and opening the issue back. I'm willing to take the issue but can't promise about when 🙂

ulgens:
@firaskafri It seems this issue still exists. Not sure why it got closed.

PR #731, opened 2019-08-03
added support for partial updates in serializers

dan-klasson:
Related to this issue: https://github.com/graphql-python/graphene-django/issues/726


jkimbo:
@dan-klasson are you able to add those tests?

thomaszdxsn:
It not work on one serializer support both create/update.

because create often need all fields required. and partial update don't need, so I think should make two  InputType

PR #777, opened 2019-09-17
Convert nullable BooleanField to nullable Boolean.

### LINKED ISSUE #486: Django 2.1 BooleanField(null=True) syntax not supported by DjangoObjectType
[As of Django 2.1,](https://docs.djangoproject.com/en/2.1/ref/models/fields/#django.db.models.BooleanField) the syntax `BooleanField(null=True)` is now supported by Django (in addition to the syntax NullBooleanField() but the `null=True` is ignored by `DjangoObjectType`.

This only a problem with Django 2.1.

The bug can be reproduced like

    import graphene
    from django.db import models
    
    class Car(models.Model):
        class Meta:
            app_label = 'app'
        is_fast = models.BooleanField(null=True)
    
    class CarNode(DjangoObjectType, model=Car):
        pass
    
        car = graphene.Field(CarNode)
    
        def resolve_car(self, info):
            return Car(is_fast=None)
        
    result = schema.execute('{ car { isFast } }')
    
    print(result.data, result.errors)

jckw:
BooleanFields can now be made nullable using `null=True` in Django. Graphene used to convert all BooleanFields to a NonNull type; this commit fixes converting nullable BooleanFields.

Fixes #486.

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -127,13 +127,9 @@ def convert_field_to_int(field, registry=None):
     return Int(description=field.help_text, required=not field.null)
 
 
+@convert_django_field.register(models.NullBooleanField)
 @convert_django_field.register(models.BooleanField)
 def convert_field_to_boolean(field, registry=None):
-    return NonNull(Boolean, description=field.help_text)
-
-
-@convert_django_field.register(models.NullBooleanField)
-def convert_field_to_nullboolean(field, registry=None):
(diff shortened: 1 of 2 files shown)


k90armstrong:
Any update on this?

melvinkcx:
any update on this? this is still prevalent.

jkimbo:
Fixed the tests. @ProjectCheshire can you review again?

ulgens:
@stale :(

PR #783, opened 2019-09-24
The current csrftoken parser breaks on multiple crsftokens

allen-munsch:
Which causes 403 errors.

Screenshot from 2019-09-24 09-11-47
For example: 

`Cookie: csrftoken=asdfasd; sessionid=asdfasdf; csrftoken=qwertyqwerty`

This can happen when multiple csrftokens for different paths.

### DIFF
--- a/graphene_django/static/graphene_django/graphiql.js
+++ b/graphene_django/static/graphene_django/graphiql.js
@@ -71,7 +71,7 @@
         try {
           return JSON.parse(responseBody);
         } catch (error) {
-          if (cookies.csrftoken.length) {
+          if (cookies.csrftoken && cookies.csrftoken.length) {
             headers['X-CSRFToken'] = cookies.csrftoken.pop();
             return getFetch(headers);
           }
--- a/graphene_django/static/graphene_django/graphiql.js
+++ b/graphene_django/static/graphene_django/graphiql.js
@@ -8,6 +8,13 @@
     x[key] = p[key] ? p[key].concat([value]) : [value];
     return Object.assign(p, x);
   }, {});
+  var headers = {
+    Accept: 'application/json',
+    'Content-Type': 'application/json'
+  };
+  if (cookies.csrftoken && cookies.csrftoken.length) {
+    headers['X-CSRFToken'] = cookies.csrftoken.pop();
+  }
   // Collect the URL parameters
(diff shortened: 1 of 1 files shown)


allen-munsch:
related: https://github.com/graphql-python/graphene-django/pull/585

graphene_django/static/graphene_django/graphiql.js
```diff
@@ -39,27 +54,31 @@
...
+      })
+      .then(function(responseBody) {
+        try {
+          return JSON.parse(responseBody);
+        } catch (error) {
+          if (cookies.csrftoken && cookies.csrftoken.length) {
+            headers['X-CSRFToken'] = cookies.csrftoken.pop();
+            console.log('retry', headers)
+            return getFetch(headers);
+          }
```
jkimbo: There could be other reasons why the GraphQL endpoint doesn't return valid JSON so always assuming it's to do with a csrf token issue doesn't feel right.

allen-munsch: see: https://stackoverflow.com/questions/576535/cookie-path-and-its-accessibility-to-subfolder-pages

From what I understand, the cookie sub path `/2/m/2` would have both it's subpath cookie and the root path cookie. `/`.

The two paths have different database multitenancy requirements. It would probably be more accurate to read the cookie path directly, instead of just reading, checking them all?

jkimbo: Yes I understand that cookies can have different paths but looking at the Django source code the csrf cookie should be set to a path that is defined in the settings: https://github.com/django/django/blob/5a68f024987e6d16c2626a31bf653a2edddea579/django/middleware/csrf.py#L191

So I don't think it's possible to have 2 csrf cookies at the same time and I'm wondering how that came about?

allen-munsch: endpoints per multi tenancy. `localhost:8000/:multitenantid/endpoints/here` so multiple users can get different csrftoken cookies on a single machine ... it's not the default of django, but it is a real use case.

jkimbo: Ah I didn't realise that was possible! Could you not host the graphql endpoint per tenant as well?

Issue #705, opened 2019-07-09
☂️Graphene-Django v3


jkimbo:
This issue is to track v3 of Graphene-Django which will contain some breaking changes.

WIP branch: [`v3`](https://github.com/graphql-python/graphene-django/tree/v3) ([compare](https://github.com/graphql-python/graphene-django/compare/master...v3))

## Breaking changes

* [x] Upgrade to Graphene v3 (which also upgrades graphql-core to v3) (PR work in progress: https://github.com/graphql-python/graphene-django/pull/905)
* [x] Convert MultipleChoiceField to List of type String (#611)
* [x] Start raising `DeprecationWarnings` for using `only_fields` and `exclude_fields` (see https://github.com/graphql-python/graphene-django/pull/691) (PR #980)
* [x] Start warning if neither `fields` or `exclude` are defined on `DjangoObjectTypes` https://github.com/graphql-python/graphene-django/issues/710 (PR #981)
* [x] Default `CAMELCASE_ERRORS` setting to `True` (PR #789)
* [x] Rename `DJANGO_CHOICE_FIELD_ENUM_V3_NAMING` to `DJANGO_CHOICE_FIELD_ENUM_V2_NAMING` and default it to `False` (reference #860) (PR #982)
* [x] Convert decimal fields correctly: https://github.com/graphql-python/graphene-django/issues/91

ktosiek:
Hey stale bot, this issue is not dead yet, it's just getting started!

ktosiek:
Has anyone started working on this? I'm currently trying to just make tests pass under graphene 3.

swist:
Is there an up to date todolist on this? I have a couple of days this week so can pick it up and finish off your branch

ktosiek:
I think it's complete enough to try building on top of it (porting an actual application, or checking if graphene-django-optimizer works).
It would also be nice to drop the dependency on `promise`.

patrick91:
@ktosiek we could try with https://github.com/graphql-python/swapi-graphene

I wanted to test graphene 3 at work, but we are still on v1 :D

swist:
I've just tried it on my current project. Seems to work fine minus some of our custom graphene types that allow us to use `django_more/django_enum`

ganwell:
Link PR #774

BfutureP:
graphene-django doesn;t support graphql 3, says `from graphql import get_default_backend` not found

garimagupta715:
[...]

PR #789, opened 2019-10-01
Default CAMELCASE_ERRORS to True

swist:
Closes one item from the todolist on #705


swist:
Do you guys still want this?

jkimbo:
@swist yes we do but we're just waiting until we're ready to release it as part of a v3 release.

jkimbo:
Fixed in #968

jkimbo:
Thanks for the contribution @swist !

Issue #787, opened 2019-09-30
Handling aggregate objects


SpyTec:
In my application right now I've been aggregating database entries  and serving them in a REST API. Though when it comes to graphene I've not had any luck getting aggregates to works. Any help would be thankful

Current code:
```py
aggregated_snapshots = DjangoFilterConnectionField(IssueSnapshotNode)

def resolve_aggregated_snapshots(self, info, **kwargs):
    return IssueSnapshot.objects.order_by() \
        .values('status', type=F('issue__type_name')) \
        .annotate(status_count=Count('status'))
```

Error message returned:
```
Merging 'QuerySet' classes must involve the same values in each case
```

zbyte64:
Looks like issue #758

SpyTec:
Probably. Before the latest update I also tried to do this, though at that time I got a different error I don't remember at the moment

zbyte64:
After some digging, you might want to try replacing `values` with `only` or `defer`. PR #796 revealed how the two are treated differently and that values delivers dictionaries while graphene_django want's django instances.

SpyTec:
Maybe this is just my inexperience, but from what I can see getting a field from a related model `issue__type_name` won't work with only() as it will only show you the `issue_id` field, not the `type_name`

zbyte64:
Sounds like we might want the filtering functionality of `DjangoFilterConnectionField` but with the option to change the type. Might be able to subclass and override `node_type` as well as `type` ...

SpyTec:
I'm not well into the library implementation, what does `node_type` and `type` represent and affect?

zbyte64:
I believe `node_type` is the inner type that the connection field connects to, `type` adds wrappings, basically just `NotNull` with a type check which ensures we are connecting it to a django object type.

SpyTec:
Yeah, this has been a bit over my head. Not been successful with aggregating based off of a relation of `issue__type_name`

SpyTec:
Not had any success yet

Issue #758, opened 2019-08-27
merge_querysets breaks annotations with DjangoFilterConnectionField


jarcoal:
We recently updated to v2.5 and have found that many of our connections that use `DjangoFilterConnectionField` no longer work as expected (aka the filters aren't applied).

After some digging it looks like any filter that uses annotations are having them stripped off by `DjangoConnectionField.merge_querysets` [here](https://github.com/graphql-python/graphene-django/blob/master/graphene_django/fields.py#L96).

Specifically it looks like PR #693 started this issue because before that `merge_querysets` wouldn't be called in many situations.

Thoughts?  Is there a better way to go handling annotations or this is just a bug?

aminland:
+1, this doubled the number of joins happening in our queries, causing everything to slow down.

zbyte64:
@jarcoal I believe you are correct, `merge_querysets` is to allow merging of queries from the Node and Connection.

jarcoal:
I thought I would paste our hacked together solution for this so that it could be discussed as an alternative to the current implementation.  The goal here was to pass a single queryset through the entire code path so that all interested parties can tack on their filters/annotations/prefetches/etc.

```python
from graphene_django.filter import DjangoFilterConnectionField


class NonMergingDjangoFilterConnectionField(DjangoFilterConnectionField):
    @classmethod
    def merge_querysets(cls, default_queryset, queryset):
        return queryset

    @classmethod
    def connection_resolver(
        cls,
        resolver,
        connection,
        default_manager,
        max_limit,
        enforce_first_or_last,
        filterset_class,
        filtering_args,
        root,
        info,
        **args
    ):
        first = args.get("first")
        last = args.get("last")

        if enforce_first_or_last:
            assert first or last, (
                "You must provide a `first` or `last` value to properly paginate the `{}` connection." # noqa
            ).format(info.field_name)

        if max_limit:
            if first:
                assert first <= max_limit, (
                    "Requesting {} records on the `{}` connection exceeds the `first` limit of {} records." # noqa
                ).format(first, info.field_name, max_limit)
                args["first"] = min(first, max_limit)

            if last:
                assert last <= max_limit, (
                    "Requesting {} records on the `{}` connection exceeds the `last` limit of {} records." # noqa
                ).format(last, info.field_name, max_limit)
                args["last"] = min(last, max_limit)

        iterable = resolver(root, info, **args)

        filter_kwargs = {k: v for k, v in args.items() if k in filtering_args}

[...]

Issue #429, opened 2018-04-17
DjangoConnectionField.resolve_connection replaces resolved queryset with clone, wiping query cache (causing N+1)


richardpetithory:
I'm new to graphene-django so please bear with me if my issue is nonsensical.

I've been spending some time trying to figure out why, when using a resolution method in combination with a `DjangoFilterConnectionField` any prefetches are being unused. I've traced it down to a few things.

Within `DjangoFilterConnectionField.merge_querysets` an `&` operator is used to merge two Django `QuerySet`s. The `__and__` method on `QuerySet` creates a new instance of its class, thereby losing any internal caching that may have been accomplished when using `prefetch_related()`. There doesn't seem to be much of a way around this without accessing private attributes of the `QuerySet` object, which is clearly bad practice.

I then noticed that `DjangoConnectionField.resolve_connection` checks to see `if iterable is not default_manager`. If it's not (which as far as I can tell it never will be if you have a `resolve_things` method on your Query class) it will then follow down the aforementioned path which winds up wiping cache because it creates a new `QuerySet`.

I wonder, what is the purpose of the following code in `resolve_connection`? If I disable it, the queryset my resolve_things method returns is actually used and thus the prefetch works.

```
            if iterable is not default_manager:
                default_queryset = maybe_queryset(default_manager)
                iterable = cls.merge_querysets(default_queryset, iterable)
```

For reference, my implementation is as follows:
```
class PostType(DjangoObjectType):
    class Meta:
        model = Post
        filter_fields = ['id', 'title']
        interfaces = (graphene.relay.Node, )


class Query(graphene.ObjectType):
    posts = DjangoFilterConnectionField(PostType)
    
    def resolve_posts(self, info, **kwargs):
        return Post.objects.prefetch_related('customer_posts').all()
```

richardpetithory:
@arianon Hope you don't mind me tagging you as you seem to be the one who implemented this logic.

arianon:
I'm not, I merely [lifted up that logic to a separate method](https://github.com/graphql-python/graphene-django/pull/180/files#diff-4693e93918b75fd2ff08fbe53fec5118L91) to work with Promises.

I'm no longer familiar with graphene-django so I can't help you here either, apologies and godspeed.

richardpetithory:
Gotcha. Thanks for chiming in!

richardpetithory:
[...]

PR #796, opened 2019-10-11
replace merge_queryset with resolve_queryset pattern

zbyte64:
There are a set of issues from our use of `merge_queryset`:

#429 : queryset cache gets nuked
#787 : Aggregate objects not supported
#758 : Annotations are lost

To resolve this we attempt to maintain a single queryset throughout the connection resolver instead of merging multiple querysets via the `&` operator. To accomplish this we replace the `merge_queryset` method with a `get_queryset_resolver` that returns a callable that modifies a queryset passed through. 

As a result, the `DjangoFilterConnectionField` has a more terse interface for specifying how to modify filters.

Also I am not sure if `graphene_django.filter.tests.test_fields.test_should_query_filter_node_double_limit_raises` is still a valid test, but we should decide if we still wish to check limit changes even though it doesn't seem necessary anymore (this conflict doesn't raise a an error)

What is missing are tests for annotations and aggregations. If this purposed pattern is acceptable then I will write the unit tests for these.

### DIFF
--- a/graphene_django/fields.py
+++ b/graphene_django/fields.py
@@ -39,7 +39,7 @@ class DjangoListField(Field):
         if queryset is None:
             # Default to Django Model queryset
             # N.B. This happens if DjangoListField is used in the top level Query object
-            model = django_object_type._meta.model.objects
+            model_manager = django_object_type._meta.model.objects
             queryset = maybe_queryset(django_object_type.get_queryset(model, info))
         return queryset
 
--- a/graphene_django/fields.py
+++ b/graphene_django/fields.py
@@ -39,10 +39,8 @@ class DjangoListField(Field):
         if queryset is None:
             # Default to Django Model queryset
             # N.B. This happens if DjangoListField is used in the top level Query object
-            model = django_object_type._meta.model
-            queryset = maybe_queryset(
-                django_object_type.get_queryset(model.objects, info)
-            )
+            model = django_object_type._meta.model.objects
+            queryset = maybe_queryset(django_object_type.get_queryset(model, info))
         return queryset
 
     def get_resolver(self, parent_resolver):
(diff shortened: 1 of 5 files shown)


graphene_django/fields.py
```diff
@@ -39,10 +39,8 @@ def list_resolver(django_object_type, resolver, root, info, **args):
         if queryset is None:
             # Default to Django Model queryset
             # N.B. This happens if DjangoListField is used in the top level Query object
-            model = django_object_type._meta.model
-            queryset = maybe_queryset(
-                django_object_type.get_queryset(model.objects, info)
-            )
+            model = django_object_type._meta.model.objects
```
jkimbo:
```suggestion
            model_manager = django_object_type._meta.model.objects
```

Bit of a nit pick but the variable name `model` is misleading here.

graphene_django/filter/tests/test_fields.py
```diff
@@ -608,6 +608,7 @@ def resolve_all_reporters(self, info, **args):
     assert result.data == expected
 
 
+@pytest.mark.skip(reason="no longer relevant?")
 def test_should_query_filter_node_double_limit_raises():
```
jkimbo: I think you can remove this test @zbyte64

zbyte64: Done.

zbyte64:
Thanks for the feedback. 

Added a test for annotations. Should I add a test specifically for interactions with `DjangoFilterConnectionField`?

jkimbo:
> Should I add a test specifically for interactions with DjangoFilterConnectionField?

@zbyte64 I think that would be really good actually because it feels like that part of the code is quite brittle.

zbyte64:
There seems to already be an annotation test for django-filters: `graphene_django.filter.tests.test_fields.test_annotation_is_perserved`

I guess I need to go back and write a failing test because I don't know how that one passed.

zbyte64:
Seems the key difference is using `values`. The code changes lets it get further along, but later `graphene_django.type.DjangoObjectType.is_type_of` raises an error. We can change this to return True for dictionaries, which is the result of a `values` query, but we would need to wrap it so that the key could be accessed as properties.

zbyte64:
I think going forward we can support `only` and `defer` with querysets and but `values` is still problematic.  Added a django-filters test: `test_annotatoin_with_only` to test just that.

jkimbo:
Again sorry for the very late review @zbyte64 . Merging

jkimbo:
Released as v2.7.0

PR #797, opened 2019-10-12
Add raise exception on serializer validation

no-dap:
SerializerMutation with invalid input doesn't return any validation errors.  
- on graphql browser
Screen Shot 2019-10-13 at 7 09 41 AM  
- on ajax request  
Screen Shot 2019-10-13 at 7 24 02 AM

(In my case, content input shouldn't be blank.)
  
I think we need to add argument raise_exception=True on serializer.is_valid().
- on graphql browser  
Screen Shot 2019-10-13 at 7 10 10 AM
- on ajax request  
Screen Shot 2019-10-13 at 7 24 35 AM
(worked as I expected)

### DIFF
--- a/graphene_django/rest_framework/mutation.py
+++ b/graphene_django/rest_framework/mutation.py
@@ -124,7 +124,7 @@ class SerializerMutation(ClientIDMutation):
         kwargs = cls.get_serializer_kwargs(root, info, **input)
         serializer = cls._meta.serializer_class(**kwargs)
 
-        if serializer.is_valid():
+        if serializer.is_valid(raise_exception=True):
             return cls.perform_mutate(serializer, info)
         else:
             errors = ErrorType.from_errors(serializer.errors)
--- a/graphene_django/tests/test_views.py
+++ b/graphene_django/tests/test_views.py
@@ -446,7 +446,6 @@ def test_handles_field_errors_caught_by_graphql(client):
     response = client.get(url_string(query="{thrower}"))
     assert response.status_code == 400
     assert response_json(response) == {
-        "data": None,
         "errors": [
             {
                 "locations": [{"column": 2, "line": 1}],
--- a/graphene_django/tests/test_views.py
+++ b/graphene_django/tests/test_views.py
@@ -444,7 +444,7 @@ def test_supports_pretty_printing_by_request(client):
 
 def test_handles_field_errors_caught_by_graphql(client):
     response = client.get(url_string(query="{thrower}"))
-    assert response.status_code == 200
+    assert response.status_code == 400
     assert response_json(response) == {
         "data": None,
         "errors": [
--- a/graphene_django/rest_framework/tests/test_mutation.py
+++ b/graphene_django/rest_framework/tests/test_mutation.py
@@ -1,6 +1,8 @@
 import datetime
 
 from py.test import mark, raises
+from rest_framework.exceptions import ValidationError
+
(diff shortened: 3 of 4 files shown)


graphene_django/rest_framework/mutation.py
```diff
@@ -124,7 +124,7 @@ def mutate_and_get_payload(cls, root, info, **input):
         kwargs = cls.get_serializer_kwargs(root, info, **input)
         serializer = cls._meta.serializer_class(**kwargs)
 
-        if serializer.is_valid():
+        if serializer.is_valid(raise_exception=True):
             return cls.perform_mutate(serializer, info)
         else:
             errors = ErrorType.from_errors(serializer.errors)
```
jkimbo: I would expect that this line would propagate the errors to the execution result (instead of needing to raise an exception). @no-dap do you know why it doesn't?

no-dap: I found this only happens with blank input in non-blankable field. I'm watching for what's going wrong now.

Screen Shot 2019-10-19 at 10 26 47 AM
(This works properly with an invalid input type)

no-dap: @jkimbo 
I found that when we append errors to MutationPayload class, works properly at first.  
Screen Shot 2019-10-19 at 11 15 37 AM
(The picture above is an ObjectType's `__init__` function)
  
But ObjectType initiated again to give some value for requested result, (which contains id)    
if we don't raise any exception.
Screen Shot 2019-10-19 at 11 19 04 AM
I think this comes over the ErrorType, so we get an response instead of error messages.  
I'm not sure because I'm new on graphene,  
but I guess graphene's intend was raise exception  if somethings go wrong,  
rather than just provide error to execution result.

no-dap: I think we have to check about other mutations too.  
I'm on spare time now :)

no-dap: I think this is a graphene's bug.  
I could get an error message when I requested errors as query.  
Screen Shot 2019-10-19 at 12 31 41 PM

no-dap:
document.execute doesn't return errors, but errors contained in data.  
(graphene_django.views, L276)
Screen Shot 2019-10-19 at 1 35 50 PM

Errors are append via this function, which need raising exception to call, I guess.  
(graphql.execution.utils, L150)
Screen Shot 2019-10-19 at 1 51 29 PM

document.execute() doesn't occurs any error in this case,  
we have to add  some checks to return that request was invalid.

graphene_django/tests/test_views.py
```diff
@@ -444,7 +444,7 @@ def test_supports_pretty_printing_by_request(client):
 
 def test_handles_field_errors_caught_by_graphql(client):
     response = client.get(url_string(query="{thrower}"))
-    assert response.status_code == 200
+    assert response.status_code == 400
```
no-dap: @jkimbo 
Isn't it should be 400?

no-dap:
Sorry about things are messed up :(
I think this should be separated into two issues  
(Raising errors in is_valid call and return 400 if inputs are invalid)  
I'll close this issue when I can separate it.

graphene_django/rest_framework/mutation.py
no-dap: @jkimbo 
https://github.com/graphql-python/graphene-django/pull/803#issuecomment-544212540

PR #803, opened 2019-10-20
Add raise exception on serializer validation

no-dap:
re-created PR
Refs : [#797](https://github.com/graphql-python/graphene-django/pull/797)

### DIFF
--- a/graphene_django/rest_framework/tests/test_mutation.py
+++ b/graphene_django/rest_framework/tests/test_mutation.py
@@ -5,6 +5,7 @@ from rest_framework import serializers
 
 from graphene import Field, ResolveInfo
 from graphene.types.inputobjecttype import InputObjectType
+from rest_framework.exceptions import ValidationError
 
 from ...settings import graphene_settings
 from ...types import DjangoObjectType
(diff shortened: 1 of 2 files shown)


no-dap:
## Overview
Screen Shot 2019-10-19 at 1 35 50 PM
  
Errors didn't appended to ExecutionResult's errors if we don't raise errors,  
they just exists inside data.  

Screen Shot 2019-10-19 at 1 00 25 PM

We can request errors as a result query, but this looks super weird.  
  
## Another issue
DjangoFormMutation, SerializerMutation with invalid inputs return status code 200.  
I opened another issue [#804](https://github.com/graphql-python/graphene-django/issues/804)

no-dap:
hello there?

graphene_django/rest_framework/tests/test_mutation.py
```diff
@@ -204,20 +205,23 @@ class Meta:
             serializer_class = MySerializer
 
     # missing required fields
-    result = MyMutation.mutate_and_get_payload(None, mock_info(), **{})
-    assert len(result.errors) > 0
+    with raises(ValidationError):
+        result = MyMutation.mutate_and_get_payload(None, mock_info(), **{})
+        assert len(result.errors) > 0
```
zbyte64: Does this assert ever get called? If I understand correctly line 209 should raise a `ValidationError` thus line 210 never gets called.

no-dap: No it doesn't. I think line 210 is unnecessary indeed.

PR #805, opened 2019-10-21
django-filter: resolve field along with lookup expression to properly lookup form field

### LINKED ISSUE #781: Filtering not working correctly with 2.6.0
Related to #750. Appreciate the fix for this problem!!

### Problem
When using filter_fields I get an error about using wrong types which started appearing in 2.4.0.
`Variable "userEmail" of type "String" used in position expecting type "ID".` The error does not occur with graphene-django 2.3.2

### Context
- using django-filter 2.2.0
- django 2.4.0
- graphene-django 2.6.0
### 

**model.py**
```
class Membership(TimeStampedModel):
    user = models.ForeignKey(User, on_delete=models.CASCADE)
    tenant = models.ForeignKey(Tenant, on_delete=models.CASCADE)



class User(TimeStampedModel, AbstractBaseUser, PermissionsMixin):
    email = EmailField(unique=True, verbose_name=_('email'))

    USERNAME_FIELD = 'email'
    REQUIRED_FIELDS = []

    objects = UserManager()
```

**Schema.py**
```
class MembershipNode(DjangoObjectType):
    class Meta:
        model = Membership
        filter_fields = {
            'id': ['exact'],
            'user__email': ['exact'],
        }
        interfaces = (MembershipNodeInterface,)

```

**Query:**
```
QUERY_MEMBERSHIPS = '''
    query memberships($tenant: String!, $userEmail: String) {
        memberships(tenant: $tenant, user_Email: $userEmail) {
            edges {
                node {
                    id
                    isFitter
                    isMonitor
                    isAdmin
                    isStaff
                }
            }
        }
    }
'''
```

**Result:**
`Variable "userEmail" of type "String" used in position expecting type "ID".`

### Solution
Should be related to #750. Might be a special case due to the `email` being the identifying field of the `User`

> I am confident it is related to this PR: https://github.com/graphql-python/graphene-django/pull/682/files . In graphene_django/filter/utils.py the way how to retrieve the Type of a field was changed.

Keep on rocking :)

zbyte64:
Attempt to fix #781 , here are somethings I didn't know when I started:

`filter.tests.test_filter_filterset_based_on_mixin` ( https://github.com/graphql-python/graphene-django/blob/def6b15e5bf6bb0129932b2286938a2fbb45cfca/graphene_django/filter/tests/test_fields.py#L873 ) demonstrates the ability to have the django-filter expression be based off a graphene method. In this example `viewer` is virtually mapped to an article as a reporter and is queried as `viewer__email__in`. This means we need to power our field lookup with `filter_field.field_name` and not the key name.

Another assumption violated is that only the last part specifies nonfield processing, it seems django can now chain some functions into the query, ie: `__lower__first3chars__reversed` https://docs.djangoproject.com/en/2.2/ref/models/lookups/#query-expression
I don't know how our system performs with that feature.

### DIFF
--- a/graphene_django/filter/tests/test_fields.py
+++ b/graphene_django/filter/tests/test_fields.py
@@ -877,7 +877,7 @@ def test_filter_filterset_based_on_mixin():
             filters = super(FilterSet, cls).get_filters()
             filters.update(
                 {
-                    "viewer__email__in": django_filters.CharFilter(
+                    "reporter__email__in": django_filters.CharFilter(
                         method="filter_email_in", field_name="reporter__email__in"
                     )
                 }
@@ -897,15 +897,10 @@ def test_filter_filterset_based_on_mixin():
             interfaces = (Node,)
 
     class NewArticleFilterNode(DjangoObjectType):
-        viewer = Field(NewReporterNode)
-
         class Meta:
             model = Article
             interfaces = (Node,)
             filterset_class = NewArticleFilter
-
-        def resolve_viewer(self, info):
-            return self.reporter
 
     class Query(ObjectType):
         all_articles = DjangoFilterConnectionField(NewArticleFilterNode)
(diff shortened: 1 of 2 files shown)


jkimbo:
Sorry for the late review on this @zbyte64 . Merging now.

Issue #815, opened 2019-11-28
DjangoFilterConnectionField ignoring DjangoObjectType's get_queryset


bellini666:
The PR #796 broke DjangoFilterConnectionField making it always get the raw queryset from the model to apply the filters in it.

I noted this when trying to update my extension (https://github.com/0soft/graphene-django-plus/tree/master/graphene_django_plus) to work with 2.7.0 and seeing some tests failing. There I override DjangoObjectType's get_queryset method to add some permissioning checks and my tests broke when a query that should only return some objects that the test user had access returned everything.

IMO this is a critical bug as people doing things like this (e.g. filtering objects for the current user) could potentially leak unwanted data.

I'm sending a PR at this moment and hope you can release a quick fix ASAP.

PR #816, opened 2019-11-28
Keep original queryset on DjangoFilterConnectionField

bellini666:
The PR #796 broke DjangoFilterConnectionField making it always get the
raw queryset from the model to apply the filters in it.

This makes sure that the DjangoObjectType's .get_queryset is called,
keeping any filtering it might have made.

====

I noted this when trying to update my extension (https://github.com/0soft/graphene-django-plus/tree/master/graphene_django_plus) to work with 2.7.0 and seeing some tests failing. There I override DjangoObjectType's get_queryset method to add some permissioning checks and my tests broke when a query that should only return some objects that the test user had access returned everything.

Fixes https://github.com/graphql-python/graphene-django/issues/815

### DIFF
--- a/graphene_django/filter/fields.py
+++ b/graphene_django/filter/fields.py
@@ -55,10 +55,11 @@ class DjangoFilterConnectionField(DjangoConnectionField):
     def resolve_queryset(
         cls, connection, iterable, info, args, filtering_args, filterset_class
     ):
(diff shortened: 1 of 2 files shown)


jkimbo:
Thanks for the quick PR @bellini666 . I've added a regression test and I'll release this as v2.7.1

PR #827, opened 2019-12-15
Add headers arg to GraphQLTestCase.query

### LINKED ISSUE #817: Add ability to set headers in GraphQLTestCases
When calling `self.query` within tests, there is currently no way to specify headers. It would be helpful to be able to specify headers in cases where, e.g., `GraphQLView` is overridden to check the `Authorization` header.

I've identified the line that needs to be updated:

https://github.com/graphql-python/graphene-django/blob/254e59c36fa289ddf86b32b528afeae54dfa1bb1/graphene_django/utils/testing.py#L56

### LINKED ISSUE #824: How to test APIs that require JWT?
I have been following the code as mentioned in the docs [here](https://docs.graphene-python.org/projects/django/en/latest/testing/) but its not working.I have the following issues/questions-

1. As Django carries testing in an ephemeral database, how can I create a user so that I can validate JWT auth in all the APIs? This is the error log of a test query
```
404                                                                                                                                                                                                                
functools.partial(<bound method Client._parse_json of <django.test.client.Client object at 0x7f0c87fb70f0>>, <HttpResponseNotFound status_code=404, "text/html">)
```

2. The `test_query_with_variables` function is not working and throws this error 
`TypeError: query() got an unexpected keyword argument 'variables'`
 
Please help with testing APIs that need JWT auth and also how to write test cases for them.

chibie:
Closes #817, closes #824 

Now headers can be passed to graphql URL

For example, authorization header can be set thus:

```
...
response = self.query(
    '''
    mutation updateUser($input: UserUpdateInput!) {
        updateUser(input: $input) {
            ok
            errors {
                field
                messages
            }
        }
    }
    ''',
    op_name='updateUser',
    input_data={
        'firstName': 'Dumebi'
    },
    headers={'HTTP_AUTHORIZATION': f'JWT {token}'}
)
```

### DIFF
--- a/graphene_django/utils/testing.py
+++ b/graphene_django/utils/testing.py
@@ -55,7 +55,10 @@ class GraphQLTestCase(TestCase):
                 body["variables"] = {"input": input_data}
         if headers:
             resp = self._client.post(
-                self.GRAPHQL_URL, json.dumps(body), content_type="application/json", **headers
+                self.GRAPHQL_URL,
+                json.dumps(body),
+                content_type="application/json",
+                **headers
             )
         else:
             resp = self._client.post(
--- a/graphene_django/utils/testing.py
+++ b/graphene_django/utils/testing.py
@@ -53,10 +53,14 @@ class GraphQLTestCase(TestCase):
                 body["variables"]["input"] = input_data
             else:
                 body["variables"] = {"input": input_data}
-
-        resp = self._client.post(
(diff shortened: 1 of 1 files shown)


nisayama:
I want this feature

alimirjahani7:
Why should we use HTTP_AUTHORIZATION instead of Authorization?

everdrone:
Second this, the `HTTP_AUTHORIZATION` is undocumented as far as i know.
Why use it instead of `Authorization`?

Issue #785, opened 2019-09-26
Enums are generating duplicated types


Tibiritabara:
I think I read this issue in the past, graphene-django is currently throwing the next exception on my end:

"Found different types with the same name in the schema: currency, currency"

The currency field is just a field with enums
```
class Test:
    CURRENCY = (
        ("EUR", "EUR"),
        ("USD", "USD"),
    )

   currency = models.CharField(
        "Operation currency",
        max_length=5,
        choices=CURRENCY,
        default=CURRENCY[0]
    )
```

There is a serializer for this model:
```
class TestSerializer(serializers.ModelSerializer):
    class Meta:
        model = Test
        fields = [
            'id',
            'currency',
        ]
```

The schema is using the `Test` model 

```
class TestNode(DjangoObjectType):
    class Meta:
        """Transaction GraphQL Type Metaclass
        Modifies the basic behavior of the type
        """
        model = Test
        filter_fields = {
            'currency': ['exact'],
        }
        interfaces = (node.Relay, )
```

and there are 2 mutations, one for creation based on the serializer and a custom one:
```
class CreateTestMutation(SerializerMutation):
    class Meta:
        serializer_class = TestSerializer
        model_operations = [
            'create',
        ]

class UpdateTestinput(InputObjectType):
    currency = graphene.String()

class UpdateTestMutation(graphene.Mutation):
    id = graphene.String()
    currency = graphene.String()

    class Arguments:
        """Arguments for the given mutation
        It defines which arguments are mandatory for the
        given mutation. The id is the one that will be used
        for searching and the input is the required object 
        for the update
        """
        id = graphene.String(required=True)
        input = UpdateTestInput(required=True)

    def mutate(self, info, id, input=None):
        test = Test.objects.get(pk=str(id))
        for key, value in input.items():
            setattr(transaction, key, value)
        test.save()
        # Notice we return an instance of this mutation
        return UpdateTestMutation(
            id=id,
            currency=transaction.currency,
        )
```

Now, this is currently throwing the next exception when using the latest release (2.6.0):
`Found different types with the same name in the schema: currency, currency.`

On 2.5.0 this error does not happen.

jl-DaDar:
i have exact same problem

begimai:
it is all because both DjangoObjectType and SerializerMutation are converting choice field to graphene Enum. I am struggling with the same error. 
Currently I see the easiest solution to move all Django Enums to Graphene Enums and resolve everything manually

[...]

PR #851, opened 2020-01-12
Adding support for disabling enum creation on SerializerMutation

leewardbound:
A workaround for #785 --

Appropriately, `DjangoObjectType` supports a `convert_choices_to_enum` which disables the creation of GraphQL `Enum` types for every `ChoiceField` on a model.

However, `SerializerMutation` (similarly a subclass of `ObjectType` and registered in the GQL schema) does not support such a flag, and forces creating an `Enum` for every `ChoiceField`. This behavior can be undesirable because it can trigger a duplicate type registration error `Found different types with the same name in the schema:`.

For consistency's sake, a flag to disable the `Enum` creation should be added to the `SerializerMutation`; this commit implements such changes.

Sidenote --
In my sandbox, this error actually triggered even when the ChoiceField was only registered one single time, on one single Mutation, and not at all present in the `DjangoObjectType`, which leads me to believe that the `SerializerMutation`'s `Enum` definition is colliding with itself?

There's actually 0% test coverage for the current existing behavior of creating `Enums` for ` SerializerMutation` because [none of these tests](https://github.com/graphql-python/graphene-django/tree/master/graphene_django/rest_framework/tests) actually implement any `Model` with a `ChoicesField`; the test coverage stops at ensuring the automatic type conversion returns a proper value (and I added a second test for my branching logic). I hesitate to suggest it, but it seems that in the current implementation of `Enums` in `SerializerMutations`, simply adding a SerializerMutation with any ChoicesField to a schema might be enough to throw the error?

I might have time to dig deeper into that, but for now, here's an effective and reasonable bandaid.

### DIFF
--- a/graphene_django/rest_framework/mutation.py
+++ b/graphene_django/rest_framework/mutation.py
@@ -19,7 +19,13 @@ class SerializerMutationOptions(MutationOptions):
     serializer_class = None
 
 
-def fields_for_serializer(serializer, only_fields, exclude_fields, is_input=False):
+def fields_for_serializer(
+    serializer,
+    only_fields,
+    exclude_fields,
+    is_input=False,
+    convert_choices_to_enum=True,
+):
     fields = OrderedDict()
     for name, field in serializer.fields.items():
         is_not_in_only = only_fields and name not in only_fields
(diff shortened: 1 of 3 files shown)


jkimbo:
Thanks @leewardbound !

Issue #843, opened 2019-12-31
AssertionError: Found different types with the same name in the schema: ...


arielnmz:
I know this has been submitted a million times, and I know there are several solutions BUT, despite all things it seems there still isn't a way to actually know *where* the duplicates come from, e.g. The error message would read:

    AssertionError: Found different types with the same name in the schema: status, status.

But that doesn't give much information as to where do they come from. This is specially frustrating when plugging in projects. A suggestion would be to print the canonical name of the class the type is based on? something like 'my_module.types.ProblematicType'.

Anyway, this issue arises specifically because I defined a Type based on an Enum on my types module and I'm using that everywhere, but still graphql complains about different types having the same name. I double checked and I don't even use that name anywhere in my project, so it must be a conflicting dependency, but which one? where is that type being defined? The only way to know would be unplugging all dependencies and start plugging them one by one to find the offenders but this would be so inefficient and frankly, very stupid.

(in case you wonder I also searched in my `site-packages` and found nothing, so it must be auto generated somewhere...)

I'm using these packages:
django==2.2.9
graphene-django==2.7.1
graphene==2.1.8

And my types worked just fine with these versions:
django==2.1.5
graphene-django==2.2.0
graphene==2.1.3

leewardbound:
Hi, I didn't want to open a new issue, but are you seeing this when using the `SerializerMutation`? If so, this might help --

I've isolated this bug in a test case -- I have verified that a simple test for adding a `SerializerMutation` with any simple `ChoiceField` causes the error to be thrown.

I think we can all agree that [this test should not fail on the last line](https://github.com/leewardbound/graphene-django/blob/tests/failing-simple-choices/graphene_django/rest_framework/tests/test_schema.py#L40), but it does.

I have also submitted a PR  #851 to allow us to disable this `Enum` creation, but I think we have positive confirmation here that these `Enum` types are causing schema definition collisions even in simple use cases.

adrengifo:
when is this going to be merged and pushed?

jkimbo:
Sorry for the delay everyone. #860 has just been released as v2.9.0 Go try it out!

ulgens:
@jkimbo I'm not sure #860 solves this issue. If the reason of naming clash was not choices but something else, the error message is still not helpful.

adrengifo:
still seeing the same issue with v2.9.0

jkimbo:
[...]

PR #860, opened 2020-01-29
Add options to override how Django Choice fields are converted to Enums

### LINKED ISSUE #185: Different types with the same name in the schema - Django fields with choices
Hi,

When you try to create a model with choices field and then create a second model which inherit from the previous one, you will receive error while schema generating.

Here you have a simple [project](https://github.com/KaRi94/graphene-enum-bug-example) with models and basic configuration. Try to generate schema and you will get following error:
```
AssertionError: Found different types with the same name in the schema: PageType, PageType.
```

It seems that it is unable to inherit models with choices fields.

jkimbo:
This PR resolves #185 and #843 by adding 2 new settings to customise the naming of the Enum types generated from Django Choice fields.

* `DJANGO_CHOICE_FIELD_ENUM_V3_NAMING` changes all enum type names to a naming convention that should prevent duplicates. This format will be the default in v3.
* `DJANGO_CHOICE_FIELD_ENUM_CUSTOM_NAME` allows you to define a custom function to determine what the enum type name should be

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -75,7 +75,7 @@ def generate_enum_name(django_model_meta, field):
     ):
         name = graphene_settings.DJANGO_CHOICE_FIELD_ENUM_CUSTOM_NAME(field)
     elif graphene_settings.DJANGO_CHOICE_FIELD_ENUM_V3_NAMING is True:
-        name = "DjangoModel{app_label}{object_name}{field_name}Choices".format(
+        name = "{app_label}{object_name}{field_name}Choices".format(
             app_label=to_camel_case(django_model_meta.app_label.title()),
             object_name=django_model_meta.object_name,
             field_name=to_camel_case(field.name.title()),
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -70,7 +70,7 @@ def convert_choices_to_named_enum_with_descriptions(name, choices):
 
 
 def generate_enum_name(django_model_meta, field):
-    if graphene_settings.CHOICES_TO_ENUM_UNIQUE_TYPE_NAME is True:
+    if graphene_settings.CHOICES_TO_ENUM_V3_NAMING is True:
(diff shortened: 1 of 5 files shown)


graphene_django/converter.py
```diff
@@ -68,6 +69,29 @@ def description(self):
     return Enum(name, list(named_choices), type=EnumWithDescriptionsType)
 
 
+def generate_enum_name(django_model_meta, field):
+    if graphene_settings.DJANGO_CHOICE_FIELD_ENUM_CUSTOM_NAME and callable(
+        graphene_settings.DJANGO_CHOICE_FIELD_ENUM_CUSTOM_NAME
+    ):
+        name = graphene_settings.DJANGO_CHOICE_FIELD_ENUM_CUSTOM_NAME(field)
+    elif graphene_settings.DJANGO_CHOICE_FIELD_ENUM_V3_NAMING is True:
+        name = "DjangoModel{app_label}{object_name}{field_name}Choices".format(
```
syrusakbary: I don't think prepending `DjangoModel` to the enum name is good, it leaks the storage to the API.

I think it will be better just:

```suggestion
        name = "{app_label}{object_name}{field_name}Choices".format(
```

jkimbo: Fair enough. Thanks @syrusakbary

jkimbo:
@zbyte64 good point about using module loading rather than direct references. I'll update the PR.

Also I don't understand the rest of your comment? Are you suggesting changes to the documentation or just re-iterating what the purpose of the change is?

zbyte64:
Was just re-iterating / thinking out loud. Updating the settings to allow a string specified import was my only recommend action.

jkimbo:
@zbyte64 PR is ready for another review

bennett-jacob:
When is this going to be merged?

renato-bvdash:
Someone merge and deploy this please

jkimbo:
Sorry for the delay with this (I've been on holiday). Shipping this now

PR #862, opened 2020-01-29
Only warn if a field doesn't exist on the Django model

jkimbo:
Also don't warn if the field name matches a custom field.

Resolves issues from #842

### DIFF
--- a/graphene_django/tests/test_types.py
+++ b/graphene_django/tests/test_types.py
@@ -320,26 +320,41 @@ def test_django_objecttype_fields_exclude_type_checking():
 
 @with_local_registry
 def test_django_objecttype_fields_exclude_exist_on_model():
-    with pytest.raises(Exception, match=r"Field .* doesn't exist"):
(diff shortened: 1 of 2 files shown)


zbyte64:
I understand not having a warning or error for custom fields. But I am less clear on the reasoning for not raising an exception for when there is no attribute on the model. Might be useful to state the consequence of each warning in the warning. For instance, "Field name .* exists on Django model .* but it's not a model field" - does this mean my fields wont be read? If there isn't a real consequence then I would drop the warning and embrace the feature.

jkimbo:
@zbyte64 the discussion is here: https://github.com/graphql-python/graphene-django/pull/842 . IMO switching to a warning rather than an exception is just to make the migration path a bit less severe for people. Good point about adding the consequence for each warning 👍 I'll update the messages.

jkimbo:
@zbyte64 I've expanded the error messages to be more descriptive.

PR #863, opened 2020-01-29
Allow for easier template overrides in graphiql

benhowes:
Use case: I use https://pypi.org/project/django-impersonate/ to allow testing my API as various users, and it's always handy to have a header which reminds you who you're impersonating.

By making graphiql populate a div rather than body, it's easier to override the template with additional content. Currently graphiql replaces all the content in `body` meaning any other content you add is also removed.

### DIFF
--- a/graphene_django/static/graphene_django/graphiql.js
+++ b/graphene_django/static/graphene_django/graphiql.js
@@ -97,6 +97,6 @@
   // Render <GraphiQL /> into the body.
   ReactDOM.render(
     React.createElement(GraphiQL, options),
-    document.getElementsByClassName("editor")[0]
+    document.getElementById("editor")
   );
 })();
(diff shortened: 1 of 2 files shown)


graphene_django/templates/graphene/graphiql.html
```diff
@@ -31,6 +31,7 @@
           crossorigin="anonymous"></script>
 </head>
 <body>
+  <div class="editor"></div>
```
jkimbo:
```suggestion
  <div id="editor"></div>
```

graphene_django/static/graphene_django/graphiql.js
```diff
@@ -97,6 +97,6 @@
   // Render <GraphiQL /> into the body.
   ReactDOM.render(
     React.createElement(GraphiQL, options),
-    document.body
+    document.getElementsByClassName("editor")[0]
```
jkimbo: Minor point but I think it's better to select by ID instead of class since there is only going to be 1 editor on a page.

```suggestion
    document.getElementById("editor")
```

benhowes:
@jkimbo no problem with that at all :)

benhowes:
Thanks @jkimbo! Sorry I forgot to update the CSS when copying changes from my overrides!

jkimbo:
No worries! Thanks for the contribution anyway @benhowes

PR #868, opened 2020-01-31
Detect format from output file path

jkimbo:
If the output filepath has a .graphql extension then export the schema in the graphql format.


jkimbo:
@zbyte64 PR ready for another review

PR #871, opened 2020-02-07
Allow defining fields as an empty list

### LINKED ISSUE #864: How to get `only_fields = ((),)` behavior? Goal: only shows fields declared in object type.
First: Apologies for this if it's not descriptive or covered by graphene-django. I actually think this is something that's requested in other issues, via #671 #97

When updating graphene-django from 2.4.0 -> 2.8.0, behavior with `only_fields = ((), )` changed

I inherited this as a code snippet and don't know where it's from.

What it used to do: Only fields that were in defined in the graphene object type would be shown.

Is there a way to get the equivalent to that?

Also, is there a way to get computed fields to also show via `fields = ()`?

Assume this example:

```python
from django.db import models

class Reporter(models.Model):
    first_name = models.CharField(max_length=30)
    last_name = models.CharField(max_length=30)
    email = models.EmailField()

    def website(self):
        return "https://www.google.com"
```

Before: Only showed model fields declared in object type
Now: 

`fields = ((),)` gives:

```
  File "lib/python3.7/site-packages/graphene/utils/subclass_with_meta.py", line 52, in __init_subclass__
    super_class.__init_subclass_with_meta__(**options)
  File "lib/python3.7/site-packages/graphene_django/types.py", line 174, in __init_subclass_with_meta__
    construct_fields(model, registry, fields, exclude, convert_choices_to_enum),
  File "lib/python3.7/site-packages/graphene_django/types.py", line 42, in construct_fields
    if hasattr(model, name):
```

Changing to `fields = []` will show all fields on model.

The problem with this is its very exhaustive maintain a blacklist of fields via `exclude`. Since the model lives elsewhere in the code, it's thought of a bit separately. (Aside: This issue also exists in `django.contrib.admin` itself.)

```python
class Reporter(DjangoObjectType):
    first_name = graphene.String(required=True, description='First name')
    class Meta:
        model = ReporterModel
        fields = []

# expected
fields = list(Reporter._meta.fields.keys())
assert fields == ["first_name"]
```

Before: Allowed computed fields
Now: `Exception: "my_computed field" exists on model <class 'mymodel.Reporter'> but it's not a field.`

```python
class Reporter(DjangoObjectType):
    website = graphene.String(required=True, description='First name')
    class Meta:
        model = ReporterModel
        fields = ['website']

# Expected
fields = list(Reporter._meta.fields.keys())
assert fields == ["website"]
```

jkimbo:
I think it makes sense that if the fields list is defined as an empty list on a DjangoObjectType then no model fields will be selected.

Resolves #864

### DIFF
--- a/graphene_django/tests/test_types.py
+++ b/graphene_django/tests/test_types.py
@@ -231,6 +231,17 @@ def test_django_objecttype_fields():
 
 
 @with_local_registry
+def test_django_objecttype_fields_empty():
+    class Reporter(DjangoObjectType):
+        class Meta:
+            model = ReporterModel
+            fields = ()
+
+    fields = list(Reporter._meta.fields.keys())
+    assert fields == []
+
+
+@with_local_registry
 def test_django_objecttype_only_fields_and_fields():
     with pytest.raises(Exception):
 
--- a/graphene_django/types.py
+++ b/graphene_django/types.py
@@ -35,9 +35,15 @@ def construct_fields(
 
     fields = OrderedDict()
     for name, field in _model_fields:
-        is_not_in_only = only_fields and name not in only_fields
+        is_not_in_only = (
+            only_fields is not None
+            and only_fields != ALL_FIELDS
+            and name not in only_fields
+        )
         # is_already_created = name in options.fields
-        is_excluded = name in exclude_fields  # or is_already_created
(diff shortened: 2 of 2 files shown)


tony:
@jkimbo Nice!  Can you rebase this by chance?

PR #873, opened 2020-02-08
Check exclude fields correctly

jkimbo:
v2.8.1 introduced a bug where exclude fields would not be validated properly. This PR fixes that by checking exclude fields separately.

Ref: #842


fulaphex:
@zbyte64 When do you plan to merge it and release a new version?

jkimbo:
@fulaphex sorry for the delay. v2.8.2 is going out now

PR #878, opened 2020-02-14
Use actual root/variables/context args of the execute schema function

GeyseR:
Since the version 2.3.0, graphql-core switched to the old arguments for the `execute` function for execution queries to schema. This PR should fix a bunch of deprecation warnings from this method.


jkimbo:
Thanks @GeyseR !

PR #879, opened 2020-02-14
Avoid: AssertionError at /graphql

nietzscheson:
This avoids the confusing error of: AssertionError at /graphql

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -59,9 +59,10 @@ The most basic ``schema.py`` looks like this:
 .. code:: python
 
     import graphene
+    from graphene_django.debug import DjangoDebug
 
     class Query(graphene.ObjectType):
-        pass
+        debug = graphene.Field(DjangoDebug, name="_debug")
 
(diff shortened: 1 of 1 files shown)


radekwlsk:
I would rather see the error be more informative and explain in the docs why this happens and what you have to do o actually get minimal working example.

Debug is also rather bad choice as it may be easy to later forget to disable that in production for less experienced people. Also the DjangoDebug [exists for very specific reason](http://docs.graphene-python.org/projects/django/en/latest/debug/#django-debug-middleware).

Wouldn't it be better to add simplest 'Hello World!' field on query with a resolver?

jkimbo:
I agree with @radekwlsk . I think doing something similar to what is shown in the Graphene tutorial would be better: https://docs.graphene-python.org/en/latest/quickstart/#creating-a-basic-schema

jkimbo:
This has been resolved in https://github.com/graphql-python/graphene-django/pull/994

PR #886, opened 2020-02-27
Use the Django TestCase's Client

tnightingale:
Use the `django.test.Client` instance that Django provides with its `TestCase` class.

This allows GraphQL tests to make use of the stateful client methods like `client.login()`.


ulgens:
ping @jkimbo

graphene_django/utils/testing.py
```diff
@@ -1,6 +1,6 @@
 import json
 
-from django.test import TestCase, Client
+from django.test import Client, TestCase
```
ulgens, line 3: After the change below, is that `Client` class used anywhere?

tnightingale:
```suggestion
from django.test import TestCase
```

ulgens:
@tnightingale @zbyte64 Any change we rebase & merge this?

zbyte64:
Looks like `graphene_django/tests/test_utils.py` needs to be updated to reference `GraphqlTestCase`? Seeing `ModuleNotFoundError: No module named 'graphene_django.utils.testing.Client'; 'graphene_django.utils.testing' is not a package`

Not sure how master is passing but this does not...

ulgens:
This PR can be closed, in favor of https://github.com/graphql-python/graphene-django/pull/1084

PR #891, opened 2020-03-04
refactor how _convert_choices_to_enum is calculated.

a-c-sreedhar-reddy:


jkimbo:
@a-c-sreedhar-reddy is this PR fixing a bug? If it is can you create a test for it?

a-c-sreedhar-reddy:
@jkimbo It is not a fix for a bug. It makes code more readable.

jkimbo:
I don't agree, I actually think the change makes the code less readable. Generally splitting out the logic into multiple steps is more readable than using ternarys.

Sorry I'm not going to accept this change. I'm happy to accept changes that make the code easier to reason about but I don't think this change does that.

jkimbo:
Also I don't want to discourage you from contributing @a-c-sreedhar-reddy , I just don't think this change is a good one.

PR #899, opened 2020-03-11
Adding Django filter as an installed app

RoySegall:
While trying to figure out how to filter results I came across the reference to the cookbook but there was missing a section that shows that we need to register the app we installed (Django-filters).

Hope this will be useful for me as for others as well.

### DIFF
--- a/examples/cookbook/cookbook/settings.py
+++ b/examples/cookbook/cookbook/settings.py
@@ -41,6 +41,7 @@ INSTALLED_APPS = [
     "graphene_django",
     "cookbook.ingredients.apps.IngredientsConfig",
     "cookbook.recipes.apps.RecipesConfig",
+    "django_filters",
 ]
 
 MIDDLEWARE = [


radekwlsk:
@RoySegall It may be a good idea to also mention it in the [docs on Filtering](http://docs.graphene-python.org/projects/django/en/latest/filtering/#filtering). Docs mention cookbook itself, but not everyone is going to check that out.

RoySegall:
> but not everyone is going to check that out

Well.. I did 😁. Any way, the docs you mentioned located in [here](https://github.com/graphql-python/graphene/tree/master/docsP), right?

radekwlsk:
> Well.. I did 😁. Any way, the docs you mentioned located in [here](https://github.com/graphql-python/graphene/tree/master/docsP), right?

https://github.com/graphql-python/graphene-django/blob/master/docs/filtering.rst

RoySegall:
10x! I'll update it later on this week.

RoySegall:
@jkimbo Will do, on my todo list - right after Passover. Have a great day and stay at home!

PR #900, opened 2020-03-11
Debug

joerhodes3:


joerhodes3:
Allow DEBUG by curl or graphiql by passing Auth in the header

PR #932, opened 2020-04-14
Make tests order independent

ganwell:
* Reset the global registry after each test (teardown)

* Create a settings fixtures that returns graphene_settings and resets the graphene_settings after use (teardown)

* Convert test_mutation tests from unittests.TestCase to pytest

* Convert test_mutation PetType to a pet_type fixtures that reregisters the type

The core change is setup/teardown in conftest.py

https://github.com/graphql-python/graphene-django/blob/a16cebea7c2fb2bb33a5c677e96ad8a0cfe8100d/graphene_django/conftest.py#L8-L18

### DIFF
--- a/graphene_django/conftest.py
+++ b/graphene_django/conftest.py
@@ -1,0 +1,18 @@
+import pytest
+
+from graphene_django.settings import graphene_settings as gsettings
+
+from .registry import reset_global_registry
+
+
+@pytest.fixture(autouse=True)
+def reset_registry_fixture(db):
+    yield None
+    reset_global_registry()
+
+
+@pytest.fixture()
+def graphene_settings():
+    settings = dict(gsettings.__dict__)
+    yield gsettings
+    gsettings.__dict__ = settings
--- a/graphene_django/forms/tests/test_mutation.py
+++ b/graphene_django/forms/tests/test_mutation.py
@@ -1,14 +1,23 @@
+import pytest
 from django import forms
-from django.test import TestCase
 from django.core.exceptions import ValidationError
 from py.test import raises
 
-from graphene import ObjectType, Schema, String, Field
+from graphene import Field, ObjectType, Schema, String
 from graphene_django import DjangoObjectType
-from graphene_django.tests.models import Film, Pet
+from graphene_django.tests.models import Pet
 
-from ...settings import graphene_settings
 from ..mutation import DjangoFormMutation, DjangoModelFormMutation
+
+
+@pytest.fixture()
+def pet_type():
+    class PetType(DjangoObjectType):
+        class Meta:
+            model = Pet
+            fields = "__all__"
+
+    return PetType
 
 
 class MyForm(forms.Form):
@@ -36,18 +45,6 @@ class PetForm(forms.ModelForm):
         return age
 
 
-class PetType(DjangoObjectType):
-    class Meta:
-        model = Pet
-        fields = "__all__"
-
-
-class FilmType(DjangoObjectType):
-    class Meta:
-        model = Film
-        fields = "__all__"
-
-
 def test_needs_form_class():
     with raises(Exception) as exc:
 
@@ -73,7 +70,7 @@ def test_has_input_fields():
     assert "text" in MyMutation.Input._meta.fields
 
 
-def test_mutation_error_camelcased():
+def test_mutation_error_camelcased(pet_type, graphene_settings):
     class ExtraPetForm(PetForm):
         test_field = forms.CharField(required=True)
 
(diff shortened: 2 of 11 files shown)


ganwell:
I will follow-up with the reset of the failures (removing the fixed seed) once this is reviewed.

ganwell:
I will leave the python2.7 failure till an initial review. Actually I propose to remove python2.7.

jkimbo:
@ganwell if you need to drop Python 2.7 support then why not create this PR against the v3 branch?

ganwell:
@jkimbo I rebased to v3 and removed WIP

graphene_django/conftest.py
```diff
@@ -0,0 +1,13 @@
...
+
+from .registry import reset_global_registry
+
+
+@pytest.fixture(autouse=True)
+def reset_registry_fixture(db):
+    settings = dict(graphene_settings.__dict__)
+    yield None
+    reset_global_registry()
+    graphene_settings.__dict__ = settings
```
jkimbo, line 13: Why do you have to restore the graphene_settings dict?

graphene_django/forms/tests/test_mutation.py
```diff
@@ -1,51 +1,57 @@
...
 
 from ...settings import graphene_settings
 from ..mutation import DjangoFormMutation, DjangoModelFormMutation
 
 
-class MyForm(forms.Form):
-    text = forms.CharField()
+@pytest.fixture()
+def my_form():
+    class MyForm(forms.Form):
```
jkimbo, line 16: Why do the forms need to be fixtures?

graphene_django/forms/tests/test_mutation.py
```diff
@@ -93,227 +99,231 @@ class MockQuery(ObjectType):
...
+
+    # A pet was not created
+    Pet.objects.count() == 0
+
+    fields_w_error = [e.field for e in result.errors]
+    assert len(result.errors) == 2
+    assert result.errors[0].messages == ["This field is required."]
+    assert result.errors[1].messages == ["This field is required."]
+    assert "age" in fields_w_error
+    assert "name" in fields_w_error
```
jkimbo, line 329: Much prefer plain pytest tests rather than Django's TestCase 👍

pytest.ini
```diff
@@ -1,2 +1,3 @@
 [pytest]
 DJANGO_SETTINGS_MODULE = django_test_settings
+addopts = --random-order
```
jkimbo, line 3: Pytest config shouldn't be defined in multiple places. I would remove the `pytest.ini` file and just define it in the `setup.cfg` file.

graphene_django/forms/tests/test_mutation.py
ganwell: @jkimbo I think only the types have to be fixtures, because they get registered in the global registry through some magic when python executes the class. I put everything into fixtures just for consistency.

If you like I can check what needs to be fixtures and only more these to fixtures.

jkimbo: I think that would be cleaner. As far as I'm aware only the Django models and Graphene-Django DjangoObjectType's register themselves into a registry.

graphene_django/conftest.py
ganwell: There are tests that change the settings and don't change it back.

There are several options:

1. Provide a graphene_settings fixture (probably the best way). This fixture resets the settings after use

2. Just reset the settings on every test, so it is impossible to leak any settings. I decided to use this approach because it is best for an environment where contributors often change.

3. Manually reset the settings in the tests

We usually use 1. but we have regular contributors. Let me know what you prefer.

Uses of graphene_settings:

```
$> git grep graphene_settings | grep '/tests/'
graphene_django/forms/tests/test_mutation.py:from ...settings import graphene_settings
graphene_django/forms/tests/test_mutation.py:    graphene_settings.CAMELCASE_ERRORS = True
graphene_django/forms/tests/test_mutation.py:    graphene_settings.CAMELCASE_ERRORS = False
graphene_django/rest_framework/tests/test_mutation.py:from ...settings import graphene_settings
graphene_django/rest_framework/tests/test_mutation.py:    graphene_settings.CAMELCASE_ERRORS = True
graphene_django/rest_framework/tests/test_mutation.py:    graphene_settings.CAMELCASE_ERRORS = False
graphene_django/tests/test_converter.py:from ..settings import graphene_settings
graphene_django/tests/test_converter.py:    graphene_settings.DJANGO_CHOICE_FIELD_ENUM_V3_NAMING = True
graphene_django/tests/test_converter.py:    graphene_settings.DJANGO_CHOICE_FIELD_ENUM_V3_NAMING = False
graphene_django/tests/test_query.py:from ..settings import graphene_settings
graphene_django/tests/test_query.py:    graphene_settings.RELAY_CONNECTION_ENFORCE_FIRST_OR_LAST = True
graphene_django/tests/test_query.py:    graphene_settings.RELAY_CONNECTION_MAX_LIMIT = 100
graphene_django/tests/test_query.py:    graphene_settings.RELAY_CONNECTION_ENFORCE_FIRST_OR_LAST = False
graphene_django/tests/test_query.py:    graphene_settings.RELAY_CONNECTION_MAX_LIMIT = 100
graphene_django/tests/test_query.py:    graphene_settings.RELAY_CONNECTION_ENFORCE_FIRST_OR_LAST = False
graphene_django/tests/test_types.py:from ..settings import graphene_settings
graphene_django/tests/test_types.py:        graphene_settings.DJANGO_CHOICE_FIELD_ENUM_V3_NAMING = True
graphene_django/tests/test_types.py:        graphene_settings.DJANGO_CHOICE_FIELD_ENUM_V3_NAMING = False
graphene_django/tests/test_types.py:        graphene_settings.DJANGO_CHOICE_FIELD_ENUM_CUSTOM_NAME = (
graphene_django/tests/test_types.py:        graphene_settings.DJANGO_CHOICE_FIELD_ENUM_CUSTOM_NAME = None
```

jkimbo: I think a fixture would be good as well. If it's going to be too much work to change it all though I'm ok leaving it like this.

ganwell:
@jkimbo I think I addressed all the issues.

ganwell:
@jkimbo I will split this PR into one that can be merge to master and the rest that can only be merged to v3. In about 2 hours.

jkimbo:
If you feel it's important @ganwell . I'm fine with just merging this into v3

ganwell:
@jkimbo I moved this PR to master again. It now contains everything except adding pytest-random-order and can therefore be merged to master.

ganwell:
@jkimbo Here is the part that only works on v3 11e1403 / #940

Sorry for making it so complicated I felt it is the right thing to do in case it takes longer to merge v3 to master.

I will checkin on #932 and #940 the **next days**. In case you think it is all too complicated just do what you think is best: ie checking it in yourself.

graphene_django/conftest.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@pytest.fixture(autouse=True)
+def reset_registry_fixture(db):
+    yield None
+    reset_global_registry()
+
+
+@pytest.fixture()
+def settings():
```
jkimbo, line 15:
```suggestion
def graphene_settings():
```

Otherwise this will clash with the pytest django settings fixture: https://pytest-django.readthedocs.io/en/latest/helpers.html#settings

ganwell:
@jkimbo, I changed the fixture settings() to graphene_settings()

When you merge, could you please also merge master into v3 again, so I can update #940? It currently includes a merge, which makes it difficult to read the diff.

jkimbo:
@ganwell v3 updated

Issue #933, opened 2020-04-15
DjangoFormMutation should allow filtering of input and output separately


jtratner:
`DjangoFormMutation` by default filters both input and output fields based on `only_fields` and `exclude_fields`.  Ideally it'd allow you to filter input and output separately. (I'm going to put up a PR alongside this, but figured it was best form to have an issue too).

h3. Desired Behavior vs. Actual Behavior

**Desired Behavior**: Mutation that has a payload with only `clientMutationId` + my custom output field, while still having all input fields
**Actual Behavior**: either have to filter out all input fields (making form unusable) 

h3. Background and Example

<details><summary>Imports + model definition + form definition</summary>
<pre><code>
from graphene_django import DjangoObjectType
from graphene import relay, ObjectType, Int
from graphene_django.forms.types import ErrorType
from graphene_django.forms.mutation import DjangoFormMutation
from spar_web.core.models import Report
from django import forms
from django.db import models
import graphene

class Report(models.Model):
    display_name = models.CharField(max_length=255)
    config = models.CharField(max_length=255)
    directory = models.CharField(max_length=255)

class CreateReportForm(forms.Form):
    display_name = forms.CharField()
    report_type = forms.CharField()
    directory = forms.CharField()
    config = forms.CharField(required=False)

    def clean(self):
        if self.errors:
            return
        cleaned_data = super().clean()
        if not cleaned_data.get('config'):
            # dummy example
            cleaned_data['config'] = cleaned_data['report_type'] + 'config'
        return cleaned_data
</pre></code>
</details>

```
class ReportNode(DjangoObjectType):
    class Meta:
        model = Report
        interfaces = (relay.Node,)


class CreateReportMutation(DjangoFormMutation):
    report = graphene.Field(ReportNode)

    class Meta:
        form_class = Create
        only_fields = ['report']

    @classmethod
    def perform_mutate(cls, form, info):
        data = form.cleaned_data
        obj = Report.objects.create(
            display_name=data['display_name'],
            directory=data['directory'],
            config=data['config']
        )
        kwargs = {'report': obj}
        return cls(errors=[], **kwargs)


class Query(graphene.ObjectType):
    report = relay.Node.Field(ReportNode)


class Mutation(graphene.ObjectType):
    create_report = CreateReportMutation.Field()

schema = graphene.Schema(
    query=Query,
    mutation=Mutation
)
```

jtratner:
(pushed PR with example in #934 :) )

[...]

PR #934, opened 2020-04-15
Update django form mutation to more granularly handle input/output fields

jtratner:
Example of how I'd like to change DjangoFormMutation as per #933. I think this should be backwards compatible with the existing one.  Check out docstring on `DjangoFormMutation` for an overview of what the changes (ought to!) do :)

I also tried to add some docs about class behavior.

This still needs a test case and probably to be merged into the public graphene django docs but I figured this'd be good basis for discussion :)

Ultimately this is motivated by me having desire to be able to have a graphql mutation that just returns `clientMutationId` and a single object even tho I'm not using a `DjangoModelFormMutation`.

### DIFF
--- a/graphene_django/forms/tests/test_mutation.py
+++ b/graphene_django/forms/tests/test_mutation.py
@@ -1,9 +1,10 @@
 import pytest
 from django import forms
+from django.test import TestCase
 from django.core.exceptions import ValidationError
 from py.test import raises
 
-from graphene import Field, ObjectType, Schema, String
+from graphene import Field, Int, ObjectType, Schema, String
 from graphene_django import DjangoObjectType
 from graphene_django.tests.models import Pet
 
@@ -22,12 +23,16 @@ def pet_type():
 
 class MyForm(forms.Form):
     text = forms.CharField()
+    another = forms.CharField(required=False)
 
     def clean_text(self):
         text = self.cleaned_data["text"]
         if text == "INVALID_INPUT":
             raise ValidationError("Invalid input")
         return text
+
+    def clean_another(self):
+        self.cleaned_data["another"] = self.cleaned_data["another"] or "defaultvalue"
 
     def save(self):
         pass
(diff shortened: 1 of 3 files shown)


jkimbo:
Thanks @jtratner ! I get the use case for this but I'm just wondering if this is the best API for it because it seems quite verbose. Could we simply it by just having 2 options: `input_fields` and `output_fields` where each is a list of fields or the string `__all__` to select all fields? Is there much benefit to being able to specify `only` and `exclude`?

Related: for DjangoObjectType's we are planning on removing the `only_fields` and `exclude_fields` attributes in favour of `only` and `exclude` attributes: https://github.com/graphql-python/graphene-django/pull/691 We should probably maintain that naming convention elsewhere in the library as well.

jtratner:
@jkimbo - totally agree that it’s wordy 😂😅  I was trying to follow the existing convention.

Honestly if I were designing this class from the start I prob would have users explicitly specify output fields for the mutation. I’m not clear on a use case where you’d want to post the fields from `cleaned_data` back. Then you’d only need `only` and `include` to manage the input fields.

If you wanted to do that string, I’d suggest defining a constant

```
import graphene_django

...
Meta:
    input = graphene_django.ALL_FIELDS
    output = ()
```

jkimbo:
I think that api makes more sense @jtratner so let's move towards that implementation. We can first start by preferring `input_fields` (or maybe just `input`) and `output_fields` options and emitting deprecation warnings if `only_fields` or `exclude_fields` are defined. Then in a later release we can drop `only_fields` and `exclude_fields` completely and also raise exceptions if `input_fields` or `output_fields` are not defined. I would follow a similar logic to https://github.com/graphql-python/graphene-django/pull/691

Don't have a strong opinion on defining a constant verses just writing out `__all__`. They can be both be valid. (The `__all__` came from DjangoRestFramework)

jtratner:
What about getting rid of the automatic output fields in a backwards compatible way:

1. `only_fields` and `exclude_fields` keep the existing behavior (and warns about deprecation)
2. No fields specified keeps existing behavior (deprecation warning about required to specify)
3. Specifying `fields` or `exclude` will be required in future version. , If specified in current version, it means that only output field is `clientMutationId`. `exclude=graphene_django.ALL_FIELDS` means no input fields at all

This would make the API much simpler

jkimbo:
> Specifying fields or exclude will be required in future version (they only modify input fields). If specified in current version, it means that only output field is clientMutationId. exclude=graphene_django.ALL_FIELDS means no input fields at all.

`DjangoFormMutations` has an `errors` field defined on the output type which should stay there. Also I think we should just drop the `clientMutationId` field completely because it's no longer necessary for relay or any other frontend client.

I think I like the idea of defaulting to no output fields which means you have to be very explicit about what outputs you want. Unfortunately I don't have many usecases for DjangoFormMutation so I'm not that familiar with what people use it form. What kind of things are you using it form @jtratner ?

jtratner:
(See #933 for a simplifies version of our use case)

graphene_django/forms/tests/test_mutation.py
```diff
@@ -150,6 +183,54 @@ class Mutation(ObjectType):
...
+        assert "some_integer" in MyMutation._meta.fields
+
+        class Mutation(ObjectType):
+            my_mutation = MyMutation.Field()
+
+        schema = Schema(query=MockQuery, mutation=Mutation)
+
+        result = schema.execute(
+            """ mutation MyMutation {
+                myMutation(input: { text: "VALID_INPUT" }) {
```
jtratner, line 217: @jkimbo - I am super confused why this test errors with `text not a valid keyword argument` when the assertion statements above are true. Any thoughts?

```
>       self.assertIs(result.errors, None)
E       AssertionError: [GraphQLLocatedError("'text' is an invalid keyword argument for MyMutation")] is not None
```

graphene_django/forms/tests/test_mutation.py
```diff
@@ -68,6 +73,83 @@ class Meta:
...
+    assert "some_integer" in MyMutation._meta.fields
+
+    class Mutation(ObjectType):
+        my_mutation = MyMutation.Field()
+
+    schema = Schema(query=MockQuery, mutation=Mutation)
+
+    result = schema.execute(
+        """ mutation MyMutation {
+            myMutation(input: { text: "VALID_INPUT" }) {
```
jtratner, line 137: @jkimbo - I am super confused why this test errors with `text not a valid keyword argument` when the assertion statements above are true. Any thoughts?

```
>       assert result.errors is None
E       assert [GraphQLLocatedError("'text' is an invalid keyword argument for MyMutation")] is None
E        +  where [GraphQLLocatedError("'text' is an invalid keyword argument for MyMutation")] = <graphql.execution.base.ExecutionResult object at 0x103d25af8>.errors```

jtratner:
@jkimbo - circling back to this after a while :) Has your API break completed now? (I _think_ that was a blocker in the past).  If so, is there a set way of naming that you'd like me to use?  If not, I can try to get back in the zone for this PR and finish it up.

PR #940, opened 2020-04-19
Make tests order independent (v3)

ganwell:
* Add pytest-random-order to change test-order on each pytest-run

Contains the part that can only be merged on v3


ganwell:
@jkimbo CI is ok and the PR is now trivial.

PR #958, opened 2020-05-11
Convert Django form / DRF decimals to Graphene decimals

### LINKED ISSUE #91: Convert DecimalField to String rather than Float
DecimalField entries are currently being converted to Float (https://github.com/graphql-python/graphene-django/blob/8f5a425859ea380d0640c441c5c7f3b1acae17b4/graphene_django/form_converter.py#L56-L59)

Full support for Decimal would be great, but in the meantime I propose that DecimalField should be converted to String instead. Once you've converted a Decimal to a Float, you've probably lost the benefit of why you used Decimal in the first place, since you'll now have values like `3.14000000000000012`, rather than `3.14`. If it is converted to String, it is simple enough for the user to convert that String back to Decimal, whereas converting from Float to Decimal is not straightforward.

ulgens:
Fixes #91 

(Branch name is outdated. I can create a new PR if it's necessary.)

### DIFF
--- a/graphene_django/filter/tests/test_fields.py
+++ b/graphene_django/filter/tests/test_fields.py
@@ -671,7 +671,7 @@ def test_should_query_filter_node_limit():
     schema = Schema(query=Query)
     query = """
         query NodeFilteringQuery {
-            allReporters(limit: 1) {
+            allReporters(limit: "1") {
                 edges {
                     node {
                         id
--- a/graphene_django/filter/tests/test_fields.py
+++ b/graphene_django/filter/tests/test_fields.py
@@ -5,7 +5,7 @@ import pytest
 from django.db.models import TextField, Value
 from django.db.models.functions import Concat
 
-from graphene import Argument, Boolean, Field, Float, ObjectType, Schema, String
+from graphene import Argument, Boolean, Decimal, Field, ObjectType, Schema, String
 from graphene.relay import Node
 from graphene_django import DjangoObjectType
 from graphene_django.forms import GlobalIDFormField, GlobalIDMultipleChoiceField
@@ -388,7 +388,7 @@ def test_filterset_descriptions():
     field = DjangoFilterConnectionField(ArticleNode, filterset_class=ArticleIdFilter)
     max_time = field.args["max_time"]
     assert isinstance(max_time, Argument)
-    assert max_time.type == Float
+    assert max_time.type == Decimal
     assert max_time.description == "The maximum time"
 
 
--- a/graphene_django/forms/converter.py
+++ b/graphene_django/forms/converter.py
@@ -1,11 +1,22 @@
 from django import forms
 from django.core.exceptions import ImproperlyConfigured
 
-from graphene import ID, Boolean, Float, Int, List, String, UUID, Date, DateTime, Time
+from graphene import (
+    Boolean,
+    Date,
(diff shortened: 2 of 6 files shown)


zbyte64:
#91 is fixed, is this still relevant? What's this about `test_should_query_filter_node_limit` ?

ulgens:
@zbyte64 I don't think this is still relevant, i'll test and update it asap. The issue with the test, as far as i remember, registering a decimal as string affected incoming data type in pagination, paginator was expecting a decimal value (not sure how and why).

ulgens:
@zbyte64 This is still relevant. #1083 doesn't touch Django form based conversions at all. I'll try to make it work with that new `graphene.Decimal` type.

graphene_django/filter/tests/test_fields.py
```diff
@@ -671,7 +671,7 @@ def resolve_all_reporters(self, info, **args):
     schema = Schema(query=Query)
     query = """
         query NodeFilteringQuery {
-            allReporters(limit: 1) {
+            allReporters(limit: "1") {
```
ulgens, line 674: @zbyte64 I couldn't understand why this change is necessary and i can use some help :slightly_smiling_face: 

The filter here uses `NumberFilter` from django-filters and it expects a decimal input. `Decimal(1) == Decimal("1")` is true but when that input is numeric 1, the test fails with 

```
FAILED graphene_django/filter/tests/test_fields.py::test_should_query_filter_node_limit - assert not [GraphQLError('Argument "limit" has invalid value 1.\nExpected type "Decimal", found 1.')]
```

zbyte64: I experimented with adding the following above line 671: 

```
assert Query.all_reporters.filtering_args["limit"].type is Decimal, str(
        Query.all_reporters.filtering_args["limit"].type
    )
 ```
 
 Interestingly this line did not fail. So we know we're generating the right graphene type at least.

zbyte64: It is easier to get a stack trace with v3 than v2, still working on this :/

zbyte64: Something interesting happened when I branched and rebased off v3, that particular error went away but then it stopped applying the limit, `filter_limit` no longer gets called. BUT if I then switch the form conversion back to float the method gets called by django filter. Digging further the args being passed into `DjangoFilterConnectoinField.resolve_queryset` as `limit` as `None` when we convert to Decimal but not Float.

zbyte64: Traced the issue to graphene, I'll be sending them a PR in a moment

zbyte64: This should fix the issue we were seeing: https://github.com/graphql-python/graphene/pull/1295

jkimbo (approved):
LGTM thanks @ulgens 

(this should probably be part of the v3 release since it's a breaking change)

zbyte64:
I'm scared to set rebase against v3 but I think @jkimbo is right, the decimal fix is for graphene v3

ulgens:
@zbyte64 Why did we just merged it? :slightly_smiling_face: graphene didn't release a v2 version with decimal fix. I was intended to revert 1 -> "1" after that fix.

PR #966, opened 2020-05-20
Error in signature of callproc() and execute()

### LINKED ISSUE #586: NormalCursorWrapper has invalid sigatures for callproc and execute
/debug/sql/tracking.py          

The callproc and execute methods have an incorrect signature. This is causing a substitution attempt in the standard code, which breaks if there are '%'s in the SQL function.

    def callproc(self, procname, params=()):
          return self._record(self.cursor.callproc, procname, params)
  
    def execute(self, sql, params=()):
          return self._record(self.cursor.execute, sql, params)

pauricthelodger:
Fixes #586

The default params `()` in the two `NormalCursorWrapper` methods should be `None` to match what django expects in the code block below. A bare `cursor.execute(sql)` will have `params=()` added to it by the Wrapper which fails to match the `if` in the code block below, causing the IndexError. 

```
    def execute(self, sql, params=None):
        self.db.validate_no_broken_transaction()
        with self.db.wrap_database_errors:
            if params is None:
                return self.cursor.execute(sql)
            else:
>               return self.cursor.execute(sql, params)
E               IndexError: tuple index out of range

django/db/backends/utils.py:65: IndexError
```

This fix was done @kevingill1966 in their fork, I'm just making a PR to have it added =)

### DIFF
--- a/graphene_django/debug/sql/tracking.py
+++ b/graphene_django/debug/sql/tracking.py
@@ -148,10 +148,10 @@ class NormalCursorWrapper(object):
             # We keep `sql` to maintain backwards compatibility
             self.logger.object.sql.append(_sql)
 
-    def callproc(self, procname, params=()):
+    def callproc(self, procname, params=None):
         return self._record(self.cursor.callproc, procname, params)
 
-    def execute(self, sql, params=()):
+    def execute(self, sql, params=None):
         return self._record(self.cursor.execute, sql, params)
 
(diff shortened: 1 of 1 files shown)


pauricthelodger:
Example of a similar fix I did for the same issue in Scout a few years back, seems a lot of people did it the same way! =) https://github.com/scoutapp/scout_apm_python/pull/43

jkimbo:

PR #974, opened 2020-05-26
Header functionality

joerhodes3:
enable passing a JWT AUTH token in the header METAdata for a graphql query......also enables curl use.


jkimbo:
@joerhodes3 I don't understand the need for this change. Can you explain what your usecase is?

Also I don't think including the graphiql static resources as part of this PR is necessary. Can you remove them so that this PR only contains the minimum changes?

joerhodes3:
fair, however I do need to be static as I'm in a disconnected env.  That
was my motivation for adding them to the static/ dir by default, but if
you'd like we can stay cdn for now and I will just keep a local copy of the
libs.

My goal is to get most or all of what I need into the main branch, so I'm
happy to make whatever adjustments needed.

Thanks,
~Joe

joerhodes3:
sorry I missed the usecase part.

We have an app split into several microservices.  The parts of the app
communicate by passing a JWT token around in the META data of the header.
This allow debuging with curl by passing it that same JWT token.

This is that whole graphql isn't responsible for the auth, the transport
mechanism is.  This is a long standing request, and was our/my way of
addressing (bring our graphql in line with or ReST)

ulgens:
I, too, am missing the point in this PR. We are using the library for last 8 months and jwt token auth over header was the very first thing we implemented. It is working flawlessly till this day.

joerhodes3:
how?  when it came time to debug (either graphiql or curl) the old auth was
not passed along without a session.  We have nothing but the BEARER in the
settings, so Django is not doing any auth, this is just passed to the
simplejwt to be evaluated right before execution. (this was the way auth
was handled for or ReST api and allowed us to move forward with a mix or
ReST and graphql endpoints that only executed if authorized).

I did stumble across one alternate method of sign in for django on an
external site, but could not get that to play with a mix of old and new.  I
could not find any recipe here.

That said I fine trying some other solution if explained.

zbyte64:
Typically authorization is handled with a combination of middleware instead of adding header logic at the view level. Example: https://django-graphql-jwt.domake.io/en/latest/quickstart.html#installation

Or if using DRF: https://django-rest-framework-simplejwt.readthedocs.io/en/latest/getting_started.html#installation

Implementation aside, I think this is out of scope. As we see with DRF (and already with graphene-django) the trend is to extend authorization with 3rd party libraries and not implement our own login views.

joerhodes3:
thanks.  the existing app was down DRF path, but not the
django-graphql-jwt.  I will see if that gives me what I need.

PR #977, opened 2020-06-05
Fix filtering with GlobalIDFilter

### LINKED ISSUE #542: Can't use GlobalIDFilter as FilterSet field
I noticed that when I was trying to use a `GlobalIDFilter` as a FilterSet field, it doesn't call the filter method (both my custom method `filter_table_id` and `GlobalIDFilter.filter` is not called) and returns all of the records in the table. If I use a CharFilter instead of a GlobalIDFilter, it works and filters as expected.

This works. But type is `String` in queries.
```python
class ModelFilter(FilterSet):
    def filter_table_id(self, queryset, name, value):
        _, _id = from_global_id(value)
        return queryset.filter(table1__table2__table3__id=_id)

    table_id = CharFilter(method='filter_table_id')

    class Meta:
        model = DBModel
        fields = {
            'name': ['exact', 'icontains'],
            'table_id': ['exact'],
        }
```

But if I replace the `CharFilter` with `GlobalIDFilter`, the type is `ID` in the query, but all the records are returned, none are filtered out.
```python
class ModelFilter(FilterSet):
    def filter_table_id(self, queryset, name, value):
        _, _id = from_global_id(value)
        return queryset.filter(table1__table2__table3__id=_id)

    table_id = GlobalIDFilter(method='filter_table_id')

    class Meta:
        model = DBModel
        fields = {
            'name': ['exact', 'icontains'],
            'table_id': ['exact'],
        }
```

I tried to see why it was not working, but I could not find out why the filter function nor my `filter_table_id` was being called.
Am I using the wrong thing for filtering?
Is there a better way to do what I want?

I am using venv with:
 * Python 3.7.0 (default, Sep 12 2018, 18:30:08), the default Python3.7 Ubuntu 18.04 package.
 * graphene-django==2.2.0
 * django-filter==2.0.0
 * graphene==2.1.3

### LINKED ISSUE #808: Using exact filter on Relay with ForeignKey model
I found a problem when I try to filter a model with ForeignKey. I expect to get the result below if the ID is incorrect.
```
{
    "data": {
        "forms": {
            "edges": []
        }
    }
}
```
However, I get every single row in my table returned if the value of ID is incorrect. Filtering works as expected if the ID exists.

```
# models
class Province(models.Model):
    name = models.CharField(max_length=50)

class City(models.Model):
    province = models.ForeignKey(Province,
                                 on_delete=models.CASCADE,
                                 related_name='city')
    name = models.CharField(max_length=50)
```

```
# schema
class ProvinceNode(DjangoObjectType):
    class Meta:
        model = Province
        filter_fields = {'name': ['exact']}
        interfaces = (relay.Node, )


class CityNode(DjangoObjectType):
    class Meta:
        model = City
        filter_fields = {'province': ['exact']}
        interfaces = (relay.Node, )

cities = DjangoFilterConnectionField(queries.CityNode)
```

```
# query
{
  cities(province: "UHJvdmluY2VOb2RlOjE3") {
    edges {
      node { id name province { id name } }
    }
  }
}
```

hubertsiuzdak:
resolves #808, resolves #542

Filtering with not valid global ID doesn't raise any exception, and the query set without filter applied is returned instead. That happens due to the fact that django-filter no longer performs error handling itself (see carltongibson/django-filter#1078). This PR solves the issue by validating the filterset in DjangoFilterConnectionField.

### DIFF
--- a/graphene_django/filter/fields.py
+++ b/graphene_django/filter/fields.py
@@ -1,6 +1,7 @@
 from collections import OrderedDict
 from functools import partial
 
+from django.core.exceptions import ValidationError
 from graphene.types.argument import to_arguments
 from ..fields import DjangoConnectionField
 from .utils import get_filtering_args_from_filterset, get_filterset_class
@@ -59,7 +60,12 @@ class DjangoFilterConnectionField(DjangoConnectionField):
             connection, iterable, info, args
         )
         filter_kwargs = {k: v for k, v in args.items() if k in filtering_args}
-        return filterset_class(data=filter_kwargs, queryset=qs, request=info.context).qs
+        filterset = filterset_class(
+            data=filter_kwargs, queryset=qs, request=info.context
+        )
+        if filterset.form.is_valid():
+            return filterset.qs
+        raise ValidationError(filterset.form.errors.as_json())
 
     def get_queryset_resolver(self):
         return partial(
--- a/graphene_django/filter/tests/test_fields.py
+++ b/graphene_django/filter/tests/test_fields.py
@@ -400,6 +400,114 @@ def test_global_id_field_relation():
     assert id_filter.field_class == GlobalIDFormField
 
 
+def test_global_id_field_relation_with_filter():
+    class ReporterFilterNode(DjangoObjectType):
+        class Meta:
+            model = Reporter
+            interfaces = (Node,)
+            filter_fields = ["first_name", "articles"]
+
+    class ArticleFilterNode(DjangoObjectType):
+        class Meta:
+            model = Article
+            interfaces = (Node,)
+            filter_fields = ["headline", "reporter"]
+
+    class Query(ObjectType):
+        all_reporters = DjangoFilterConnectionField(ReporterFilterNode)
+        all_articles = DjangoFilterConnectionField(ArticleFilterNode)
+        reporter = Field(ReporterFilterNode)
+        article = Field(ArticleFilterNode)
+
+    r1 = Reporter.objects.create(first_name="r1", last_name="r1", email="r1@test.com")
+    r2 = Reporter.objects.create(first_name="r2", last_name="r2", email="r2@test.com")
+    Article.objects.create(
+        headline="a1",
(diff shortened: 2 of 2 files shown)


graphene_django/filter/fields.py
```diff
@@ -59,7 +61,13 @@ def resolve_queryset(
...
         )
         filter_kwargs = {k: v for k, v in args.items() if k in filtering_args}
-        return filterset_class(data=filter_kwargs, queryset=qs, request=info.context).qs
+        filterset = filterset_class(
+            data=filter_kwargs, queryset=qs, request=info.context
+        )
+        if filterset.is_valid():
+            return filterset.qs
+        else:
+            raise ValidationError(json.dumps(filterset.errors.get_json_data()))
```
jkimbo, line 70:
```suggestion
        raise ValidationError(json.dumps(filterset.errors.get_json_data()))
```

No need for the `else` statement.

graphene_django/filter/fields.py
```diff
@@ -59,7 +60,12 @@ def resolve_queryset(
             connection, iterable, info, args
         )
         filter_kwargs = {k: v for k, v in args.items() if k in filtering_args}
-        return filterset_class(data=filter_kwargs, queryset=qs, request=info.context).qs
+        filterset = filterset_class(
+            data=filter_kwargs, queryset=qs, request=info.context
+        )
+        if filterset.form.is_valid():
+            return filterset.qs
+        raise ValidationError(filterset.form.errors.as_json())
```
hubertsiuzdak, line 68: To support Python 2.7 and django-filter 1.1 that `form` property is needed. In Graphene-Django v3 we use django-filter 2.x which allows to call `is_valid` and `errors` directly on filterset instances. We may want to refactor that for v3 as it slightly improves readability.

hubertsiuzdak:
@jkimbo Can you take a look again?

jkimbo:
Thanks @hubertsiuzdak

jkimbo:

PR #981, opened 2020-06-07
Warn if `fields` or `exclude` are not defined on `DjangoObjectType`

### LINKED ISSUE #710: Start warning if `fields` or `exclude` are not defined on `DjangoObjectType`
So that model fields aren't accidentally exposed through DjangoObjectType I propose that we start warning if either `fields` or `exclude` aren't defined with the intention to error completely in the future. This would also align the API more with Django Rest Framework which hopefully makes it more familiar to most developers.

radekwlsk:
To make use of Django-Graphene more familiar to folks using DRF ~`PendingDeprecationWarning`~ `DeprecationWarning` (edit: changed in [805480](https://github.com/graphql-python/graphene-django/pull/981/commits/8054803fde619c820a16cd855c8471d2ee62f1a1)) is issued when `DjangoObjectType` has neither `fields` nor `exclude` option set.

Current deprecations of `exclude_fields` and `only_fields` are taken into account as this check happens after they are copied over to `fields` or `exclude`.

All tests, examples and docs are also updated so that they won't issue this warning. Documentation mentions that warning in [docs/queries.rst#specifying-which-fields-to-include](https://github.com/radekwlsk/graphene-django/blob/topic-710/docs/queries.rst#specifying-which-fields-to-include) and instructs how to avoid it.

Change is part of v3 release discussed in #705. 

Closes #710

### DIFF
--- a/docs/authorization.rst
+++ b/docs/authorization.rst
@@ -111,6 +111,7 @@ If you are using ``DjangoObjectType`` you can define a custom `get_queryset`.
     class PostNode(DjangoObjectType):
         class Meta:
             model = Post
+            fields = '__all__'
 
         @classmethod
         def get_queryset(cls, queryset, info):
--- a/docs/filtering.rst
+++ b/docs/filtering.rst
@@ -45,6 +45,7 @@ For example:
         class Meta:
             # Assume you have an Animal model defined with the following fields
             model = Animal
+            fields = '__all__'
             filter_fields = ['name', 'genus', 'is_domesticated']
             interfaces = (relay.Node, )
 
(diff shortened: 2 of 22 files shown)


radekwlsk:
@jkimbo let me know if the tests and docs changes are expected here or if I should just leave it throwing warnings and don't mention in docs until it is `DeprecationWarning` instead of `PendingDeprecationWarning`.

radekwlsk:
@jkimbo Warning level bumped to `DeprecationWarning` and I've also updated the information in docs to be have stronger tone about this soft requirement.

PR #986, opened 2020-06-09
Fix hasNextPage - revert to count. Fix after

pcraciunoiu:
This is a follow up fix from #965.

Unfortunately, that broke the logic for `has_next_page` (which is still... confusing anyways).

This fixes it to perform a count again, since we need the total length to decide if there is a next page. Except for the new test and the changes in `fields.py`, the other test changes are a revert, essentially.

This should make it possible to do infinite scroll style pagination on `DjangoConnectionField`s now!

### DIFF
--- a/graphene_django/fields.py
+++ b/graphene_django/fields.py
@@ -133,21 +133,23 @@ class DjangoConnectionField(ConnectionField):
     def resolve_connection(cls, connection, args, iterable, max_limit=None):
         iterable = maybe_queryset(iterable)
 
-        max_limit = max_limit or 0
         if isinstance(iterable, QuerySet):
             list_length = iterable.count()
-            list_slice_length = max(max_limit, list_length)
+            list_slice_length = (
+                min(max_limit, list_length) if max_limit is not None else list_length
+            )
         else:
             list_length = len(iterable)
-            list_slice_length = max(max_limit, list_length)
+            list_slice_length = (
+                min(max_limit, list_length) if max_limit is not None else list_length
+            )
 
         after = get_offset_with_default(args.get("after"), -1) + 1
-        list_slice_length += after
 
         connection = connection_from_list_slice(
-            iterable,
+            iterable[after:],
             args,
-            slice_start=0,
+            slice_start=after,
             list_length=list_length,
             list_slice_length=list_slice_length,
             connection_type=connection,
--- a/graphene_django/fields.py
+++ b/graphene_django/fields.py
@@ -2,7 +2,10 @@ from functools import partial
 
 import six
 from django.db.models.query import QuerySet
-from graphql_relay.connection.arrayconnection import connection_from_list_slice
+from graphql_relay.connection.arrayconnection import (
+    connection_from_list_slice,
+    get_offset_with_default,
+)
 from promise import Promise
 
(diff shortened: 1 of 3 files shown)


graphene_django/tests/test_query.py
```diff
@@ -1126,6 +1126,44 @@ class Query(graphene.ObjectType):
...
+            model = Reporter
+            interfaces = (Node,)
+
+    class Query(graphene.ObjectType):
+        all_reporters = DjangoConnectionField(ReporterType)
+
+    schema = graphene.Schema(query=Query)
+    # Need first: 4 here to trigger the `has_next_page` logic in graphql-relay
+    # See `arrayconnection.py::connection_from_list_slice`:
+    # has_next_page=isinstance(first, int) and end_offset < upper_bound
```
pcraciunoiu, line 1145: IMO this is a bug in `graphql-relay`. Without `first` param, you get `has_next_page=False` even if there are more elements... Sigh.

graphene_django/fields.py
```diff
@@ -129,25 +132,29 @@ def resolve_queryset(cls, connection, queryset, info, args):
...
-            _len = max_limit or iterable.count()
+            list_length = iterable.count()
+            list_slice_length = max_limit or list_length
         else:
-            _len = max_limit or len(iterable)
+            list_length = len(iterable)
+            list_slice_length = max_limit or list_length
+
+        after = get_offset_with_default(args.get("after"), -1) + 1
+        list_slice_length += after
```
pcraciunoiu, line 144: Without this, `graphql-relay` v2 slices an empty list here:

https://github.com/graphql-python/graphql-relay-py/blob/54fa42716e587c21e1fd8e6fc7d39cb32db28039/graphql_relay/connection/arrayconnection.py#L79-L82

Might be fixed in 3.0, but I didn't try that since it might be a rabbit hole.

pcraciunoiu:
@jkimbo thoughts on this?

jkimbo:
@pcraciunoiu sorry for the delay with this. Looking into it now.

jkimbo:
OK @pcraciunoiu I think I've found the solution. You were right you should use `min` but it looks like the `slice_start` and the iterable needs to be modified before I can get all the tests to pass. I've created a commit with the changes here: https://github.com/graphql-python/graphene-django/commit/81eb81ab5ceac4dc400b4313e496317a8c0b8a44 (I couldn't push the changes to your branch because I don't think you've given maintainers permission to edit it)

If you can either let me push to your branch or cherry pick that commit then we can get this merged.

pcraciunoiu:
I tried following [this](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/allowing-changes-to-a-pull-request-branch-created-from-a-fork), but I don't see a button to allow. I cherry-picked exactly what you had and left a comment, but I think this makes sense.

jkimbo:
@pcraciunoiu I fixed the issue you've mentioned here: https://github.com/graphql-python/graphene-django/pull/986#discussion_r437631002 : https://github.com/graphql-python/graphene-django/pull/993

pcraciunoiu:
@jkimbo awesome! Thanks for taking this one step further!

Issue #985, opened 2020-06-08
Wrong supported python versions in PyPi metadata


AndreyMZ:
According to [the test configuration](https://github.com/graphql-python/graphene-django/blob/c00203499b00d40a696a092ab3c64759eddad0f8/.github/workflows/tests.yml#L10-L15) it looks like graphene-django:
1. Supports Python 3.7, 3.8.
2. Does not support Python 3.4, 3.5.

But [the metadata](https://github.com/graphql-python/graphene-django/blob/c00203499b00d40a696a092ab3c64759eddad0f8/setup.py#L48-L54) of the [`graphql-python`](https://pypi.org/project/graphene-django/) project on PyPi differs.

**What is the current behavior?**

https://pypi.org/project/graphene-django/

> Programming Language
> * Python :: 2
> * Python :: 2.7
> * Python :: 3
> * Python :: 3.4
> * Python :: 3.5
> * Python :: 3.6
> * Python :: Implementation :: PyPy

**What is the expected behavior?**

> Framework
> * Django
> * Django :: 1.11
> * Django :: 2.2
> * Django :: 3.0
>
> Programming Language
> * Python :: 2
> * Python :: 2.7
> * Python :: 3
> * Python :: 3.6
> * Python :: 3.7
> * Python :: 3.8
> * Python :: Implementation :: PyPy

PR #987, opened 2020-06-10
Update setup.py classifiers

jkimbo:
Fixes https://github.com/graphql-python/graphene-django/issues/985


jkimbo:

PR #992, opened 2020-06-24
Move to_const function from Graphene into Graphene-Django

jkimbo:
This PR moves the `to_const` string converter function from the Graphene project into the Graphene-Django one because it is only used in the Graphene-Django project. See: https://github.com/graphql-python/graphene/issues/1208

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -20,13 +20,14 @@ from graphene import (
     Time,
 )
 from graphene.types.json import JSONString
-from graphene.utils.str_converters import to_camel_case, to_const
+from graphene.utils.str_converters import to_camel_case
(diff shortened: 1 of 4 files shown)


jkimbo:

noelleleigh:
Can this patch also be applied to the v3 branch? I'm getting an import error after installing graphene-django 3.0.0b3, which installed graphene 3.0.0b3:

```
File "/usr/local/lib/python3.6/dist-packages/graphene_django/converter.py", line 25, in <module>
    from graphene.utils.str_converters import to_camel_case, to_const
ImportError: cannot import name 'to_const'
```

jkimbo:
@noelleleigh just released v3.0.0b4 that includes this change. Sorry about that.

PR #996, opened 2020-06-26
Move to_const function from Graphene into Graphene-Django (#992)

ganwell:


ganwell:
cherry-pick #992 to v3

sbhatm1213:
I still get the import error for `to_const`. I have `graphene-django==2.0.0` .

Issue #998, opened 2020-07-01
"Negative indexing is not supported" after changes to relay


bellini666:
After the changes to relay connection I started noticing this issue in my project.

It happens when passing `after` with a higher number than the iterable's total length.

The fix is very simple and I'm already working on a PR which a should submit in some minutes.

```
Traceback (most recent call last):
  File "/home/bellini/dev/cliqueimudei/cliqueimudei/backend/cm/urls.py", line 53, in execute_graphql_request
    raise error.original_error
  File "/home/bellini/.virtualenvs/cm/lib/python3.8/site-packages/promise/promise.py", line 489, in _resolve_from_executor
    executor(resolve, reject)
  File "/home/bellini/.virtualenvs/cm/lib/python3.8/site-packages/promise/promise.py", line 756, in executor
    return resolve(f(*args, **kwargs))
  File "/home/bellini/.virtualenvs/cm/lib/python3.8/site-packages/graphql/execution/middleware.py", line 75, in make_it_promise
    return next(*args, **kwargs)
  File "/home/bellini/.virtualenvs/cm/lib/python3.8/site-packages/graphene_django/fields.py", line 216, in connection_resolver
    return on_resolve(iterable)
  File "/home/bellini/.virtualenvs/cm/lib/python3.8/site-packages/graphene_django/fields.py", line 153, in resolve_connection
    connection = connection_from_list_slice(
  File "/home/bellini/.virtualenvs/cm/lib/python3.8/site-packages/graphql_relay/connection/arrayconnection.py", line 82, in connection_from_list_slice
    _slice = list_slice[
  File "/home/bellini/.virtualenvs/cm/lib/python3.8/site-packages/django/db/models/query.py", line 290, in __getitem__
    assert ((not isinstance(k, slice) and (k >= 0)) or
AssertionError: Negative indexing is not supported.
```

PR #999, opened 2020-07-01
Do not break when after is greater than list_length

bellini666:
I added a test that, without the change do `graphene_django/fields.py` breaks with the traceback described in the issue bellow.

Fix https://github.com/graphql-python/graphene-django/issues/998

### DIFF
--- a/graphene_django/fields.py
+++ b/graphene_django/fields.py
@@ -144,7 +144,10 @@ class DjangoConnectionField(ConnectionField):
                 min(max_limit, list_length) if max_limit is not None else list_length
             )
 
-        after = get_offset_with_default(args.get("after"), -1) + 1
+        # If after is higher than list_length, connection_from_list_slice
+        # would try to do a negative slicing which makes django throw an
+        # AssertionError
+        after = min(get_offset_with_default(args.get("after"), -1) + 1, list_length)
 
         if max_limit is not None and "first" not in args:
             args["first"] = max_limit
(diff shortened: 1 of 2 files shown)


jkimbo:
Thanks for the raising this issue @bellini666 . I'm wondering if it would be better if an error was raised (one that it is more helpful) rather than silently accepting an `after` value which is greater than the list length. What's the situation where your client is passing an `after` value that is greater than the list length?

bellini666:
Hi @jkimbo .

Yeah, I thought of that too, that probably the client should be more careful when passing an after keyword. But on the other hand, if we think of this in terms of python slicing or even sql limit/offset, if you pass something an offset greater than the list/result's length will be an empty list/set and will be silently regarding the "issue".

Another way to look at this is to do what the relay specifications (or the original node implementation) does. If it says to raise an error then we should definitely raise an error =P

wkoot:
I don't think it would be weird to return an empty set as it is the default behaviour for filtering in general.
Say I filter `field_Icontains: "non_existant"` I would also recieve an empty result, without errors :)
Perhaps it would be useful to return such warnings in a meta field though. Is there anything about it in the spec?

jkimbo:
Sorry for the delay @bellini666 . I've been thinking about this a lot and I agree with yours and @wkoot argument so I'm going to merge this. 

> Perhaps it would be useful to return such warnings in a meta field though. Is there anything about it in the spec? 
As far as I know there isn't anything in the spec about this.

jkimbo:
Released this in v2.11.1

PR #1002, opened 2020-07-12
GraphiQL cleanup

eabruzzese:
I went back through my GraphiQL upgrade changes and cleaned up a bit:

* Upgrade the `window.fetch` polyfill to the latest version.
* Add [subresource integrity (SRI) hashes](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) for all CDN-loaded assets.
* Run the `graphiql.js` file through [`prettier`](https://prettier.io/) (using prettier's default settings).
* Fix an issue where the `SUBSCRIPTION_PATH` value wasn't making it all the way to the view.

### DIFF
--- a/graphene_django/static/graphene_django/graphiql.js
+++ b/graphene_django/static/graphene_django/graphiql.js
@@ -1,43 +1,55 @@
 (function (
   document,
+
   GRAPHENE_SETTINGS,
   GraphiQL,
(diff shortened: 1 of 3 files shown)


jkimbo:
Yes thanks @eabruzzese !

PR #1004, opened 2020-07-13
Remove operation name from the regex and default to query

### LINKED ISSUE #1003: GraphiQL not working in v2.12.0
  
In version v2.12.0 the GraphiQL playground no longer works. I always get the error `Cannot read property '1' of null` in my response. In version v2.11.1 everything works fine.

nicememe ZbVai8zFifldc0UdeSZ

  - Version: graphene-django v2.12.0; graphene v2.1.8
  - Platform: Linux Ubuntu 20.04

jkimbo:
Fixes #1003

### DIFF
--- a/graphene_django/static/graphene_django/graphiql.js
+++ b/graphene_django/static/graphene_django/graphiql.js
@@ -135,7 +135,9 @@
     // Run a regex against the query to determine the operation type (query, mutation, subscription).
     var operationRegex = new RegExp(
       // Look for lines that start with an operation keyword, ignoring whitespace.
-      "^\\s*(query|mutation|subscription)\\s+" +
+      "^\\s*(query|mutation|subscription)\\s*" +
+        // The operation keyword should be followed by whitespace and the operationName in the GraphQL parameters (if available).
+        (graphQLParams.operationName ? ("\\s+" + graphQLParams.operationName) : "") +
         // The line should eventually encounter an opening curly brace.
         "[^\\{]*\\{",
       // Enable multiline matching.
--- a/graphene_django/static/graphene_django/graphiql.js
+++ b/graphene_django/static/graphene_django/graphiql.js
@@ -136,14 +136,15 @@
     var operationRegex = new RegExp(
       // Look for lines that start with an operation keyword, ignoring whitespace.
       "^\\s*(query|mutation|subscription)\\s+" +
-        // The operation keyword should be followed by the operationName in the GraphQL parameters.
-        graphQLParams.operationName +
         // The line should eventually encounter an opening curly brace.
         "[^\\{]*\\{",
       // Enable multiline matching.
       "m",
     );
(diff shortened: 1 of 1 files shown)


eabruzzese:
@jkimbo this will break if there are multiple operations being sent. For example:

```graphql
query FirstOperation {
  ...
}

subscription SecondOperation {
  ...
}
```

When you click the button to execute the operation, it'll have you choose the `operationName` from the dropdown. If you choose `SecondOperation`, the regex in this PR will match `query`, and won't send it over the websocket -- causing the operation to fail.

Let me take a look at that issue.

jkimbo:
Good point @eabruzzese . Could check if `operationName` is defined and only use it in the regex if it is? Also I found this in the graphiql examples: https://github.com/graphql/graphiql/blob/main/examples/graphiql-cdn-subscriptions/index.html#L61-L74

But it requires the `parse` method which I assume comes from graphql. That might be a more robust option though.

eabruzzese:
@jkimbo yeah, I saw that and tried to incorporate the parser, but couldn't get it working because of the format of the graphql CDN package (it also seemed like overkill for just detecting the operation type).

I like your idea there: if there's no operation name, just don't include it in the regex. That should solve most cases for now. I'll try to incorporate the parser again tonight.

eabruzzese:
@jkimbo I think you could just change `graphQLParams.operationName +` to `(graphQLParams.operationName || '') +` to get it working, yeah?

graphene_django/static/graphene_django/graphiql.js
```diff
@@ -136,14 +136,15 @@
     var operationRegex = new RegExp(
       // Look for lines that start with an operation keyword, ignoring whitespace.
       "^\\s*(query|mutation|subscription)\\s+" +
-        // The operation keyword should be followed by the operationName in the GraphQL parameters.
-        graphQLParams.operationName +
         // The line should eventually encounter an opening curly brace.
         "[^\\{]*\\{",
       // Enable multiline matching.
       "m",
     );
```
eabruzzese, line 143:
```suggestion
    var operationRegex = new RegExp(
      // Look for lines that start with an operation keyword, ignoring whitespace.
      "^\\s*(query|mutation|subscription)\\s*" +
        // The operation keyword should be followed by whitespace and the operationName in the GraphQL parameters (if available).
        (graphQLParams.operationName ? ("\\s+" + graphQLParams.operationName) : '') +
        // The line should eventually encounter an opening curly brace.
        "[^\\{]*\\{",
      // Enable multiline matching.
      "m",
    );
```

jkimbo: Thanks @eabruzzese . Made the change.

eabruzzese: Gonna try it out quick in my project; one moment.

eabruzzese:
@jkimbo Just tested with this branch. Works as expected.

PR #1016, opened 2020-07-31
Add headers support to GraphiQL

### LINKED ISSUE #850: Add custom header support to GraphiQL
Related issue: https://github.com/graphql/graphiql/issues/500

GraphiQL by default does not implement a way to pass custom request headers, which are still fairly commonly used for example by authentication middleware. It would be nice if graphene-django's built-in GraphiQL view had support for doing authenticated requests with custom HTTP headers.

radekwlsk:
* add GRAPHIQL_HEADER_EDITOR_ENABLED setting with False default
* use GRAPHIQL_HEADER_EDITOR_ENABLED to set headerEditorEnabled GraphiQL option
* update graphQLFetcher and httpClient in static/graphiql.js to support custom query/mutation headers

Making use of https://github.com/graphql/graphiql/pull/1543 this PR introduces ability to set custom headers in GraphiQL once `GRAPHIQL_HEADER_EDITOR_ENABLED` setting is changed to `True`.

This should resolve #850

Probably some docs for this functionality have to be added (and maybe a test).

### DIFF
--- a/graphene_django/settings.py
+++ b/graphene_django/settings.py
@@ -41,6 +41,10 @@ DEFAULTS = {
     "DJANGO_CHOICE_FIELD_ENUM_CUSTOM_NAME": None,
     # Use a separate path for handling subscriptions.
     "SUBSCRIPTION_PATH": None,
+    # Set to True to enable GraphiQL headers editor tab
+    # This sets headerEditorEnabled GraphiQL option, for details go to
+    # https://github.com/graphql/graphiql/tree/main/packages/graphiql#options
+    "GRAPHIQL_HEADER_EDITOR_ENABLED": False,
 }
 
 if settings.DEBUG:
--- a/graphene_django/settings.py
+++ b/graphene_django/settings.py
@@ -41,10 +41,10 @@ DEFAULTS = {
     "DJANGO_CHOICE_FIELD_ENUM_CUSTOM_NAME": None,
     # Use a separate path for handling subscriptions.
     "SUBSCRIPTION_PATH": None,
-    # Set to True to enable GraphiQL headers editor tab
+    # By default GraphiQL headers editor tab is enabled, set to False to hide it
     # This sets headerEditorEnabled GraphiQL option, for details go to
     # https://github.com/graphql/graphiql/tree/main/packages/graphiql#options
-    "GRAPHIQL_HEADER_EDITOR_ENABLED": False,
+    "GRAPHIQL_HEADER_EDITOR_ENABLED": True,
 }
 
 if settings.DEBUG:
--- a/graphene_django/static/graphene_django/graphiql.js
+++ b/graphene_django/static/graphene_django/graphiql.js
@@ -61,17 +61,29 @@
   var fetchURL = locationQuery(otherParams);
 
   // Defines a GraphQL fetcher using the fetch API.
-  function httpClient(graphQLParams) {
-    var headers = {
-      Accept: "application/json",
-      "Content-Type": "application/json",
-    };
+  function httpClient(graphQLParams, opts = { headers: {} }) {
+    let headers = opts.headers;
+    // Convert headers to an object.
+    if (typeof headers === 'string') {
+      headers = JSON.parse(opts.headers);
+    }
     if (csrftoken) {
-      headers["X-CSRFToken"] = csrftoken;
+      Object.assign(
+          {
+            'X-CSRFToken': csrftoken
+          },
+          headers,
+      )
     }
     return fetch(fetchURL, {
       method: "post",
-      headers: headers,
+      headers: Object.assign(
+          {
+            'Accept': 'application/json',
+            'Content-Type': 'application/json'
+          },
+          headers,
+      ),
       body: JSON.stringify(graphQLParams),
       credentials: "include",
     })
@@ -108,7 +120,7 @@
   var activeSubscription = null;
 
   // Define a GraphQL fetcher that can intelligently route queries based on the operation type.
-  function graphQLFetcher(graphQLParams) {
+  function graphQLFetcher(graphQLParams, opts = { headers: {} }) {
     var operationType = getOperationType(graphQLParams);
 
     // If we're about to execute a new operation, and we have an active subscription,
@@ -126,7 +138,7 @@
         },
       };
     } else {
-      return httpClient(graphQLParams);
+      return httpClient(graphQLParams, opts);
     }
   }
 
@@ -173,6 +185,7 @@
     onEditQuery: onEditQuery,
     onEditVariables: onEditVariables,
     onEditOperationName: onEditOperationName,
+    headerEditorEnabled: GRAPHENE_SETTINGS.graphiqlHeaderEditorEnabled || false,
     query: parameters.query,
   };
   if (parameters.variables) {
--- a/graphene_django/static/graphene_django/graphiql.js
+++ b/graphene_django/static/graphene_django/graphiql.js
@@ -175,7 +175,7 @@
     onEditQuery: onEditQuery,
     onEditVariables: onEditVariables,
     onEditOperationName: onEditOperationName,
-    headerEditorEnabled: GRAPHENE_SETTINGS.graphiqlHeaderEditorEnabled || true,
+    headerEditorEnabled: GRAPHENE_SETTINGS.graphiqlHeaderEditorEnabled,
     query: parameters.query,
   };
   if (parameters.variables) {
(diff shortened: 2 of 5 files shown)


graphene_django/static/graphene_django/graphiql.js
```diff
@@ -61,17 +61,29 @@
   var fetchURL = locationQuery(otherParams);
 
   // Defines a GraphQL fetcher using the fetch API.
-  function httpClient(graphQLParams) {
-    var headers = {
-      Accept: "application/json",
-      "Content-Type": "application/json",
-    };
+  function httpClient(graphQLParams, opts = { headers: {} }) {
+    let headers = opts.headers;
```
KubaJastrz, line 65: It's rather safe to assume that this file won't be transpiled into es5-compatible version during the build process, so I'd suggest not to use default parameters (shipped in es6). Same for `let`, let's use `var` as it was before.

```suggestion
  function httpClient(graphQLParams, opts) {
    if (typeof opts === 'undefined') {
      opts = {};
    }
    var headers = opts.headers || {};
```

graphene_django/static/graphene_django/graphiql.js
```diff
@@ -61,17 +61,29 @@
...
+    // Convert headers to an object.
+    if (typeof headers === 'string') {
+      headers = JSON.parse(opts.headers);
+    }
```
KubaJastrz, line 69: I'm not sure where the `string` type comes from, as in `graphiql` definitions, `headers` can only be an object ([source](https://unpkg.com/browse/graphiql@1.0.3/esm/components/GraphiQL.d.ts)):

```ts
export declare type FetcherParams = {
  query: string;
  operationName: string;
  variables?: string;
};
export declare type FetcherOpts = {
  headers?: {
    [key: string]: any;
  };
  shouldPersistHeaders: boolean;
};
export declare type FetcherResult = {
  data: IntrospectionQuery;
} | string | {
  data: any;
};
export declare type Fetcher = (graphQLParams: FetcherParams, opts?: FetcherOpts) => Promise<FetcherResult> | Observable<FetcherResult>;
```

graphene_django/static/graphene_django/graphiql.js
```diff
@@ -61,17 +61,29 @@
...
     if (csrftoken) {
-      headers["X-CSRFToken"] = csrftoken;
+      Object.assign(
+          {
+            'X-CSRFToken': csrftoken
+          },
+          headers,
+      )
```
KubaJastrz, line 76: `Object.assign` does 2 things:
1. mutates the `target` (first param) object with `source` (second param) by doing a shallow comparison
2. returns the mutated object (target)

So in this case, you'd mutate an object literal and then return it, but the returned value is not used in any way, so the whole operation is essentially useless. I think this change is rather unnecessary in what you want to achieve.

Also, `Object.assign` is available from es6+, so you can't use that here probably anyway.

graphene_django/static/graphene_django/graphiql.js
```diff
@@ -61,17 +61,29 @@
...
     return fetch(fetchURL, {
       method: "post",
-      headers: headers,
+      headers: Object.assign(
+          {
+            'Accept': 'application/json',
+            'Content-Type': 'application/json'
+          },
+          headers,
+      ),
```
KubaJastrz, line 86: Similar issue with es6+. You can avoid it by mutating headers object earlier in code:
```js
headers['Accept'] = headers['Accept'] || 'application/json';
headers['Content-Type'] = headers['Content-Type'] || 'application/json';
```

graphene_django/static/graphene_django/graphiql.js
```diff
@@ -108,7 +120,7 @@
   var activeSubscription = null;
 
   // Define a GraphQL fetcher that can intelligently route queries based on the operation type.
-  function graphQLFetcher(graphQLParams) {
+  function graphQLFetcher(graphQLParams, opts = { headers: {} }) {
```
KubaJastrz, line 123: no need for default params here, if the `httpClient` function does all validation anyway

radekwlsk:
@KubaJastrz I've applied your suggestions in 29f28bc

graphene_django/settings.py
```diff
@@ -41,6 +41,10 @@
     "DJANGO_CHOICE_FIELD_ENUM_CUSTOM_NAME": None,
     # Use a separate path for handling subscriptions.
     "SUBSCRIPTION_PATH": None,
+    # Set to True to enable GraphiQL headers editor tab
+    # This sets headerEditorEnabled GraphiQL option, for details go to
+    # https://github.com/graphql/graphiql/tree/main/packages/graphiql#options
+    "GRAPHIQL_HEADER_EDITOR_ENABLED": False,
```
eabruzzese, line 47: Just curious, why not always have it enabled? I think the majority use case for GraphiQL in Django is going to be development, no? Having the header panel always enabled seems like it would be fine. If someone _really_ wants to disable it, they can copy the template and configure GraphiQL themselves.

If we decide to keep it, you'll also need to update `views.py` to pass through the configuration option [like we do with `subscription_path`](https://github.com/graphql-python/graphene-django/blob/master/graphene_django/views.py#L168-L169) as well as the [template where it gets injected](https://github.com/graphql-python/graphene-django/blob/master/graphene_django/templates/graphene/graphiql.html#L43-L49).

radekwlsk: I did not want to enable it by default to not change current behaviour. But I have no objections to make it `True` by default if that is desired.

I will have a look at `views.py` and template later today, thanks for info.

eabruzzese: @radekwlsk I love the focus on stability and definitely appreciate the forethought!

Given that GraphiQL is provided by `graphene-django` as a convenient developer tool outside of the critical path, I don't think it has the same requirements for interface stability as the rest of the library. We upgraded GraphiQL to the latest version very recently in 2.12 which changed the interface quite a bit, and had I known about this feature at the time, I would have just turned it on in that PR.

@jkimbo @mvanlonden any feelings on this? Turning it on by default would make this a very small JS-only PR and one less setting to care about.

radekwlsk: @eabruzzese I've injected the setting to template. With enabling headers tab by default I will let a little bit more discussion to happen first.

jkimbo: I'm fine with enabling the setting by default since it's purely an enhancement and it shouldn't break anything.

radekwlsk:
> As mentioned I think defaulting the headers setting to `True` is better.

@jkimbo Should I just flip the default setting to `True` or remove setting entirely and as @eabruzzese suggested just make the change in JS file?

jkimbo:
@radekwlsk I think having the setting is good. I would just default it to true

graphene_django/templates/graphene/graphiql.html
```diff
@@ -45,6 +45,9 @@
     {% if subscription_path %}
       subscriptionPath: "{{subscription_path}}",
     {% endif %}
+    {% if header_editor_enabled %}
+      graphiqlHeaderEditorEnabled: "{{ graphiql_header_editor_enabled}}",
```
KubaJastrz, line 49: doesn't that mean it will be a `"true"` or `"false"` (string instead of boolean)?

graphene_django/static/graphene_django/graphiql.js
```diff
@@ -173,6 +175,7 @@
     onEditQuery: onEditQuery,
     onEditVariables: onEditVariables,
     onEditOperationName: onEditOperationName,
+    headerEditorEnabled: GRAPHENE_SETTINGS.graphiqlHeaderEditorEnabled || true,
```
jkimbo, line 178:
```suggestion
    headerEditorEnabled: GRAPHENE_SETTINGS.graphiqlHeaderEditorEnabled,
```

Otherwise setting `graphiqlHeaderEditorEnabled` to `false` will still set the option to `true`.

graphene_django/templates/graphene/graphiql.html
```diff
@@ -45,6 +45,9 @@
     {% if subscription_path %}
       subscriptionPath: "{{subscription_path}}",
     {% endif %}
+    {% if header_editor_enabled %}
+      graphiqlHeaderEditorEnabled: {{ graphiql_header_editor_enabled}},
+    {% endif %}
```
jkimbo, line 50:
```suggestion
      graphiqlHeaderEditorEnabled: {{ graphiql_header_editor_enabled|yesno:"true,false" }},
```

`header_editor_enabled` is not defined and the python boolean type needs to be converted to the js type.

Issue #303, opened 2017-10-27
Don't know how to convert the Django class django_mysql.models.fields.json.JSONField


quytang:
I've just added a [JSONField](http://django-mysql.readthedocs.io/en/latest/model_fields/json_field.html) on my models and graphene said `Don't know how to convert the Django class django_mysql.models.fields.json.JSONField`.
I think it wasn't supported yet but are there any way to work around?

thanks

quytang:
I end up with this snippet, thanks
```
import graphene
from graphene_django.converter import convert_django_field
from django_mysql.models import JSONField

@convert_django_field.register(JSONField)
def convert_json_field_to_string(field, registry=None):
    return graphene.String()
```

spockNinja:
That's definitely the way to go for a quick way to add your own fields that graphene-django doesn't support.

We would also welcome an addition like that to `converter.py` that does something similar to the postgres `JSONField`.

I will also note that we personally overrode the graphene type to be a `graphene.GenericScalar`, and then the JSON flows through and you do not have to re-parse it on the client side.

denis-ryzhkov:
I you found this issue looking for postgres-JSONField like me, maybe this patch will help you:
* Description: Ask `graphene_django` to convert `JSONField` to `GenericScalar` (JSON as is) instead of default `JSONString`
* Code: https://gist.github.com/denis-ryzhkov/59efc5be73eebd099a66cd549226fba9

cb109:
`return graphene.GenericScalar()` did not work for me, I had to import like `from graphene.types.generic import GenericScalar` and then `return GenericScalar()` instead. It is a great workaround, the JSON is passed through as expected, thanks!

dopeboy:
Adding some more color to this in case there are others who arrive from google:

Suppose you have a django model like so:

```
from django.contrib.postgres.fields import JSONField

class Rapper(models.Model):
    name = models.CharField()
    bars = JSONField()
```

Here's what the `Query` object and connecting type would look like:

```
from graphene.types import generic
from graphene_django.types import DjangoObjectType

from .models import Rapper

class RapperType(DjangoObjectType):
    bars = generic.GenericScalar()
    
    class Meta:
        model = Rapper
```

And here's what a sample GraphQL query and response end up looking like:

```
{
  rapper {
    name
    bars
  }
}
```

```
{
  "rapper": {
    "subject": {
      "name": "50 Cent",
      "bars": [
[...]

PR #1017, opened 2020-08-02
Add converter for django 3.1 JSONField

nikolaik:
Should probably wait until django 3.1 is released and possible to add to the test matrix.

Also found #303  thread suggesting people have a use case for converting JSONField (django 3.1, django_mysql and postgres ones) to `GenericScalar`. I'm not sure what is most common.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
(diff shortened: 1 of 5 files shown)


jkimbo:
Thanks @nikolaik . Django v3.1 has been released now so could you add it to the test matrix?

nikolaik:
> Thanks @nikolaik . Django v3.1 has been released now so could you add it to the test matrix?

yaz, done

Issue #809, opened 2019-11-01
offset pagination support for relay connection 


marxlow:
Hi all, I have a use case to paginate results based on an offset. I figured I would be able to add "offset" into my graphQL query to achieve this. (source: https://graphql.org/learn/pagination/)

Specifically, I want to implement fetching page 3 results of a list, where each page has a size of 20.  Doing this query throws "unknown argument "offset" on field allEntityA" 

```
query {
  allEntityA(first: 20, offset: 40) {
     edges {
       node {
       }
     }
  }
}
```

I have already trasnformed my Django model into a type `relay.Node`)

```
class entityANode(DjangoObjectType):
  class Meta:
    model = models.EntityA
    interfaces = (relay.Node,)
```

How can I achieve jumping from page 1 to 3 without using cursor navigation?

NateScarlet:
I've archive this by #563, but no one come to review.
You can try my https://github.com/NateScarlet/graphene-django-tools

tpict:
Our team would find #563 really helpful, is there anything we can do to help get it merged?

tpict:
Not stale! This is a very common requirement and @NateScarlet's PR provides it.

HenryKuria:
was stuck there too.. until i went to the [https://www.howtographql.com/graphql-python/8-pagination/](docs)
use skip to achieve the same functionality

HenryKuria:
> Hi all, I have a use case to paginate results based on an offset. I figured I would be able to add "offset" into my graphQL query to achieve this. (source: https://graphql.org/learn/pagination/)
> 
> Specifically, I want to implement fetching page 3 results of a list, where each page has a size of 20. Doing this query throws "unknown argument "offset" on field allEntityA"
> 
> ```
> query {
>   allEntityA(first: 20, offset: 40) {
>      edges {
>        node {
>        }
>      }
>   }
> }
> ```
> 
> I have already trasnformed my Django model into a type `relay.Node`)
> 
> ```
> class entityANode(DjangoObjectType):
>   class Meta:
>     model = models.EntityA
>     interfaces = (relay.Node,)
> ```
> 
> How can I achieve jumping from page 1 to 3 without using cursor navigation?

I have written a blog to address this issue. It can be achieved using "first" and "last" arguments only
read more here https://cloudworks.dukamneti.co.ke/blog/go-specific-page-graphene-django/

tcleonard:
[...]

PR #1013, opened 2020-07-28
Add offset pagination

tcleonard:
issue #809

PR #1022, opened 2020-08-12
Support use `first` and `last` at same time as `offset`

tpict:
This is @NateScarlet's work on https://github.com/graphql-python/graphene-django/pull/563. I can take over the PR since it's already had reviews and I think it's a very common use case. I'm not sure what the overlap with #1013 is.

Fixes https://github.com/graphql-python/graphene-django/issues/809


zbyte64:
Is this still relevant after #1013 getting merged in?

Issue #1020, opened 2020-08-07
GRAPHQL_SCHEMA is not included in the docs' testing example


karmek-k:
**Note: for support questions, please use stackoverflow**. This repository's issues are reserved for feature requests and bug reports.

* **What is the current behavior?**

I was following [the docs' article on testing](https://docs.graphene-python.org/projects/django/en/latest/testing/) and noticed that the example code raises the following error:

`AttributeError: Variable GRAPHQL_SCHEMA not defined in GraphQLTestCase.`

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via
a github repo, https://repl.it or similar (you can use this template as a starting point: https://repl.it/@jkimbo/Graphene-Django-Example).

[Here is the code from docs](https://gist.github.com/karmek-k/e361a5896ee03c1c87f5dd062af69644).
Put it somewhere in a Django project with graphene-django installed and run `python3 manage.py test`. 

* **What is the expected behavior?**

The tests should pass.

* **What is the motivation / use case for changing the behavior?**

It took me a while to investigate the problem. Hopefully this change may save others' time.

* **Please tell us about your environment:**
  
  - Version: Python 3.8.3
  - Platform: Manjaro Linux with KDE Plasma

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow)

The article should mention the GRAPHQL_SCHEMA property and how to set it properly.

nikolaik:
Docs look correct for version 0.23.0 and forward, but would fail with older versions. Pre v0.23.0 `GraphQLTestCase.GRAPHQL_SCHEMA` was actually only validated and not ever used. It was removed when refactoring test helpers in #1015 .

sjdemartini:
Closing since this is now obsolete in newer versions of graphene-django, and setting the schema is done via `GRAPHENE['SCHEMA'}` in `settings` as the docs now indicate ([here](https://github.com/graphql-python/graphene-django/blob/2358bd30a452859c50b6bdacdc27ae518716153b/docs/testing.rst#using-unittest) and [here](https://github.com/graphql-python/graphene-django/blob/2358bd30a452859c50b6bdacdc27ae518716153b/docs/settings.rst#schema))

PR #1023, opened 2020-08-13
Fix docs testing for GRAPHQL_SCHEMA (#1020)

elena:
Resolve error: "AttributeError: Variable GRAPHQL_SCHEMA not defined in
GraphQLTestCase." and issue https://github.com/graphql-python/graphene-django/issues/1020

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -16,7 +16,12 @@ Usage:
 
     from graphene_django.utils.testing import GraphQLTestCase
 
+    from my_fancy.schema import schema
+
+
     class MyFancyTestCase(GraphQLTestCase):
+        GRAPHQL_SCHEMA = schema
+
         def test_some_query(self):
             response = self.query(
                 '''


nikolaik:
Pre v2.13.0 `GraphQLTestCase.GRAPHQL_SCHEMA` was actually only validated and not ever used. It was removed when refactoring test helpers in #1015 . Should we mention it is only required with 2.12.1 and below?

jkimbo:
Thanks for the pull request @elena but @nikolaik is right: this is no longer needed for versions v2.13.0 onwards. Currently we don't handle documentation for different versions and I just assume people use the latest version but I'm happy to have a notice saying that for versions before v2.13 you need to define a schema on the GraphQLTestCase class. What do you think @elena ?

Also just to note that I do try and make minor changes backwards compatible so it should always be possible to be on the latest minor version. However I am aware that sometimes there are accidental backwards incompatible changes.

PR #1024, opened 2020-08-16
Fix testing doc

artofhuman:


artofhuman:
@jkimbo Hi, sorry for pr without description. As you can see we define pytest fixture and use the new fixture in the test case as an argument. But call the wrong fixture in test body. Can we merge this?

jkimbo (approved):

jkimbo:
Thanks @artofhuman

PR #1026, opened 2020-08-18
Update testing.rst

donegjookim:
- Import `json` as a dependency
- Add a missing colon on the function definition


r0n4ld (approved):

ulgens:
Why `json` is required in that code fragment?

DoctorJohn:
> Why `json` is required in that code fragment?

The json import is indeed used at the end of [the code snippet](https://github.com/graphql-python/graphene-django/blob/master/docs/testing.rst#using-pytest). (It doesn't show the full snippet in the code diff view).

jkimbo (approved):
Thanks @donegjookim

PR #1028, opened 2020-08-24
Allow passing of meta object in SerializerMutation options

jaw9c:
This follows the same behaviour as in `Mutation` and `ObjectType` in graphene, allowing the use of the _meta class to carry extra properties in superclasses


jkimbo (approved):
Thanks @jaw9c

Issue #906, opened 2020-03-17
Cannot use SerializerMutation with model's id


hairypalm:
**SerializerMutation** inherits from **ClientIDMutation**, which seems to remove **id** and insert **clientMutationId** into the mutation input specification.

Therefore, when I try to follow the tutorial at https://docs.graphene-python.org/projects/django/en/latest/mutations/ under **Create/Update Operations** it cannot work as described:  **id** is not exposed and so it cannot work as a **lookup_field**.

As such, I cannot find a way to use a **SerializerMutation** to update based on the instance **id**.  Am I using this correctly?  The Django **id** field is not shadowed by **clientMutationID**, just removed from the input spec.

From Documentation:
```
from graphene_django.rest_framework.mutation import SerializerMutation
from .serializers import MyModelSerializer


class AwesomeModelMutation(SerializerMutation):
    class Meta:
        serializer_class = MyModelSerializer
        model_operations = ['create', 'update']
        lookup_field = 'id'
```

My code:
```
class CompanyType(DjangoObjectType):
    class Meta:
        model = Company
        fields = '__all__'

class CompanySerializer(serializers.ModelSerializer):
    class Meta:
        model = Company
        # fields = '__all__'
        fields = ['pk', 'id', 'name']

class CompanyMutation(SerializerMutation):
    class Meta:
        serializer_class = CompanySerializer
        lookup_field = 'id'

class Mutation(graphene.ObjectType):
    company = CompanyMutation.Field()
```

A mutation like the below should work to update the Company instance's name with **id=1**, but the graphql service does not accept an **id**.

```
mutation {
  company(input: {id: 1, name: "Updated Name"}) {
    id
    name
  }
}
```

Replacing **id:1** with **clientMutationId:"1"** just creates a new Company with the provided name.

If I modify line 38 of **rest_framework/mutation.py** as follows then it works as I expect it to:
```
...
def fields_for_serializer(
    serializer,
    only_fields,
    exclude_fields,
    is_input=False,
    convert_choices_to_enum=True,
):
    fields = OrderedDict()
    for name, field in serializer.fields.items():
        is_not_in_only = only_fields and name not in only_fields
        is_excluded = any(
            [
                name in exclude_fields,
                field.write_only
                and not is_input,  # don't show write_only fields in Query
                field.read_only and is_input # don't show read_only fields in Input
                and name not in ['id', 'pk'],  # ADDED TO ALLOW FOR ID-BASED PARTIAL UPDATE
            ]
[...]

PR #1029, opened 2020-08-25
BUGFIX: don't filter out lookup_field as input (required for update)

CBuiVNG:
When using  `SerializerMutation` the pkey/lookup field is filtered out. see (https://github.com/graphql-python/graphene-django/issues/906). It was introduced by https://github.com/graphql-python/graphene-django/pull/882.

This seems incorrect. https://github.com/graphql-python/graphene-django/blob/ac1f9ac360138c7773774a61dd39da0e1d510425/graphene_django/rest_framework/mutation.py#L120 is using lookup field to determine if we are updating or creating. Because of https://github.com/graphql-python/graphene-django/pull/882 primary key fields are treated as read-only and can never be used for lookup. We need to allow lookup_field in input.

The workaround https://github.com/graphql-python/graphene-django/issues/906#issuecomment-602078174 does not work with relay. (Id field are typically numeric vs string for relay ids)

### DIFF
--- a/graphene_django/rest_framework/mutation.py
+++ b/graphene_django/rest_framework/mutation.py
@@ -36,7 +36,8 @@ def fields_for_serializer(
                 name in exclude_fields,
                 field.write_only
                 and not is_input,  # don't show write_only fields in Query
-                field.read_only and is_input
+                field.read_only
+                and is_input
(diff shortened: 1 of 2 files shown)


zbyte64 (approved):
:+1:

jkimbo (approved):

PR #1032, opened 2020-08-26
Add msg params to testing class

ulgens:
Both `assertIn` and `assertNotIn` has `msg` param to provide some extra / useful information in case of testing failures. That parameter was missing in `GraphQLTestCase`'s methods. I'm not sure how to test this behaviour enhancement.

Also, should i open this PR to v3 branch too or are we going to update v3 branch with a merge or sth?


jkimbo (approved):
Makes sense thanks @ulgens 

Also I'll update the v3 branch after this is released.

PR #1010, opened 2020-07-23
Fix backward Relay pagination (#1009)

lucas-bremond:
This PR is an attempt to fix the backward pagination issue #1009.

Issue #1009, opened 2020-07-23
Backward pagination when QuerySet larger than Relay Connection limit


lucas-bremond:
It appears that backward pagination may not work as expected when the size of the underlying QuerySet is larger than the [RELAY_CONNECTION_MAX_LIMIT](https://docs.graphene-python.org/projects/django/en/latest/settings/#relay-connection-max-limit).

Below is an example.

Assume a `Project` model, with a `number` property used as the default ordering, and that the database contains 500 projects (with `number` incrementing from `1` to `500`).

The following query:

```gql
query {
  projects (last: 2) {
    edges {
      node {
        number
      }
    }
  }
}
```

will return:

```json
{
  "data": {
    "projects": {
      "edges": [
        {
          "node": {
            "number": 99
          }
        },
        {
          "node": {
            "number": 100
          }
        }
      ]
    }
  }
}
```

instead of `Project 499` and `Project 500` (the actual last 2 projects in the database).

This is because `RELAY_CONNECTION_MAX_LIMIT=100` and therefore, the last 2 projects of the first 100 projects are returned instead.

*To confirm my understanding of pagination, I've tried running a similar query on the public [GitLab GraphQL API](https://gitlab.com/-/graphql-explorer), which behaves as expected.*

---

Environment:
  
  - Version: `v2.12.1`
  - Platform: `Linux`

PR #1046, opened 2020-10-16
Fix backward Relay pagination

tcleonard:
New version of the pull request #1010 (to solve issue #1009) which adds unit tests.


tcleonard:
Not sure what is causing those test failures... it fails in an import in django rest framework (and the import is totally valid).
Tests are passing locally.

zbyte64:
This is the relevant part I see from the test failure:

```
 .tox/py37-django111/lib/python3.7/site-packages/rest_framework/fields.py:15: in <module>
      from django.core.validators import (
  E   ImportError: cannot import name 'ProhibitNullCharactersValidator' from 'django.core.validators' (/home/runner/work/graphene-django/graphene-django/.tox/py37-django111/lib/python3.7/site-packages/django/core/validators.py)
```

DRF wants to import a validator from Django that doesn't exist in Django 1.11

tcleonard:
Indeed, which is not related to anything I have changed... Maybe DRF simply decided to drop support for older versions of Django and we might need to do the same here.

jkimbo:
@tcleonard yep DRF dropped support for Django 1.11. Fixed the tests here: https://github.com/graphql-python/graphene-django/pull/1047

tcleonard:
thanks @jkimbo, merged your changes and the CI now passes

zbyte64 (approved):
Thank you for this

vecchp:
@zbyte64 is there anything that needs to be done to get this in?

tcleonard:
I don't think so... for me it's ready to merge.. 🤷

graphene_django/fields.py
```diff
@@ -135,22 +135,23 @@ def resolve_connection(cls, connection, args, iterable, max_limit=None):
...
 
         # If after is higher than list_length, connection_from_list_slice
         # would try to do a negative slicing which makes django throw an
         # AssertionError
         after = min(get_offset_with_default(args.get("after"), -1) + 1, list_length)
 
         if max_limit is not None and "first" not in args:
-            args["first"] = max_limit
+            if "last" in args:
+                args["first"] = list_length
```
pcraciunoiu, line 151: Wondering why is this needed? In #1074 I didn't have to include it, seems like removing it wouldn't change anything.

graphene_django/tests/test_query.py
```diff
@@ -1213,6 +1213,103 @@ class Query(graphene.ObjectType):
...
+    def test_should_query(self, graphene_settings):
+        """
+        Backward pagination should work as expected
+        """
+        schema = self.setup_schema(graphene_settings, max_limit=100)
+        self.do_queries(schema)
+
+    def test_should_query_with_low_max_limit(self, graphene_settings):
+        """
+        When doing backward pagination (using last) in combination with a max limit higher than the number of objects
```
pcraciunoiu, line 1306: In this case, max limit is lower than the number of objects, right?

It looks like this test covers the same case as [this one](https://github.com/graphql-python/graphene-django/pull/1074/files#diff-cb65a9e9b2bd8192f5623c19ed4124223b411da82d4daf9770910074673adbb4R730), which is great!

pcraciunoiu (approved):
Makes sense to me. Just one change I'd suggest.

pcraciunoiu:
@jkimbo any chance we could get this merged?

Because graphene-django doesn't support true cursor pagination, the most reliable way to get recent items first is to go backwards, and this bug gets in the way of that.

jkimbo:
Thanks @zbyte64

graphene_django/fields.py
tcleonard: I would have to check if it is actually used by `connection_from_list_slice` but I set it at least for consistency with the other case where it is set to `max_limit`

tcleonard:
Thanks @zbyte64

PR #1050, opened 2020-10-23
fix `variables` key in body

robertobarreda:
There is a typo on checking if the key `variables` already exists in the body request.


jkimbo (approved):
Thanks!

PR #1054, opened 2020-11-01
Allow to use camel case in order by field

### LINKED ISSUE #1008: OrderingFilter should expose camelCase fields
Graphene Django's [documentation on ordering](https://docs.graphene-python.org/projects/django/en/latest/filtering/#ordering) exposes the `orderBy` filter:

```py
query {
  group(id: "xxx") {
    users(orderBy: "-created_at") {
      createdAt
    }
  }
}
```

Since the `created_at` Django field is actually exposed in GraphQL as `createdAt`,
shouldn't the ordering filter follow a consistent style by default?

```py
query {
  group(id: "xxx") {
    users(orderBy: "-createdAt") {
      createdAt
    }
  }
}
```

The workaround I've found is to do:

```py
order_by = OrderingFilter(
  fields = (
    ('created_at', 'createdAt')
  )
)
```

but the casing conversion probably be handled by the framework itself, to avoid the unnecessary noise / maintenance bloat.

artofhuman:
Fixes #1008

### DIFF
--- a/graphene_django/filter/fields.py
+++ b/graphene_django/filter/fields.py
@@ -3,6 +3,7 @@ from functools import partial
 
 from django.core.exceptions import ValidationError
 from graphene.types.argument import to_arguments
+from graphene.utils.str_converters import to_snake_case
 from ..fields import DjangoConnectionField
 from .utils import get_filtering_args_from_filterset, get_filterset_class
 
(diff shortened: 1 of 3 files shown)


zbyte64 (approved):
I like that this is backwards compatible :+1:

PR #1056, opened 2020-11-04
Fix issue #1055

### LINKED ISSUE #1055: Cant export schema to stdout in graphql format
The format for the schema(json or graphql SDL) is defined by the extension of the `--out` flag. If `--out=-`(stdout), there's no way to generate a GraphQL SDL representation of the schema.

leonardoarroyo:
Add option to use `--out=-.graphql` to export the schema on graphql SDL format to stdout.
Closes #1055.

### DIFF
--- a/graphene_django/management/commands/graphql_schema.py
+++ b/graphene_django/management/commands/graphql_schema.py
@@ -60,8 +60,10 @@ class Command(CommandArguments):
 
     def get_schema(self, schema, out, indent):
         schema_dict = {"data": schema.introspect()}
-        if out == "-":
+        if out == "-" or out == "-.json":
             self.stdout.write(json.dumps(schema_dict, indent=indent, sort_keys=True))
+        if out == "-.graphql":
+            print(print_schema(schema))
         else:
             # Determine format
             _, file_extension = os.path.splitext(out)
--- a/graphene_django/management/commands/graphql_schema.py
+++ b/graphene_django/management/commands/graphql_schema.py
@@ -63,7 +63,7 @@ class Command(CommandArguments):
         if out == "-" or out == "-.json":
             self.stdout.write(json.dumps(schema_dict, indent=indent, sort_keys=True))
         elif out == "-.graphql":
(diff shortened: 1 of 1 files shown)


graphene_django/management/commands/graphql_schema.py
```diff
@@ -60,8 +60,10 @@ def save_graphql_file(self, out, schema):
 
     def get_schema(self, schema, out, indent):
         schema_dict = {"data": schema.introspect()}
-        if out == "-":
+        if out == "-" or out == "-.json":
             self.stdout.write(json.dumps(schema_dict, indent=indent, sort_keys=True))
+        elif out == "-.graphql":
+            print(print_schema(schema))
```
zbyte64, line 66: This should use stdout instead of print.

graphene_django/management/commands/graphql_schema.py
```diff
@@ -60,8 +60,10 @@ def save_graphql_file(self, out, schema):
 
     def get_schema(self, schema, out, indent):
         schema_dict = {"data": schema.introspect()}
-        if out == "-":
+        if out == "-" or out == "-.json":
```
zbyte64, line 63: Not sure if this is intuitive. Personally prefer an explicit option but I see that you are continuing an existing pattern, so.. :+1:

zbyte64 (changes requested):
For sure fix the print statement.

graphene_django/management/commands/graphql_schema.py
leonardoarroyo: I would prefer a command line flag too, but I decided to go with this since this was how the code was laid out first. I don`t wanna introduce any breaking changes since I just need this to work for my project :-)

graphene_django/management/commands/graphql_schema.py
leonardoarroyo: Changed, my friend.

zbyte64 (approved):

Issue #1038, opened 2020-09-16
AttributeError: 'function' object has no attribute '__func__' error while using SerializerMutation 


sfahad1414:
graphene-django version 2.13.0

gettting following error

  File "D:\miniconda\envs\deeplearning\lib\site-packages\graphene_django\rest_framework\mutation.py", line 110, in __init_subclass_with_meta__
    super(SerializerMutation, cls).__init_subclass_with_meta__(
  File "D:\miniconda\envs\deeplearning\lib\site-packages\graphene\relay\mutation.py", line 43, in __init_subclass_with_meta__
    if cls.mutate and cls.mutate.__func__ == ClientIDMutation.mutate.__func__:
AttributeError: 'function' object has no attribute '__func__'

zbyte64:
Would it be possible to show us your Mutation? Also would be good to know which version of DRF you're using.

zbyte64:
Was able to reproduce this, it happens when you define a mutation and the mutate function is not a classmethod.

zbyte64:
Open issue covering usage: https://github.com/graphql-python/graphene/issues/810

Other issue is our docs are out of date, the following does not use `classmethod`: https://docs.graphene-python.org/projects/django/en/latest/mutations/

zbyte64:
Closing issue as the documentation is now updated.

PR #1059, opened 2020-11-08
Use explicit classmethod in simple mutation example

artofhuman:
rel #1038
rel https://github.com/graphql-python/graphene/issues/810


zbyte64 (approved):
:+1: I think this will eliminate future headaches for some people

PR #1060, opened 2020-11-08
Replace Unidecode package with text_unidecode package

### LINKED ISSUE #1014: Replace Unidecode package with text_unidecode package

  The Unidecode package is licensed under GPL, which makes it impossible to include it in any non-GPL/MIT licensed software. Since graphene-django is licensed under MIT, it would be great to allow graphene-django to be included in non-GPL/MIT licensed software, which is currently disallowed by the GPL of unidecode (which is the only dependency with such a problem). 

  The alternative package text_unidecode offers a drop-in replacement for unidecode with a less restrictive license. However, as stated to the project website unidecode offers better memory usage and better transliteration quality over text_unidecode. Will this be a show stopper or would it be possible to exchange the packages?

  graphene-django should be usable within any project.

  Allow graphene-django to be used in non-GPL/MIT licensed projects.

  - Version: 2.12.1
  - Platform: Linux

artofhuman:
Closes #1014

### DIFF
--- a/graphene_django/utils/str_converters.py
+++ b/graphene_django/utils/str_converters.py
@@ -1,5 +1,5 @@
 import re
-from unidecode import unidecode
+from text_unidecode import unidecode
 
 
 def to_const(string):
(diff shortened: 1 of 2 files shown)


zbyte64 (approved):
:+1: 

What helped convince me was watching wagtail respond to this issue: https://github.com/wagtail/wagtail/issues/3311

PR #1064, opened 2020-11-18
Add a default msg to show the response content.

jackton1:
This seems like an issue with using assertResponseNoErrors and assertResponseHasErrors 

Which doesn't include any errors specific to the response and currently just shows.

```python
    self.assertNotIn("errors", list(content.keys()))
AssertionError: 'errors' unexpectedly found in ['errors', 'data']
```

Expected:

```python
self.assertNotIn("errors", list(content.keys()), content)
AssertionError: 'errors' unexpectedly found in ['errors', 'data'] : {'errors': [{'message': "mutate() missing 1 required positional argument: 'info'", 'locations': [{'line': 3, 'column': 17}], 'path': ['eventCancel']}], 'data': None}
```

@artofhuman @syrusakbary

### DIFF
--- a/graphene_django/utils/testing.py
+++ b/graphene_django/utils/testing.py
@@ -109,8 +109,8 @@ class GraphQLTestCase(TestCase):
         the call was fine.
         :resp HttpResponse: Response
         """
-        self.assertEqual(resp.status_code, 200)
         content = json.loads(resp.content)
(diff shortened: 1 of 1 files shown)


tcleonard:
Nice

jottenlips (approved):

PR #1074, opened 2020-12-15
Fix pagination for 'last' page.

pcraciunoiu:
When paginating from the end of the results, the last items were not returned correctly.

Added a test to showcase the issue.

### DIFF
--- a/graphene_django/fields.py
+++ b/graphene_django/fields.py
@@ -147,21 +147,19 @@ class DjangoConnectionField(ConnectionField):
 
         if isinstance(iterable, QuerySet):
             list_length = iterable.count()
-            list_slice_length = (
-                min(max_limit, list_length) if max_limit is not None else list_length
-            )
         else:
             list_length = len(iterable)
-            list_slice_length = (
-                min(max_limit, list_length) if max_limit is not None else list_length
-            )
+
+        list_slice_length = list_length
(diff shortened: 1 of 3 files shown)


graphene_django/fields.py
```diff
@@ -147,21 +147,19 @@ def resolve_connection(cls, connection, args, iterable, max_limit=None):
...
-                min(max_limit, list_length) if max_limit is not None else list_length
-            )
         else:
             list_length = len(iterable)
-            list_slice_length = (
-                min(max_limit, list_length) if max_limit is not None else list_length
-            )
+
+        list_slice_length = list_length
+        if max_limit is not None and "last" not in args:
```
pcraciunoiu, line 154: Without checking for `last` here and below, `list_slice_length` was being set to 4, and `after` to `0`, slicing off reporters 5-8.

Query for last 2 reporters should be getting 7 and 8, but it wasn't getting those correctly.

tcleonard:
I think this is the doing basically the same thing as the PR I submitted a while ago: https://github.com/graphql-python/graphene-django/pull/1046 which is still waiting to be merged

pcraciunoiu:
@tcleonard thanks for the ping! Looks great, I left a small comment there. Closing in favor of #1046.

PR #1076, opened 2020-12-24
Fix 1075: serialize datetime value

rasheedhamdawi:
Serialize datetime

### DIFF
--- a/graphene_django/views.py
+++ b/graphene_django/views.py
@@ -1,6 +1,7 @@
 import inspect
 import json
 import re
+import datetime
 
 import six
 from django.db import connection, transaction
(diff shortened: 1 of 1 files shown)


graphene_django/views.py
```diff
@@ -55,6 +56,11 @@ def instantiate_middleware(middlewares):
         yield middleware
 
 
+def stringify_datetime(value):
+    if isinstance(value, datetime.datetime):
+        return value.__str__()
```
rganeyev, line 61: what happens if the value is not a datetime instance? shouldn't we just try to stringify everything?

```suggestion
def stringify(value):
    return str(value)
```

rganeyev (commented):
LGTM, please look to the comments

zbyte64:
I don't understand the "why". My understanding is that mapping to Graphene's DateTime is sufficient and is already done for Django's DateTimeField. A test demonstrating the need or configuration where this becomes relevant is needed.

PR #1080, opened 2020-12-27
Introduced optional_fields to SerializationMutation

rganeyev:
Added optional_fields to SerializationMutation to forcefully mark some fields as optional

### DIFF
--- a/graphene_django/rest_framework/tests/test_mutation.py
+++ b/graphene_django/rest_framework/tests/test_mutation.py
@@ -3,7 +3,7 @@ import datetime
 from py.test import raises
 from rest_framework import serializers
 
-from graphene import Field, ResolveInfo
+from graphene import Field, ResolveInfo, NonNull, String
 from graphene.types.inputobjecttype import InputObjectType
 
 from ...types import DjangoObjectType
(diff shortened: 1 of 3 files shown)


sebslomski:
Please merge, it's an excellent addition! Thanks!

zbyte64:
I don't have any objection to the feature. But please add some tests.

rganeyev:
@zbyte64, added two tests: when `optional_fields` is not set and `optional_fields` meta is set. Please review

graphene_django/rest_framework/tests/test_mutation.py
```diff
@@ -98,6 +98,25 @@ class Meta:
...
+
+    assert "cool_name" in MyMutation.Input._meta.fields
+    assert MyMutation.Input._meta.fields["cool_name"].type == NonNull(String)
+
+
+def test_model_serializer_optional_fields():
+    class MyMutation(SerializerMutation):
+        class Meta:
+            serializer_class = MyModelSerializer
+            optional_fields = "cool_name"
```
zbyte64, line 114: Isn't this suppose to be a tuple or list? ie `("cool_name",)` ; It looks like this currently works because it's doing a substring search, "cool_name" is in "cool_name".

zbyte64 (changes requested):
Just the one comment, looks good otherwise

graphene_django/rest_framework/tests/test_mutation.py
rganeyev: Good catch! At first, I wrongly set `optional_fields = ('cool_name')`, then black formatted changed it to resulting `optional_fields = "cool_name"`.

Thanks for noticing, fixed.

zbyte64 (approved):
:+1:

mahmoudmostafa0:
why this feature is remove now ?

PR #1081, opened 2020-12-28
Also convert BaseCSVFilter for custom fields

bellini666:
The PR https://github.com/graphql-python/graphene-django/pull/1070 changed the automatically generated filters to use a lista of the given type instead of forcing a string separated by commas.

This should also apply to custom filters when using `filters.BaseInFilter` for example.

### DIFF
--- a/graphene_django/filter/tests/test_in_filter.py
+++ b/graphene_django/filter/tests/test_in_filter.py
@@ -1,9 +1,11 @@
 import pytest
 
+from django_filters import FilterSet
+from django_filters import rest_framework as filters
 from graphene import ObjectType, Schema
 from graphene.relay import Node
 from graphene_django import DjangoObjectType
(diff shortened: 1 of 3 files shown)


graphene_django/filter/tests/test_in_filter.py
```diff
@@ -61,6 +82,33 @@ def test_string_in_filter():
...
+    """Test in filter on a string field with a custom filterset class."""
+    Person.objects.create(name="John")
+    Person.objects.create(name="Michael")
+    Person.objects.create(name="Angela")
+
+    schema = Schema(query=Query)
+
+    query = """
+    query {
+        people (names: ["John", "Michael"]) {
```
zbyte64, line 95: Neat.

zbyte64 (approved):
:+1:

PR #1082, opened 2020-12-29
return back instance fields so as not to overwrite not required field…

ncrypt4nix:
When working with django forms, any fields specified as not required will replace the instance field as None on save.
I propose to return these fields to kwargs to fix this.
Sorry, about my English.


zbyte64:
A test demonstrating the defect/improvement would be good.

firaskafri (changes requested):
Tests

PR #1083, opened 2020-12-29
improvement: convert decimal field to graphene decimal

### LINKED ISSUE #91: Convert DecimalField to String rather than Float

zbyte64:
Fixes #91 by converting `models.DecimalField` to graphene's `Decimal` type

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -18,6 +18,7 @@ from graphene import (
     DateTime,
     Date,
     Time,
+    Decimal,
 )
 from graphene.types.json import JSONString
 from graphene.utils.str_converters import to_camel_case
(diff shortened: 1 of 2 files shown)


rganeyev (approved):
LGTM, thanks for the tests!

ulgens (approved):

PR #1084, opened 2020-12-30
Use the Django TestCase's Client

ulgens:
Use the Django Client test utility instance that Django provides with its TestCase class. This allows GraphQL tests to make use of the stateful client methods like login()

Updated version of https://github.com/graphql-python/graphene-django/pull/886, our oldest open PR.

### DIFF
--- a/graphene_django/tests/test_utils.py
+++ b/graphene_django/tests/test_utils.py
@@ -51,7 +51,9 @@ def test_graphql_test_case_op_name(post_mock):
             pass
 
     tc = TestClass()
+    tc._pre_setup()
     tc.setUpClass()
+
     tc.query("query { }", op_name="QueryName")
     body = json.loads(post_mock.call_args.args[1])
     # `operationName` field from https://graphql.org/learn/serving-over-http/#post-request
--- a/graphene_django/utils/testing.py
+++ b/graphene_django/utils/testing.py
@@ -1,6 +1,7 @@
 import json
+import warnings
 
-from django.test import TestCase, Client
+from django.test import Client, TestCase
 
 DEFAULT_GRAPHQL_URL = "/graphql/"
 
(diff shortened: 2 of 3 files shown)


graphene_django/tests/test_utils.py
```diff
@@ -51,7 +51,9 @@ def runTest(self):
             pass
 
     tc = TestClass()
+    tc._pre_setup()
     tc.setUpClass()
```
ulgens, line 55: I'm not sure why `.setUpClass` had to be called manually here but for same reason, `._pre_setup()` should be called to. Otherwise, Django's `TestCase` doesn't populate `TestCase.client`.

graphene_django/utils/testing.py
```diff
@@ -99,10 +94,19 @@ def query(self, query, op_name=None, input_data=None, variables=None, headers=No
             input_data=input_data,
             variables=variables,
             headers=headers,
-            client=self._client,
+            client=self.client,
             graphql_url=self.GRAPHQL_URL,
         )
 
+    @property
+    def _client(self):
```
ulgens, line 102: Assuming people calling (get attribute) the client with `_client` in their tests, including myself, i didn't want to break all at once. But i'm not sure how to handle set attribute case, or does it even necessary.

graphene_django/utils/tests/test_testing.py
```diff
@@ -0,0 +1,24 @@
+import pytest
```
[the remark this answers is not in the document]
zbyte64: Best I came up with is test_harness, which I didn't think is any better.

graphene_django/utils/testing.py
zbyte64: Sensible.

zbyte64 (approved):
:+1:

PR #1087, opened 2020-12-31
Fix project setup

lucas-bremond:
Fixing some issues when starting from scratch, and simply following [CONTRIBUTING.md](https://github.com/graphql-python/graphene-django/blob/main/CONTRIBUTING.md)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,6 @@ tests_require = [
     "coveralls",
     "mock",
     "pytz",
-    "psycopg2-binary",
     "django-filter>=2",
     "pytest-django>=3.3.2",
 ] + rest_framework_require
--- a/setup.py
+++ b/setup.py
@@ -20,6 +20,7 @@ tests_require = [
     "coveralls",
     "mock",
     "pytz",
+    "psycopg2-binary",
     "django-filter>=2",
     "pytest-django>=3.3.2",
 ] + rest_framework_require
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -20,6 +20,7 @@
     "coveralls",
     "mock",
     "pytz",
+    "psycopg2-binary",
```
ulgens, line 23: Why this is a graphene dependency?

zbyte64: It's under "test_requires" but I am shy about adding this as well.

lucas-bremond: I've just been following the CONTRIBUTING guide starting from a fresh install, within a virtual environment:

```bash
git clone git@github.com:graphql-python/graphene-django.git
cd graphene-django
virtualenv venv
source venv/bin/activate
make dev-setup
# Fix `DJANGO_SETTINGS_MODULE` in setup.cfg
make tests
```

yields:

```txt
FAILED graphene_django/tests/test_query.py::test_should_query_postgres_fields - ModuleNotFoundError: No module named 'psycopg2'
== 1 failed, 308 passed, 5 skipped, 15 warnings in 5.27s ==
```

I guess either `psycopg2` is really required, or this test should be skipped by default?

zbyte64: Interesting, looks like a regression with Django 3.1 as it provides JSONField but not RangeField, so it no longer skips the test if postgres isn't installed. The test should check if RangeField is missing instead. You should definately drop the dependency. If you fix the test decorator that would be a bonus, or we can take care of it in a separate PR if you wish. Thank you for trying the guide and giving feedback.

lucas-bremond: Alright! Please check :)

zbyte64 (approved):
Thank you for improving the test as well :fire:

PR #1090, opened 2021-01-08
Validate in and range filter inputs

tcleonard:
- Fix `_In` filtering when providing an empty list.
- Validate input of `_Range` filtering to force to provide 2 values


zbyte64 (approved):
Stellar :rocket:

Issue #1088, opened 2021-01-06
Django test assertions for graphene responses no longer working on 2.15.0


baldwinchang:
**Note: for support questions, please use stackoverflow**. This repository's issues are reserved for feature requests and bug reports.

* **What is the current behavior?**

assertions that used to pass no longer pass, with no changes to the code base

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via
a github repo, https://repl.it or similar (you can use this template as a starting point: https://repl.it/@jkimbo/Graphene-Django-Example).

test case code:
https://repl.it/@baldwinchang/LightgrayRaggedBlock

* **What is the expected behavior?**

Assertions do not fail

* **What is the motivation / use case for changing the behavior?**

assert statements no longer equate

* **Please tell us about your environment:**
  
  - Version: 2.15.0
  - Platform: Debian Docker

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow)

Downgrading the version to 2.14.0 fixes the issue. 

Looks like #1084 is the culprit

ulgens:
What is the `TestCase` class in example code?

PR #1093, opened 2021-01-09
Fix backward compability on GraphQLTestCase._client setter (v2)

ulgens:
I didn't handle backward compatability for setter in https://github.com/graphql-python/graphene-django/pull/1084/files#r549921387 and it seems it caused problems for several users.

Relates to #1088, hopefully fixes it :slightly_smiling_face:


zbyte64 (approved):
So this will allow people to set `_client` and fix the problem for most it seems. Cool :+1:

PR #1094, opened 2021-01-09
Fix backward compability on GraphQLTestCase._client setter

ulgens:
main / v3 version of https://github.com/graphql-python/graphene-django/pull/1093


zbyte64 (approved):
:+1:

PR #1108, opened 2021-01-29
Only override non declared filters 

DonQueso89:
As the docstring of `replace_csv_filters` already states, declared filters should not be converted to InFilter/RangeFilter since it is possible that these are intentionally customised. This function is now actually also converting declared filters


zbyte64:
I believe this got fixed up with the recent merge of #1119 which restored the `in` lookup behavior. Let us know if that is not the case.

Issue #1072, opened 2020-12-08
Catching GraphQLLocatedError for tests


laurent-brisbois:
**Is your feature request related to a problem? Please describe.**
No, question about doing a particular thing

**Describe the solution you'd like**
I'd like to know the way to catch GraphQLLocatedError, or a way to `assertRaises` them.

I've tried :

- try...except way :
```python
try:
    response = self.client.execute(query, variables)
except GraphQLLocatedError:
    self.fail(...)
```

- with self.assertRaises(...) way :
```python
with self.assertRaises(GraphQLLocatedError):
    response = self.client.execute(query, variables)
```

But none of these work. How can I do this ?

ulgens:
I think what does `do_smth` do here is important. Graphql executor doesn't just throw that error and break the operation. It tries to return the error in response body. You may want to look at `graphene_django.utils.testing.GraphQLTestCase`

laurent-brisbois:
@ulgens Thanks for your answer.
> I think what does `do_smth` do here is important. Graphql executor doesn't just throw that error and break the operation. It tries to return the error in response body. You may want to look at `graphene_django.utils.testing.GraphQLTestCase`

I edited my post. the `do_smth` is in my case the execution of the query. In the test I'm trying to write, I execute the test once while being logged, which should be okay (I use graphql-jwt for authentication and I've set the `@login_required` decorator on my query).

And I also execute the query whereas I'm not logged in in a second test. The goal is to check that an exception is thrown If I perform a query while not logged in. If, for instance, an `assertRaises` fails, that would mean for me that someone has removed the `@login_required` decorator.

I hope I'm clear :D

zbyte64:
Could provide a custom ExecutionContext class with it's own `handle_field_error` that could stash the original errors for you and then assert they are proper in the test. We need more documentation covering this pattern.

laurent-brisbois:
> Could provide a custom ExecutionContext class with it's own `handle_field_error` that could stash the original errors for you and then assert they are proper in the test. We need more documentation covering this pattern.

@zbyte64 
Could your provide a basic structure for that ? I'm not sure to understand what you mean.

zbyte64:
In the past I have used the following class to get stack traces of exceptions:

```
from graphql.execution import ExecutionContext

class DebugExecutionContext(ExecutionContext):
        def handle_field_error(
            self, raw_error: Exception, field_nodes, path, return_type,
        ) -> None:
            import traceback            
[...]

PR #1109, opened 2021-02-03
Add ability to pass `execution_context_class` to `GraphQLView.as_view()`

kubami:
Currently when passing `execution_context_class` like this:

```
GraphQLView.as_view(execution_context_class=CustomContext)
```

you get the following error from `View.as_view()`
```
TypeError: GraphQLView() received an invalid keyword 'execution_context_class'. as_view only accepts arguments that are already attributes of the class.
```

this PR fixes the `hasattr` check in `.as_view`.

Fixes: #1072

### DIFF
--- a/graphene_django/views.py
+++ b/graphene_django/views.py
@@ -85,7 +85,7 @@ class GraphQLView(View):
     batch = False
     subscription_path = None
     execution_context_class = None
-    
+
     def __init__(
         self,
         schema=None,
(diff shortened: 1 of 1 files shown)


Didericis (approved):

zbyte64:
Great catch. Would like to add a test for this, let me know if you're up for it, otherwise we'll merge as is.

kubami:
Hi @zbyte64 thanks for looking at this. As you can see I am a bit strapped for time at this moment. Let's merge it as is.

kubami:
Is it possible to merge this in as is?

weilu:
This got me too. Can we have this released please? On a somewhat relevant note, per fail fast principle can we have `UnforgivingExecutionContext` be the default context so graphene doesn't swallow exceptions?

PR #1115, opened 2021-02-15
graphene-neo4j  

MajidHeydari:
I will be glad if you update graphene-neo4j  with Django 3.


ulgens:
I'm not sure this is the correct repo for this PR. The target should be https://github.com/MardanovTimur/graphene-neo4j

PR #1117, opened 2021-02-15
Doc clarification for headers arg in testing utils

isundaylee:
I think it might be helpful to add an explicit hint that HTTP headers should be prepended with `HTTP_` as required by `django.test.Client` (at least it was not super obvious to me when I tried to use it).

### DIFF
--- a/graphene_django/utils/testing.py
+++ b/graphene_django/utils/testing.py
@@ -28,7 +28,9 @@ def graphql_query(
         variables (dict)            - If provided, the "variables" field in GraphQL will be
                                       set to this value.
         headers (dict)              - If provided, the headers in POST request to GRAPHQL_URL
-                                      will be set to this value.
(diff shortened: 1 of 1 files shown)


zbyte64:
Great idea adding this call out to the docs. The failure is unrelated to your commit, I noticed someone had a PR that tacked on a fix: https://github.com/graphql-python/graphene-django/pull/1112/files#diff-1d9d6e9e707386e73d8f3b6766c664ff038fd4268bc2b7321e42468f7f31b437
That PR might take a little longer as it adds another database query that it shouldn't. We'll discuss how to resolve this blocker in tomorrow's meeting.

zbyte64:
Main branch should be fixed now. Please git fetch main, merge, and push to see if it resolves the build.

zbyte64 (approved):
Documentation improvements are much appreciated :+1:

PR #1119, opened 2021-02-16
Add enum support to filters and fix filter typing (v3)

tcleonard:
- Add filtering support for choice fields converted to graphql Enum (or not)
- Fix type of various filters (used to default to String)
- Fix bug with contains introduced in previous PR
- Fix bug with declared filters being overridden (see PR #1108)
- Fix support for ArrayField and add documentation
103427653-58ec3a80-4b77-11eb-9543-775c9c208dbd

### DIFF
--- a/graphene_django/filter/fields.py
+++ b/graphene_django/filter/fields.py
@@ -43,8 +43,8 @@ class DjangoFilterConnectionField(DjangoConnectionField):
             if self._extra_filter_meta:
                 meta.update(self._extra_filter_meta)
 
-            filterset_class = self._provided_filterset_class or (
-                self.node_type._meta.filterset_class
+            filterset_class = (
+                self._provided_filterset_class or self.node_type._meta.filterset_class
             )
             self._filterset_class = get_filterset_class(filterset_class, **meta)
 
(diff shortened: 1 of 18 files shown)


graphene_django/filter/fields.py
```diff
@@ -68,7 +87,7 @@ def filter_kwargs():
                 if k in filtering_args:
                     if k == "order_by" and v is not None:
                         v = to_snake_case(v)
-                    kwargs[k] = v
+                    kwargs[k] = convert_enum(v)
```
zbyte64, line 90: Does this mean `order_by` arguments can be an enum?

zbyte64 (approved):
Amazing work, just have the one question with regards to `order_by`

graphene_django/filter/fields.py
tcleonard: no `order_by` have to be strings since it is supposed to be the name of the field you want to order by (and `to_snake_case` only works for strings)
`convert_enum` method  only transform `Enum`s and left other things untouched (so a string wouldn't get modified by the function)

PR #1122, opened 2021-02-16
enhancement: DjangoDebugContext captures exceptions

### LINKED ISSUE #1121: Print exception traceback of GraphQLLocatedError
Hi!

I just spent the whole afternoon trying to find out how to display the exception traceback when the exception happens as a result of a request. I have a KeyError somewhere in my code and all I can see is

```graphql
{
  "errors": [
    {
      "message": "'username'",
      "locations": [
        {
          "line": 17,
          "column": 3
        }
      ],
      "path": [
        "CustomUserUpdate"
      ]
    }
  ],
  "data": null
}
```

Would be way more informative if either the message field could somehow show the whole traceback, or, preferably, the traceback was printed into the console.

How do I achieve that, if possible at all?

Thanks!

zbyte64:
DjangoDebugContext captures exceptions and allows captured stack traces to be queried. Example usage added to documentation. 

Fixes #1121

Discovered that `next` does not return a promise in our test environment. Do I need to check if the return is a promise?

### DIFF
--- a/docs/debug.rst
+++ b/docs/debug.rst
@@ -4,7 +4,7 @@ Django Debug Middleware
 You can debug your GraphQL queries in a similar way to
 `django-debug-toolbar <https://django-debug-toolbar.readthedocs.org/>`__,
 but outputting in the results in GraphQL response as fields, instead of
-the graphical HTML interface.
+the graphical HTML interface. Exceptions with their stack traces are also exposed.
 
 For that, you will need to add the plugin in your graphene schema.
 
@@ -62,6 +62,10 @@ the GraphQL request, like:
       _debug {
         sql {
           rawSql
+        }
+        exceptions {
+          message
+          stack
         }
       }
     }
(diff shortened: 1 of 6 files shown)


tcleonard (approved):
Awesome! Was looking for that for a while!

tcleonard:
> Discovered that next does not return a promise in our test environment. Do I need to check if the return is a promise?

Just saw that comment... not sure why next does not return a promise in the first place, but what is the implication of this?

zbyte64:
Could mean an exception might get raised and not logged. Or maybe the variable name is incorrect. My current understanding is it gets call by this: https://github.com/graphql-python/graphql-core/blob/6494c7fcfa62418a4fe36cf23e0bbfebc10fe342/src/graphql/execution/middleware.py#L36 which doesn't seem to do anything with promises...

khink:
The try-except around `promise = next(root, info, **args)` in e9f25ecf is the cause of issue #1367. It makes raising a GraphQLError return a `"Promise rejected"` instead of an "errors" list. From what i can gather, raising GraphQLError is a pattern that is supported throughout Graphene. 

If the try-except could be removed, that would make graphene-django work more as people might expect, i think.

PR #1124, opened 2021-02-19
Fix subscriptions in JS

ygbourhis:
The "play" button on the graphiql web page does not work as expected with subscription.
With the following fix, when launching a subscription the "play" button changes to a "stop" button, allowing to stop subscriptions properly.
See the following video capture which shows how the fix works:
https://youtu.be/2Aks6wnYeRc

Fixes https://github.com/graphql-python/graphene-django/issues/1125

### DIFF
--- a/graphene_django/static/graphene_django/graphiql.js
+++ b/graphene_django/static/graphene_django/graphiql.js
@@ -123,8 +123,8 @@
     if (operationType === "subscription") {
       return {
         subscribe: function (observer) {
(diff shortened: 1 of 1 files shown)


ygbourhis:
These tests fail locally even without my change...

zbyte64:
Main branch should be fixed now. Please git fetch main, merge, and push to see if it resolves the build.

zbyte64 (approved):
:+1:

PR #1131, opened 2021-02-22
Fix: Allow add external resolver

sebsasto:
Currently you cannot use an external resolver in `DjangoListField`. If you pass a resolver as a parameter in the Field definition it is ignored completely. This PR allows to pass the the resolver as an argument in the Field definition.

Example:

```python
def external_resolver(parent, info, **args):
    return ...

    all_xxx = DjangoListField(XXXType, resolver=external_resolver)
```

### DIFF
--- a/graphene_django/fields.py
+++ b/graphene_django/fields.py
@@ -61,12 +61,13 @@ class DjangoListField(Field):
         return queryset
 
     def wrap_resolve(self, parent_resolver):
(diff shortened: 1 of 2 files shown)


sebsasto:
@zbyte64 I added the test cases. Please let me know if it is ok

zbyte64 (approved):
Thank you for making the resolver argument work with DjangoListField :+1:

PR #1135, opened 2021-02-24
Fix main branch tests failing due to wrong instancing of Missing class

sebsasto:
Main tests are broken in the main branch because the MissingType does not take any arguments.

### DIFF
--- a/graphene_django/compat.py
+++ b/graphene_django/compat.py
@@ -1,5 +1,6 @@
 class MissingType(object):
-    pass
+    def __init__(self, *args, **kwargs):
+        pass
 
 
 try:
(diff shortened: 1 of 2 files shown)


graphene_django/compat.py
```diff
@@ -1,5 +1,6 @@
 class MissingType(object):
-    pass
+    def __init__(self, *args, **kwargs):
```
zbyte64, line 2: Not sure why this is needed?

zbyte64 (approved):
Thank you for catching that. Just the one question :+1:

graphene_django/compat.py
sebsasto: When you are initializing things like the ArrayField it will send some args to the constructor that by default the MissingType is not expecting and therefore will fail saying that the Field was overloaded

tcleonard (approved):

PR #1136, opened 2021-02-24
fix: declaration of required variable in filters v2

tcleonard:
This is to fix the problem you reported @ulgens

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -69,7 +69,11 @@ def convert_choices_to_named_enum_with_descriptions(name, choices):
         def description(self):
             return named_choices_descriptions[self.name]
 
-    return Enum(name, list(named_choices), type=EnumWithDescriptionsType)
+    if named_choices == []:
+        # Python 2.7 doesn't handle enums with lists with zero entries, but works okay with empty sets
+        named_choices = set()
+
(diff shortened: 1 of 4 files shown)


ulgens:
Is there any chance we can add test for this too? I didn't understand how tests are didn't fail before.

tcleonard:
yep I can add a test for it, good point. (it's when you don't use any automatic filters and only custom ones)

tcleonard:
There is a test failing which I have no idea what it is about... it doesn't seem linked to what I have changed I can't reproduce this failure locally... any idea?

zbyte64:
I'll play with it this weekend, looks like names is an empty string when it attempts to generate an enum for the file path field.

zbyte64:
Can hack around it by modifying: graphene_django/converter.py: convert_choices_to_named_enum_with_descriptions
```
    if not len(named_choices):
        named_choices = set()
    return Enum(name, named_choices, type=EnumWithDescriptionsType)
```

Python 2.7 doesn't handle enums with lists or tuples with zero entries - but works okay if we pass a set or generator. The file path field in the test has no child folders, therefor we get an empty set of choices.

ulgens (approved):

zbyte64 (approved):
:+1:

PR #1137, opened 2021-02-24
fix: declaration of required variable in filters v3

tcleonard:
This is to fix the problem you reported @ulgens  in the v3


zbyte64 (approved):
:+1:

PR #1188, opened 2021-04-20
update js version

### LINKED ISSUE #1140: Upgrade graphiql to fix autocomplete bug
It looks like graphene uses [1.0.3 ](https://github.com/graphql-python/graphene-django/blob/c049ab7470d947594e9e947a8dbde3f76fb07951/graphene_django/views.py#L70)of graphiql. Autocomplete isn't working for me for on Graphene interfaces. This is described in [this](https://github.com/graphql/graphiql/issues/1691) graphiql issue

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** 
Create an interface field in graphene and try to use autocomplete in graphiql

From the issue above, it looks like it's fixed in graphiql 1.0.6, so perhaps it should be upgraded?

Would this be a straightforward fix?

RainshawGao:
fix #1140

### DIFF
--- a/graphene_django/views.py
+++ b/graphene_django/views.py
@@ -58,23 +58,23 @@ class GraphQLView(View):
     graphiql_template = "graphene/graphiql.html"
 
     # Polyfill for window.fetch.
-    whatwg_fetch_version = "3.2.0"
-    whatwg_fetch_sri = "sha256-l6HCB9TT2v89oWbDdo2Z3j+PSVypKNLA/nqfzSbM8mo="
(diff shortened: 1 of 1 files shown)


ulgens:
Seems fine to me but i didn't make any manual testing.

ulgens (approved):

ulgens:
@RainshawGao Can you please create a v2 version of this PR too? Thanks.

RainshawGao:
@ulgens ok, I will. as for the test, I just test three or five query on the new template and it works fine.

Issue #1207, opened 2021-06-03
Backward Pagination not working correctly when using with last and before arguments


Thiyagu2009:
* **When we query the connection field with both last and before arguments it is not working as expected**

* **Steps to reproduce**

1. Querying all nodes
![image](https://user-images.githubusercontent.com/85286091/120629085-1363ce80-c483-11eb-821b-c7f32b320f26.png)

2. Copying the third node's edge's cursor and querying for before 2 nodes
![image](https://user-images.githubusercontent.com/85286091/120630051-0f847c00-c484-11eb-98e3-ddab3ff597a5.png)

* **It should return previous two nodes instead it is returning only one**

* **This issue occurs because the slicing of the edges occurs from last node(with cursor YXJyYXljb25uZWN0aW9uOjI=) instead of before node(with cursor YXJyYXljb25uZWN0aW9uOjE=)**

* **Environment:**
  
  - Version: 3.0.0b7
  - Platform: MacOs

paltman:
I've hit this too when trying to upgrade to 3x on libraries.  I've had to downgrade to 2x to keep pagination working.  I wasn't sure if I needed to update how I had pagination implemented or not but haven't seen any backward incompatible changes mentioned anywhere.  Thanks @Thiyagu2009, looks like you found and possibly fixed the root cause.  Are things working for you on your fork?

Thiyagu2009:
@paltman Yes it is working on my fork.

PR #1208, opened 2021-06-03
Issue #1207 : Backward Pagination not working correctly when using wi…

Thiyagu2009:
Avoid array_slice variable to get sliced when before argument is present


firaskafri:
@Thiyagu2009

PR #1215, opened 2021-06-09
Only catch GraphQLException

weilu:
This allows generic Exception through in case `UnforgivingExecutionContext` is used


firaskafri:
Possible duplicate https://github.com/graphql-python/graphene-django/pull/1287

PR #1254, opened 2021-08-11
Update using `client_mutation_id` for SerializerMutation 

KingNonso:
When using relay, it overrides the id of the model and provides a global id.  This id is often what is available when making queries. 
However, it is unusable to update the model object.
This seeks to make the `client_mutation_id` usable for updating the SerializerMutation

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -166,9 +166,17 @@ def convert_field_to_string(field, registry=None):
     )
 
 
+@convert_django_field.register(models.BigAutoField)
 @convert_django_field.register(models.AutoField)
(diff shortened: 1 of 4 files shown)


wodCZ:
I don't know how SerializerMutation is implemented, but using clientMutationId for identifying model to update is not how it was meant to be used.

clientMutationId is no-longer in the Relay specification, but [here](https://blog.logrocket.com/robust-graphql-mutations-the-relay-way/#:~:text=Idempotent%20mutations%20with%20clientMutationId) is an article explaining the purpose of clientMutationId. 

Merging this will be a breaking change ⚠️

PR #1256, opened 2021-08-26
Add async view

fabienheureux:
# Async view for django >= 3.1
## Description

This is still a work in progress, but based on repl from @jkimbo (https://replit.com/@jkimbo/Graphene-async#main/schema.pyhttps://replit.com/@jkimbo/Graphene-async#main/schema.py) I got an async view working for graphene. 

I think this should be part of graphene django. 

You can use it like this for now 
```python
from django.conf.urls import url
from django.views.decorators.csrf import csrf_exempt
from graphene_django.views import AsyncGraphQLView

#  ... your django urls
urlpatterns += [ url(r"^graphql$", csrf_exempt(AsyncGraphQLView.as_view(schema=schema, graphiql=True)))]
```

🖖 any kind of help / review, even at this stage is much appreciated as I am not really experienced with graphene internals, and there might be some area for improvement here. 
Thanks !

## TODO
- [ ] Add examples with async resolvers
- [ ] Add tests
- [ ] Add documentation

### DIFF
--- a/graphene_django/views.py
+++ b/graphene_django/views.py
@@ -1,4 +1,5 @@
 import inspect
+import asyncio
 import json
 import re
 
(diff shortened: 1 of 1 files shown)


erikwrede:
@Suor mixing Sync and async resolvers is handled directly by `Graphql-Core`, not by the view. So that will work.

fabienheureux:
Hey, sorry I have not been using this project since a few months, I am not sure when I will be able to catch up, I hope before the end of 2022 but I am buried under a lot of client work at the moment 😓

PR #1269, opened 2021-11-09
Fix type hint for DjangoObjectTypeOptions.model

belkka:
Proper type is `typing.Type[Model]`, not `Model`

### DIFF
--- a/graphene_django/types.py
+++ b/graphene_django/types.py
@@ -122,7 +122,7 @@ def validate_fields(type_, model, fields, only_fields, exclude_fields):
 
 
 class DjangoObjectTypeOptions(ObjectTypeOptions):
-    model = None  # type: Model
+    model = None  # type: Type[Model]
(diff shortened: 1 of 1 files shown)


belkka:
Also, there could be added type hints for `fields` and `exclude`:

```
class DjangoObjectTypeOptions(ObjectTypeOptions):
    ...
    fields = None  # type: Union[List[str], Tuple[str, ...], Literal['__all__']]
    exclude = None  # type: Union[List[str], Tuple[str, ...]]
```

This would fix mismatching type hint inherited from `ObjectTypeOptions.fields`, which is annotated as following in graphene:
```fields = None  # type: Dict[str, Field]```
 
`typing.Literal` is new in python3.8, so could use `str` instead of `Literal['__all__']`

firaskafri (approved):

jkimbo (approved):

PR #1286, opened 2021-12-17
Make errors in form mutation non nullable

artofhuman:
`DjangoModelFormMutation` does not return null in errors field. If form is valid, errors will be empty array. If form invalid, errors will be filled as `ErrorType`. 
Specification https://spec.graphql.org/June2018/#sec-Type-System.Non-Null says in this case non-null values should be returned. So it makes sense to simplify type validation on client-side.


keithhackbarth (approved):
This looks simple and straightforward enough for me. Nice work.

firaskafri (approved):

PR #1289, opened 2022-01-19
Delay assignment of csrftoken in Graphiql

c-py:
The `csrftoken` is currently assigned only when `graphiql.js` is first loaded. 

The current `csrftoken` can rotated by Django, for instance when a user logs in. [rotate_token](https://github.com/django/django/blob/3ff7f6cf07a722635d690785c31ac89484134bee/django/middleware/csrf.py#L118) performs the rotation.

When this happens, the `csrftoken` held by `graphiql.js` is invalid and Graphiql will receive CSRF errors.

This PR delays the assignment of the `csrftoken` by moving it into the `httpClient` function so when the `csrftoken` is rotated by Django, Graphiql can pick up the new token from the cookies.

### DIFF
--- a/graphene_django/static/graphene_django/graphiql.js
+++ b/graphene_django/static/graphene_django/graphiql.js
@@ -10,14 +10,6 @@
   history,
   location,
 ) {
-  // Parse the cookie value for a CSRF token
-  var csrftoken;
-  var cookies = ("; " + document.cookie).split("; csrftoken=");
-  if (cookies.length == 2) {
-    csrftoken = cookies.pop().split(";").shift();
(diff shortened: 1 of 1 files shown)


keithhackbarth (approved):
Code seems simple and well-written.

c-py:
@keithhackbarth Do you happen to know what I should do to merge this PR? I'm not authorized to do so and the build status is still in orange.

edwinvandeven:
Ran into this issue, would be great if this can be merged and included in a future release!

firaskafri (approved):

PR #1292, opened 2022-01-24
STALEBOT: Update to stale out issues after one year

keithhackbarth:
In order to better triage issues, it would be helpful to clear out old issues. I used 365 as it seems to be the commonly used configuration for other open-source projects. Example: https://github.com/integrations/slack/blob/master/.github/stale.yml

### DIFF
--- a/.github/stale.yml
+++ b/.github/stale.yml
@@ -1,7 +1,7 @@
 # Number of days of inactivity before an issue becomes stale
-daysUntilStale: false
+daysUntilStale: 365
 # Number of days of inactivity before a stale issue is closed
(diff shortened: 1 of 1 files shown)


ulgens:
Considering how often this repo gets love from its maintainers, I wouldn't recommend closing issues just because they are old. As far as I can see, most of them are still relevant.

erikwrede:
Closing this for cleanup. Agree with @ulgens, open issues should remain open for users to identity problems they occur. Especially when there's no active maintainers to consult, stale bot would add more confusion.

PR #1296, opened 2022-01-30
fixed broken links to graphene filter documentation (master->latest)

ppk42:
I checked the new links; they work. I'm aware that .../latest is a link that may point to different content in the future. But with master it was the same situation.

### DIFF
--- a/docs/filtering.rst
+++ b/docs/filtering.rst
@@ -2,8 +2,8 @@ Filtering
 =========
 
 Graphene integrates with
-`django-filter <https://django-filter.readthedocs.io/en/master/>`__ to provide filtering of results.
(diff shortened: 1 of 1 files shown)


ulgens (changes requested):
> I'm aware that .../latest is a link that may point to different content in the future.

I agree with this. Proper replacement for `master` is `main`. Broken links are the result of branch name change.

ppk42:
Thank you for the hint. Yes, why not use the most obvious renaming? ;-) Unfortunately, I'm so arragant and deleted my fork :-(. Now I have no idea how to reassign the newly created fork to the PR. So I will close this PR and open a new one.

PR #1312, opened 2022-03-02
Fix documentation references: op_name -> operation_name

pifantastic:
The documentation makes several references to an `op_name` kwarg that is actually `operation_name`.

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -27,7 +27,7 @@ Usage:
                     }
                 }
                 ''',
-                op_name='myModel'
+                operation_name='myModel'
             )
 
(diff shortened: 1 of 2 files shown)


ulgens (approved):

ulgens:
@pifantastic Welcome 🥳 

The PR seems good to me. Do we have the same problem with v2 branch?

pifantastic:
@ulgens Seems like the `v2` branch is still using `op_name` and the docs correctly reflect that: 

https://github.com/graphql-python/graphene-django/blob/v2/graphene_django/utils/testing.py#L72

PR #1320, opened 2022-04-09
Updated deprecated variable name from format_error to GraphQLFormattedError

qe:
While using `graphene-django`, I ran into the following `ImportError`:

`ImportError: cannot import name 'format_error' from 'graphql.error'`

Looking into `graphql.error`, here is an excerpt of `graphql-core`'s `/error/__init__.py` file:

```
__all__ = [
    "GraphQLError",
    "GraphQLErrorExtensions",
    "GraphQLFormattedError",
    "GraphQLSyntaxError",
    "located_error",
]
```

I believe `graphene_django/views.py`'s `format_error` variable name is most likely deprecated, so locally, I replaced `format_error` with `GraphQLFormattedError` from the excerpt above, and it is now working locally.

[Here](https://github.com/graphql-python/graphql-core/commit/09ff14f68bdce1e9cd71decc871fd38274b01971) is `graphql-core`'s commit that deprecated the `format_error` variable name

### DIFF
--- a/graphene_django/views.py
+++ b/graphene_django/views.py
@@ -11,7 +11,7 @@ from django.views.decorators.csrf import ensure_csrf_cookie
 from django.views.generic import View
 from graphql import OperationType, get_operation_ast, parse, validate
 from graphql.error import GraphQLError
-from graphql.error import format_error as format_graphql_error
+from graphql.error import GraphQLFormattedError as format_graphql_error
 from graphql.execution import ExecutionResult
 
 from graphene import Schema


NwawelAIroume (approved):
tested the change on my side and it is working on my updated project

NwawelAIroume (changes requested):
Hi tested again the your commit in my project and it is not OK. i get an error on the change you made.

```
'GraphQLError' object is not iterable
.
.
.
graphene_django/views.py", line 389, in format_error
    return format_graphql_error(error)
  File "/usr/lib/python3.8/typing.py", line 1709, in _dict_new
    return dict(*args, **kwargs)

Exception Type: TypeError at /openapiql
Exception Value: 'GraphQLError' object is not iterable
```
look at the following code  in the python graphql package to take into consideration
**python3.8/site-packages/graphql/error/graphql_error.py**

```
def format_error(error: GraphQLError) -> GraphQLFormattedError:
    """Format a GraphQL error.

    Given a GraphQLError, format it according to the rules described by the "Response
    Format, Errors" section of the GraphQL Specification.

    .. deprecated:: 3.2
       Please use ``error.formatted`` instead. Will be removed in v3.3.
    """
    if not isinstance(error, GraphQLError):
        raise TypeError("Expected a GraphQLError.")
    return error.formatted
```
so `GraphQLError` instances which are to be formatted should be `error.formatted`

Can you review your changes? Thanks

tcleonard:
I honestly don't understand why they removed it from the init because the `format_error` method still exists... they only changed the returned type to be more specific than a dict that could basically have any keys before!
so I guess the real fix is to change the import from
```python
from graphql.error import format_error as format_graphql_error
```
to 
```python
from graphql.error.graphql_error import format_error as format_graphql_error
```

63phc (approved):

erikwrede:
Fixed in #1327

PR #1325, opened 2022-06-14
Add DjangoFormInputObjectType to forms/types

ndpu:
Adds `DjangoFormInputObjectType` (`InputObjectType` derived class) which can retrieve fields from django form. Type of fields with choices (which was converted to enum automatically) is set to custom scalar type (if `object_type` is specified in `Meta`) to dynamically convert enum values back.

With `DjangoFormInputObjectType` we can build input types with data from different models, nest it in one another etc. It's just like an ordinary input object type that have same fields as django form.

Enum values converting feature can be convenient, please comment your thoughts about this.

Example:
```python
class MyDjangoFormInput(DjangoFormInputObjectType):

    # any other fields can be placed here as well as
    # other djangoforminputobjects and intputobjects

    class Meta:
        form_class = MyDjangoModelForm
        object_type = MyModelType


class SomeMutation(graphene.Mutation):
    class Arguments:
        data = MyDjangoFormInput(required=True)

    @staticmethod
    def mutate(_root, _info, data):

        # any data field (which have choices in django)
        # with value 'A_1' (for example) from client
        # is already converted to '1' here!

        form_inst = MyDjangoModelForm(data=data)
        if form_inst.is_valid():
            django_model_instance = form_inst.save(commit=False)

        # ...

```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,3 +80,8 @@ Session.vim
 tags
 .tox/
 .pytest_cache/
+
+# pyenv
+#   For a library or package, you might want to ignore these files since the code is
+#   intended to run in multiple environments; otherwise, check them in:
+.python-version
--- a/docs/mutations.rst
+++ b/docs/mutations.rst
@@ -132,6 +132,9 @@ DjangoFormInputObjectType
 
 .. code:: python
 
+    from graphene_django.forms.types import DjangoFormInputObjectType
+
+
     class PetFormInput(DjangoFormInputObjectType):
         # any other fields can be placed here as well as
         # other djangoforminputobjects and intputobjects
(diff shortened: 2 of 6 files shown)


ndpu:
@firaskafri hi, ok, will do

firaskafri:
> @firaskafri hi, ok, will do

Hello there! Any updates?

ndpu:
@firaskafri fixed tests and pre-commit hook error. Should be ok now. (force-pushed rebased branch...)

firaskafri:
@ndpu would you also add your example to the docs please?

ndpu:
@firaskafri hi, docs added. I have an error with sphinx 1.5.3 and all is ok with latest version (7.0.0). I think docs/requirements.txt should be updated...

firaskafri (approved):

PR #1326, opened 2022-06-20
Update import name from format_error to graphql_error.format_error

schnea:
When using graphene-django with `graphql-core 3.2.1`, I run into the following ImportError:

`ImportError: cannot import name 'format_error' from 'graphql.error'`

Seems format_error has been removed from the `error/__init__.py`, but it still exists. 
See also the discussion on https://github.com/graphql-python/graphene-django/pull/1320


erikwrede:
Fixed in #1327

PR #1340, opened 2022-08-18
pre-commit & black

ulgens:
* Moves black into the pre-commit wrapper
* Applies black to all files in the repo


ulgens:
@erikwrede @nikolaik 👋🏻 🌹

erikwrede (approved):
Looks good, there can't be much wrong with black formatting!

But again, we should aim to unify or configs to have some consistency over all the projects:
https://github.com/graphql-python/graphene/blob/master/.pre-commit-config.yaml

+ why do we still have a makefile in this project?

ulgens:
> we should aim to unify or configs to have some consistency over all the projects:

I agree. I will create another PR to update hooks in graphene repo. And in this repo, we can re-order hooks to match ordering in the graphene repo.

> why do we still have a makefile in this project?

Probably because nobody removed it yet. It doesn't break anything, so IMO it's low priority.

PR #1341, opened 2022-08-20
Sync trove classifiers with tox.ini

ulgens:


ulgens:
@erikwrede @nikolaik 👋🏻 🌷

setup.py
```diff
@@ -19,7 +19,6 @@
     "coveralls",
     "mock",
     "pytz",
-    "django-filter<2;python_version<'3'",
     "django-filter>=2;python_version>='3'",
```
nikolaik, line 22: can this be unconditional now, since we are python 3 only?

nikolaik (commented):
Awesome!

Should we also go ahead with some of the changes in #1337 ?

ulgens:
> Should we also go ahead with some of the changes in https://github.com/graphql-python/graphene-django/pull/1337 ?

This PR for the legacy v2 branch. We can handle similar changes for main in a different one.

nikolaik:
> > Should we also go ahead with some of the changes in #1337 ?
> 
> This PR for the legacy v2 branch. We can handle similar changes for main in a different one.

Ah, I see, then this makes sense

setup.py
ulgens: Done.

erikwrede (approved):

PR #1350, opened 2022-09-23
Revert "fix: unit test for graphene pr#1412"

firaskafri:
Reverts graphql-python/graphene-django#1315 due changes in setup.py that probably broke tests


erikwrede (commented):
Wouldn't it be more elegant to push a fix to setup.py instead of reverting the entire PR and reopening it without the changes to setup.py?

firaskafri:
> Wouldn't it be more elegant to push a fix to setup.py instead of reverting the entire PR and reopening it without the changes to setup.py?

https://github.com/graphql-python/graphene-django/pull/1352

@erikwrede Thank you!

PR #1355, opened 2022-09-26
☂️ v3.0.0 ☂️

firaskafri:


erikwrede (approved):
🎉

ulgens:
:/ Wasn't it too fast for a major release, come on guys. At least give couple days to other maintainers.

ulgens:
Also, when and where people decided to make a new release with the current state of the repo? Slack is completely silent. Please give a quick update before making it. I understand the enthusiasm but there are better ways to handle this.

tcleonard:
What do you mean by "the current state of the repo".
I agree we should have waited for a couple of maintainers before cutting a release. 
But that one has been baking for a long time and we still didn't have a release compatible with graphene V3 leading for confusion in the community.
@ulgens anything that you think was missing before releasing this V3 versions? All functionality previously supported by the V2 are working in V3 so I don't really see a reason to wait... And tags are cheap! We can patch after if need be

firaskafri:
Hi @ulgens 
The release has been practically ready since a little bit less than two years, checklist for the release is already done as per this issuer here https://github.com/graphql-python/graphene-django/issues/705 .

Basically if we are releasing this today, then yesterday was the best time to do it. So as pointed out by @jkimbo the sooner we release this the better.

I agree with you that this happened too fast, but this is only for getting this specific release out. We can all agree on a better process structure of releasing for any upcoming release.

PR #1361, opened 2022-10-12
Fix custom foreignkey resolvers

syrusakbary:
The PR #1315 had a regression when using async resolvers for the `ForeignKey`.

This PR fixes this, as well as usage of `assert_valid_name` as this method no longer exists in the `graphql` library


syrusakbary:
Merging directly into master as it seems all tests are passing and this issue is blocking Wasmer using the latest graphene django.

Please feel free to revert if any issue comes up from merging the PR

PR #1363, opened 2022-10-27
Updating for Django 4.x

jkatzer:


erikwrede:
Thanks for the contribution! Could you please explain why your change is necessary and what side effects it might have on other versions of django used with this library?

ulgens:
@erikwrede, I'm not sure accepting contributions to v2 is a good idea. @jkatzer Is there any specific reason you / your team using v2 and not the latest release?

jkatzer:
Howdy @erikwrede & @ulgens,

First off, thanks to you and the team for delivering 3.0! 🎉 

Was investigating the viability of upgrading our Django version. Making a simple patch in the web editor made sense to document what we were doing and why.

There is no reason you would need to bring this in on the v2 branch... unless others see this and want it as well.

Is there an upgrade guide for 3.0? Going to check it out after our Django version bump.

-@jkatzer

PR #1370, opened 2022-11-11
chore: update to graphiql 2.x

dopry:


ulgens:
@dopry Can you please rebase your branch from current main? Thank you.

dopry:
@ulgens, done. Any other feedback?

klowe0100:
Oh sorry, I should have found this issue before starting https://github.com/graphql-python/graphene-django/pull/1371

I was focused on at least 1.4.7 because it addressed a security bug. Personally I ended up configuring to use 2.1.0. I'm sort of questioning if it's even useful to make it configurable. I don't foresee too many backwards compatibility issues to worry about. 

I can update the default to 2.x if we think that's the right idea and keep it configurable. Or we can just abandon 1371 and go with bumping the version in this branch. Either way, glad to see the project alive and want to help out!

dopry:
I'm not sure that a configurable version is a good idea because there other dependencies that might need to be updated as well.  I suspect it would create support problems when users pick a version that doesn't work with the built in react or fetcher.  From an end user perspective it should largely just work.  2.x is a nice upgrade from 1.x with multiple tabs and schema refresh without reload.  The layout is reversed which is a little jarring, until you realize it's much nicer to have the docs right next to your query, rather than separated from it by the output.

dopry:
@firaskafri why was this closed? I've been waiting on .maintainer feedback on this issue.

firaskafri:
Hi @dopry,
It was close in favor of #1396 as it is more recent.
Thank you!

PR #1376, opened 2022-11-27
use graphiql single html file with cdn

khasbilegt:
GraphiQL provides a single HTML file using CDN assets and a script tag. By using this, graphene view doesn't get polluted by all those package versions and serve a single static file that was previously used only by the template, with the benefit of getting the latest version of GraphiQL.


khasbilegt:
Hi @ulgens, what do you think about this change? Thanks in advance.

Issue #1379, opened 2022-12-16
Proxied models dont have associated many to many relationship access


TomsOverBaghdad:
This issue is loosely related to https://github.com/graphql-python/graphene-django/issues/319 to provide further support for proxy models

* **What is the current behavior?**

Currently, django models that are a proxy of another model don't include their many to many relationships that you get by default without proxy

So if my models.py looks like this

```python
# models.py

from django.db import models

class Publication(models.Model):
    ...

class SciencePublication(Publication):
    class Meta:
        proxy = True

class Article(models.Model):
    publications = models.ManyToManyField(
        Publication, 
        related_name="articles"
    )
```

and my graphene types in types.py

```python
# types.py

import graphene_django
from .models import (
    Publication as PublicationModel, 
    SciencePublication as SciencePublicationModel
)

class Publication(graphene_django.DjangoObjectType):
    class Meta:
        model = PublicationModel

class SciencePublication(graphene_django.DjangoObjectType):
    class Meta:
        model = SciencePublicationModel
```

When attempting to access the `articles` relationship through the publication type in the graphql query I can only access it via the Publication graphql type but not through the SciencePublication type

this is fine...
```gql
query {
  getPublications() {
    articles {
      id
    }
  }
}
```
this is errors
```gql
query {
  getSciencePublications() {
    articles { <<<< cannot query field "articles" on type "SciencePublication"
      id
    }
  }
}
```

* **What is the expected behavior?**

many to many relationships persist with proxy models so that this query doesn't error
```gql
query {
  getSciencePublications() {
    articles { # no errors, just smiles :)
      id
    }
  }
}
```

* **Please tell us about your environment:**

  - Version: 3.0.0b7

PR #1380, opened 2022-12-19
Support base class relations and reverse for proxy models

TomsOverBaghdad:
Includes the proxy's base model when determining the reverse fields for a django model

The potential fix for this issue https://github.com/graphql-python/graphene-django/issues/1379


firaskafri:
It would be great if the solution would be more generic to handle possible base classes such as [django-polymorphic](https://github.com/django-polymorphic/django-polymorphic)

```
    for base in model.__bases__:
        if is_valid_django_model(base):
            model_ancestry.append(base)
```

as opposed to:

```
    if model._meta.proxy:
        model_ancestry.append(model._meta.proxy_for_model)
```

TomsOverBaghdad:
@firaskafri Thanks for the suggestion! I implemented it with tests. Let me know if the test changes imply correctly what you were expecting.

firaskafri:
@TomsOverBaghdad Will definitely do!

Though I wanted to ask you, what the behavior would be if base class had an m2m field, not a reverse one?

TomsOverBaghdad:
@firaskafri if base class `A` has many `B` then class `C(A)` that inherits from A will also have many `B`, I updated the query_test to show.

firaskafri:
@TomsOverBaghdad Looks pretty good to me.

firaskafri:
> @firaskafri if base class `A` has many `B` then class `C(A)` that inherits from A will also have many `B`, I updated the query_test to show.

Yes reverse m2m is working correctly with this PR. But have you got the chance to deal with base m2m relation?

This PR addresses the reverse relation issue correctly I guess where if Z has an m2m with A, `A` and `B(A) the field will be added to the node.

But the example you provided is different, you are assuming that the field exists in A as an original field not a reverse relation.

graphene_django/utils/utils.py
```diff
@@ -38,17 +38,24 @@ def camelize(data):
...
+        for name, attr in _model.__dict__.items():
+            # Don't duplicate any local fields
+            if name in local_field_names:
+                continue
+
+            # "rel" for FK and M2M relations and "related" for O2O Relations
+            related = getattr(attr, "rel", None) or getattr(attr, "related", None)
+            if isinstance(related, models.ManyToOneRel):
+                yield (name, related)
+            elif isinstance(related, models.ManyToManyRel) and not related.symmetrical:
```
firaskafri, line 57: @TomsOverBaghdad this being inside a loop for base classes will break relations, I think its safer to isolate getting base fields in a different method as in this case `ManyToManyRel` is two sided, in the case of `A` it is `ManyToManyRel.related` while in the case of B(A) it is `ManyToManyRel.field`.

Try having the m2m field in the base class just like the example you mentioned in the comment, not a reverse relation

firaskafri: Something like this:

```
elif isinstance(related, models.ManyToManyRel) and not related.symmetrical:
    if _model is not model:
        yield (name, getattr(attr, "field", None))
        continue
```

But my preference would still be to get base models relations and reverse relations in a different method

TomsOverBaghdad: Heard! Good catch @firaskafri ! Let me know what you think of the changes I pushed. I separated out getting local fields and included the models ancestry similar to the reverse fields.

graphene_django/utils/utils.py
```diff
@@ -58,17 +92,14 @@ def maybe_queryset(value):
...
-
-    # Make sure we don't duplicate local fields with "reverse" version
-    local_field_names = [field[0] for field in local_fields]
+    """
+    Gets all the fields and relationships on the Django model and its ancestry.
+    Prioritizes local fields and relationships over the reverse relationships of the same name
+    Returns a tuple of (field.name, field)
+    """
+    local_fields = get_local_fields(model)
+    local_field_names = {field[0] for field in local_fields}
```
TomsOverBaghdad, line 101: I changed this to a set, but I can totally change it back to list to not break anything. It looks like `get_reverse_fields` is public so I didn't want to mess with the signature too much.

firaskafri (approved):

TomsOverBaghdad:
@firaskafri is there anything else I can do for this PR based on the `changes requested` tag? Seems like the failing tests are expected based on failing with previous merged prs.

firaskafri:
> @firaskafri is there anything else I can do for this PR based on the `changes requested` tag? Seems like the failing tests are expected based on failing with previous merged prs.

I just updated the tags. We're just waiting for more reviews.

jaw9c:
Great PR - would it be possible to use something from the Django lib [like this](https://docs.djangoproject.com/en/4.2/ref/models/meta/#retrieving-all-field-instances-of-a-model) to pull the fields from the model rather than rolling our own?

firaskafri:
> Great PR - would it be possible to use something from the Django lib [like this](https://docs.djangoproject.com/en/4.2/ref/models/meta/#retrieving-all-field-instances-of-a-model) to pull the fields from the model rather than rolling our own?

I investigated that, the problem with get_fields is that it does not get the local ManyToMany relationships.

graphene_django/utils/utils.py
sjdemartini: Seems like that was the right move to me, since you were able to keep the `get_reverse_fields` signature the same but improve lookup time-complexity for this internal use-case in `get_model_fields`, nice!

sjdemartini (approved):
This looks great! Appreciate the thorough tests, and your effort to make the functions in `utils.py` better documented and more modular.

lucas-bremond:
I believe this PR may have introduced a "regression" when combined with [django-polymorphic](https://github.com/django-polymorphic/django-polymorphic).

In Django Polymorphic, the parent type exposes accessors to the derived types, which are now picked up by the updated `get_reverse_fields` logic and added to the derived type fields.

Here's an example of what I mean. Assuming the following models:

```py
from polymorphic.models import PolymorphicModel

class Vehicle(PolymorphicModel):
    id = UUIDField()

class Car(Vehicle):
    ....

class Truck(Vehicle):
    ....
```

now the `Car` and `Truck` types will expose `car` and `truck` fields (that came from the parent). And these fields don't have associated resolvers, so the following GraphQL query will raise an Exception:

```graphql
query {
  cars {
    id
    # The presence of these two fields makes no logical sense, and resolving them raises
    car { id }
    truck { id }
  }
}
```

I suppose this is not really Graphene Django's concern to ensure downstream compatibility, but since Django Polymorphic is a quite popular library, I believe this is worth reporting here?

PR #1381, opened 2022-12-20
feat: support django 2

thibaut-lo:
## Main change

* Relax constraint to support django 2.2

## Other changes

**I had to batch replace `py.test` by `pytest` for `make tests` to pass locally.** :
* This has also been applied in the main graphene-django repo last month
* Ref: graphene-django commit: https://github.com/graphql-python/graphene-django/commit/86c5309c4537b8f898a4cf8f2b516efa14be18f8 and  https://github.com/pytest-dev/pytest/issues/10459

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,22 +1,21 @@
+.PHONY: help
+help:
+	@echo "Please use \`make <target>' where <target> is one of"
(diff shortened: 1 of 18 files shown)


ulgens:
@thibaut-lo 2.2's EOL was April 2022 and we remove the support from the repo on purpose.

Issue #1387, opened 2023-01-31
There is no current event loop in thread 'Thread-2 (process_request_thread)'.


ewen-lbh:
* **What is the current behavior?**

A simple query fails with a cryptic, seemingly asyncIO-related error:

	{
	  "errors": [
	    {
	      "message": "There is no current event loop in thread 'Thread-2 (process_request_thread)'.",
	      "locations": [
	        {
	          "line": 34,
	          "column": 3
	        }
	      ],
	      "path": [
	        "tests"
	      ]
	    }
	  ],
	  "data": {
	    "tests": null
	  }
	}

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via

Link to reproduction repository: https://github.com/ewen-lbh/bug-report-django-graphene

Steps to reproduce:

1. `git clone https://github.com/ewen-lbh/bug-report-django-graphene` 
2. ` cd bug-report-django-graphene` 
3. ` poetry install`  (or ` pip install -r requirements.txt` )
4. ` cd repro` 
5. ` poetry run python manage.py migrate && poetry run python manage.py runserver` 
6. Navigate to ` 127.0.0.1:8000/graphql` 
7. Try executing the following query:  
	
	
		{
			tests {
				testing
			}
		}

* **What is the expected behavior?**

The query executes and returns its result

* **What is the motivation / use case for changing the behavior?**

* **Please tell us about your environment:**

  - Version: graphene-django 3.0.0, django 4.1.5, python 3.10.9
  - Platform: Manjaro Linux 5.19.17-2-MANJARO

* **Other information** (e.g. detailed explanation, stacktraces, related issues, suggestions how to fix, links for us to have context, eg. stackoverflow)

bryantjamestech:
I'm seeing a similar issue when attempting to update. Any thoughts?

mbrochh:
Same here. What I know so far is that the endpoing does work in principal, but fails only for certain objects. In my case, my endpoint should return several fields, one of them being `product`. Via ipdb I can see that the product was successfully queried and returned by my resolver, but it seems that graphene fails to properly serialize it. I stumble upon this while making test fixtures with mixer, so I'm dealing with a product that might have certain fields or values set to funky values, so that might be the reason why graphene seems to crash. So my guess right now would be that graphene crashes when it tries to serialize an object and then swallows the real error message.

bryantjamestech:
[...]

PR #1388, opened 2023-02-13
Remove promise based debug middleware

jaw9c:
The exception handling by the debug middleware is broken. If using the toolbar middleware, then any exception raised in a resolver gets consumed and replaced with a cryptic `Thread pool` related error.

The root cause is that the graphql core lib expects the result of the call to exception instance rather than a rejected promise.

When the graphql core lib tries to resolve the promise as a normal result it results in the cryptic error.

I've also removed the reference to promises as they didn't seem to be helping anything (But please correct the PR if I'm missing something)

### DIFF
--- a/graphene_django/debug/middleware.py
+++ b/graphene_django/debug/middleware.py
@@ -9,32 +9,32 @@ from .types import DjangoDebug
 
 class DjangoDebugContext:
     def __init__(self):
-        self.debug_promise = None
-        self.promises = []
+        self.debug_result = None
+        self.results = []
         self.object = DjangoDebug(sql=[], exceptions=[])
         self.enable_instrumentation()
 
-    def get_debug_promise(self):
-        if not self.debug_promise:
-            self.debug_promise = Promise.all(self.promises)
-            self.promises = []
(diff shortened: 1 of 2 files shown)


jaw9c:
Poitentially fixes https://github.com/graphql-python/graphene-django/issues/1387

firaskafri:
Hello @jaw9c, this is great!
Please check the tests as they're failing

firaskafri (approved):

jaw9c:
Thanks @firaskafri  - looks like the CI is broken as the tests pass locally. 

Just looked at them and appears to be an import error with the examples folder - https://github.com/graphql-python/graphene-django/actions/workflows/tests.yml

jaw9c:
@firaskafri Updated the PR, tests now get run correctly - was a problem with the python path not being set 👍

jkimbo (approved):
I'm not that familiar with this part of the codebase but the changes look fine to me.

firaskafri:
> I'm not that familiar with this part of the codebase but the changes look fine to me.

Thank you @jkimbo 🙏

Issue #1375, opened 2022-11-25
DjangoObjectType.fields ignores select_related


lufte:
* **What is the current behavior?**

The behavior for resolving a ForeignKey field of a DjangoObjectType is not the same when adding the field to `fields` versus declaring the field explicitly as an attribute of the class. In the second case, an extra query won't be made if the field has been cached by a previous `select_related`, but the first case always executes an additional query ignoring the `select_related`.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via
a github repo, https://repl.it or similar (you can use this template as a starting point: https://repl.it/@jkimbo/Graphene-Django-Example).

Create two Models with a foreign key between them, and a schema in which you can list all the parent models with their respective children, using a `select_related` call in the resolver. In the parent DjangoObjectType, try declaring the child model using both `fields` and an explicit attribute like `child = graphene.NonNull('schema.ChildType')`, and count the queries using both strategies.

* **What is the expected behavior?**

The `select_related` field should be honored in both cases.

* **Please tell us about your environment:**

  - Version: 3.0.0
  - Platform: Python 3.10.8, graphene==3.1.1

grantmcconnaughey:
I can confirm this is also happening for me using graphene-django 3.0.0 and graphene 3.2.1.

Before the upgrade `select_related` was used, but now there are N+1 errors all throughout my project. I can also confirm that adding an explicit attribute does fix the issue.

@lufte Were you ever able to find a more global fix? I can certainly explicitly add attributes to all of my types but it would be simpler to fix this in one spot.

lufte:
@grantmcconnaughey unfortunately no, I just added the fields explicitly and forgot all about it.

firaskafri:
@lufte we have reverted a possible cause #1315, can you confirm if the issue still exists in 3.0.1?

lufte:
Hi @firaskafri. I ran a quick test using my current project (I thought it would be faster than creating a minimal reproducible example, turned out that it isn't) but I still see the behavior. If you need a demo to properly reproduce the issue let me know and I can write one.

sjdemartini:
[...]

PR #1391, opened 2023-03-07
fix(converter.py): performance fix in `convert_field_to_djangomodel.dynamic_type.CustomField

mahdyhamad:
Fix performance issue in resolving one-to-one and foreign-key field, make only one database hit to get the field instance while resolving the field.

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -1,3 +1,5 @@
+import functools
+import inspect
 from collections import OrderedDict
 from functools import singledispatch, wraps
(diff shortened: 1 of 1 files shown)


mahdyhamad:
Related issues:
- https://github.com/graphql-python/graphene-django/issues/1111

PR caused the issue: https://github.com/graphql-python/graphene-django/pull/1315/

grantmcconnaughey:
This seems to be related to #1375

PR #1394, opened 2023-04-02
[WIP] Support for Async Django

jaw9c:


jaw9c:
Progress update: All the fields resolvers are supporting both being executed in the async context and wrapping django sync code into a sync context. I've made the assumption that any resolving of `DjangoXType/Field` fields will need to be executed in a sync context which I think is pretty safe. I've started running through all the tests and adding an extra assertion that running in an async executor produces the same result as the sync one. 

Currently people switching to using the async view will need to wrap all their top level resolvers (those that are directly on the `Query` class passed into the schema) in `@sync_to_async` if they run sync code. If someone has an idea of how to prevent this that would be great, but can't seem to wrap my head around that! 

Next steps are to run through all the test files and adding the helper. 

It would great if people could test this either on their production projects (speciically when running under ASGI), as 'm interested in how swapping in/out of the sync context affects the performance of the execution.

firaskafri:
> Progress update: All the fields resolvers are supporting both being executed in the async context and wrapping django sync code into a sync context. I've made the assumption that any resolving of `DjangoXType/Field` fields will need to be executed in a sync context which I think is pretty safe. I've started running through all the tests and adding an extra assertion that running in an async executor produces the same result as the sync one.
> 
> Currently people switching to using the async view will need to wrap all their top level resolvers (those that are directly on the `Query` class passed into the schema) in `@sync_to_async` if they run sync code. If someone has an idea of how to prevent this that would be great, but can't seem to wrap my head around that!
> 
> Next steps are to run through all the test files and adding the helper.
> 
> It would great if people could test this either on their production projects (speciically when running under ASGI), as 'm interested in how swapping in/out of the sync context affects the performance of the execution.

This is great news! Would discuss with my team to try this on production once its ready!

jaw9c:
@firaskafri Ready to go for reviews on this one! Managed to solve the problem of wrapping everything in `sync_to_async` by utilising a middleware. If an implementing project would like to use async they only need to do the following:

- Swap out to using the Async View
- Add the Async middleware
- Check their schema for any types that are not extended from `DjangoObjectType` that have a resolver that uses the Sync parts of Django (including the top level `Query` type).
  - If these exist, when called they result in [SynchronousOnlyOperation](https://docs.djangoproject.com/en/4.2/ref/exceptions/#django.core.exceptions.SynchronousOnlyOperation) exception. 
  - To fix this, the user has to wrap the resolver in `sync_to_async`. 
- 🚀 Done, enjoy the speedups!!
  
Looking forward to feedback!

pcraciunoiu:
GraphiQL wasn't working for me on this branch, so I pushed a small change here https://github.com/UpliftAgency/graphene-django/commits/support-async (specific commit https://github.com/UpliftAgency/graphene-django/commit/8368ae2d86604995ef6ccd051d8b724b69a5bc35)

So far tested it with data loaders for a sample app and everything seems to be working. Thanks for your work on this!

pcraciunoiu:
I spoke too soon, there is one issue I need to narrow down - sometimes related fields try to use sync context and break, e.g. I have a `blog` that has `articles`, and each `article` has an `author`. Sometimes `article.author` seems to get resolved sync using [`attr_resolver`](https://github.com/graphql-python/graphene/blob/8ede21e06381c096589c424960a6cfaca304badb/graphene/types/resolver.py#L1) and that doesn't work in async context.

I haven't tried to isolate the issue yet because it doesn't always happen, but it could be fine to worry about it in a follow-up and get this merged!

firaskafri:
I think I will start testing next week on a large project, will get back with a feedback in it soon!

dima-kov:
Hi team! Thank you all for your great input and initiative! Could I ask please what are the plans to have this feature released? Is it feasible to release it in one-two month?

dolgidmi:
@firaskafri 
> I think I will start testing next week on a large project, will get back with a feedback in it soon!

Any news on testing?

jaw9c:
Thanks for the input and testing guys! Super appreciated.  Apologies for the lack of progress on this from me! Have been in firefighting mode recently and this slipped down the priorities.

The state of play from before is that I was testing the performance of async vs sync, both running under asgi and wsgi. During this I found a bug which made the app hang when running async on ASGI locally - I was struggling to debug this and is the first thing to focus on next. 

If you'd like to help - the best way is to test on your own  projects and get some relative performance stats under load that would be great. I'll likely be building a tool for my own projects that I can put in a gist but this will take time vs testing manually and I'd like to squash any complexity/bugs ASAP.

TL;DR; sorry for pausing work on this; if you want to help progress it, please test on your own projects and report any issues here.

jaw9c:
Good news: Fixed the hang issue. Bad news: The amount of time I wasted figuring out it was only an async incompatible middleware on our work env. I've run some preliminary load testing on our staging env, and got 2x throughput on resolving a queryset. Next up is some analysis on run times, I'll post these in a digestible format - hopefully by next week. Keen to get this merged soon.

dima-kov:
> Good news: Fixed the hang issue. Bad news: The amount of time I wasted figuring out it was only an async incompatible middleware on our work env. I've run some preliminary load testing on our staging env, and got 2x throughput on resolving a queryset. Next up is some analysis on run times, I'll post these in a digestible format - hopefully by next week. Keen to get this merged soon.

Faced exactly the same issue, having only one sync middleware makes django hang on async handlers

cpd67:
@jaw9c Thanks a lot for opening this up! It would be really, really nice if this could get reviewed and merged soon, as this is blocking us from being able to upgrade to the latest version of graphene-django.

examples/cookbook/cookbook/urls.py
```diff
@@ -1,9 +1,10 @@
...
+from django.views.decorators.csrf import csrf_exempt
 
-from graphene_django.views import GraphQLView
+from graphene_django.views import AsyncGraphQLView
 
 urlpatterns = [
-    url(r"^admin/", admin.site.urls),
-    url(r"^graphql$", GraphQLView.as_view(graphiql=True)),
+    re_path(r"^admin/", admin.site.urls),
+    re_path(r"^graphql$", csrf_exempt(AsyncGraphQLView.as_view(graphiql=True))),
```
superlevure, line 9: Do we want to make the async view the only, and default, one in the cookbook? I feel we should have both examples present, the sync one will most likely still be the most common one.

graphene_django/debug/middleware.py
```diff
@@ -67,3 +70,28 @@ def resolve(self, next, root, info, **args):
...
+            [
+                hasattr(parent_type, "graphene_type")
+                and hasattr(parent_type.graphene_type._meta, "model"),
+                hasattr(return_type, "graphene_type")
+                and hasattr(return_type.graphene_type._meta, "model"),
+                info.parent_type.name == "Mutation",
+            ]
+        ):
+            if is_sync_function(next):
+                return sync_to_async(next)(root, info, **args)
```
superlevure, line 95: Based on what I understand from the [documentation](https://github.com/django/asgiref), this is going to wrap any synchronous resolver that accesses the DB into a separate thread, isn't it?

If this is the case indeed, isn't this going to have a substantial bad impact on performances?

superlevure (commented):
Thank you for the work on this PR! 

Before merging, I think we should have benchmarks available to make sure it does not introduce performance loss for synchronous contexts.

I'm curious about your statement:

> [...] got 2x throughput on resolving a queryset

Could you share details on how that was measured? 

Also, before merging, the documentation needs to be updated as well

graphene_django/debug/middleware.py
jaw9c: For CPU bound resolvers the overhead will make the performance a lot worse. Since most Django apps are backed by an external DB, most resolvers are IO bound - so the advantage comes from the main loop can continue to serve other requests/continue resolving other parts of the query.

dima-kov: agree with @jaw9c, also this is the main approach of Django core team on the way to making Django async

superlevure: While I agree most resolvers are IO bound, we still miss benchmarks to assert that performance will improve with this pattern. The cost of creating a thread for each resolver + event loop task context switching VS blocking DB calls tradeoff is not obvious at all and I wouldn't be surprised if the synchronous version performs better for simple queries.

dima-kov: no, it still makes sense to support `sync_to_async` mode for those projects that need async support. Benchmarks for sync resolvers run under `AsyncGraphQLView` (ASGI) will be definitely worse than sync resolvers under  `GraphQLView` (WSGI) as you said.

But this PR is for projects with async resolvers running under `AsyncGraphQLView` (ASGI) - there is no possibility to run such code without this PR now

examples/cookbook/cookbook/urls.py
cpd67: @jaw9c I can help add an example with the sync view. I think having both examples present makes sense.

graphene_django/debug/middleware.py
cpd67: @jaw9c It sounds like you have a way to get some sort of benchmarks to show the performance impact, but if not I can help add those. I'm not sure what the best way would be to get them, but I can do my best :smile:

examples/cookbook/cookbook/urls.py
cpd67: @jaw9c I created a PR to have both an async & sync example of the cookbook. I basically took these changes, stuck them in `cookbook-async` and put the sync `GraphQLView` back. Should be good to merge into this PR, but let me know what you think!

firaskafri:
@jaw9c any updates?

pfcodes:
Really need this!

graphene_django/debug/middleware.py
alexandrubese: @superlevure but what is the alternative to use for dataloaders ? In the documentation there are only async dataloaders mentioned.

Dataloaders are an important part of any graphql API, I don't understand how the upgrade was done to v3 without taking into consideration dataloaders?!

There is the sync dataloaders package(https://github.com/jkimbo/graphql-sync-dataloaders) but that is an external package, what solution did you guys have in mind for dataloaders in the new graphene versions?

dima-kov:
Hey mates, ready to donate some bucks via buymeacoffee if you make this PR to live!

graphene_django/debug/middleware.py
superlevure: Agreed, the current situation regarding dataloaders is unfortunate. We've been working on adding them back on v3 thanks to `graphql-sync-dataloaders` on our fork of `graphene-django` at my company. You can check out our tentative here:

- https://github.com/loft-orbital/graphene-django/pull/37
- https://github.com/loft-orbital/graphene-django/pull/44

We've been using this in production for a few months now with success, our plan is to eventually submit this to upstream when we are confident enough and we have the bandwidth.

superlevure: Note that our implementation of dataloaders for many-to-one relations is not covered in those two PRs as it's implemented in a private repo, but it basically looks like this:

```python
    @classmethod
    def get_node(cls, info, id):
        if not graphene_settings.USE_DATALOADERS:
            return super().get_node(info, id)

        model_name: str = cls._meta.model._meta.model_name
        queryset: QuerySet = cls.get_queryset(cls._meta.model.objects, info)

        if not hasattr(info.context, "dataloaders"):
            info.context.dataloaders = {}

        if model_name not in info.context.dataloaders:

            def load_many(keys: list[UUID | str]):
                qs: QuerySet = queryset.filter(id__in=keys)
                object_map: dict[str, Any] = {str(object_.id): object_ for object_ in qs}

                return [object_map.get(str(key), None) for key in keys]

            info.context.dataloaders[model_name] = SyncDataLoader(load_many)

        return info.context.dataloaders[model_name].load(id)
```

pfcodes:
> Hey mates, ready to donate some bucks via buymeacoffee if you make this PR to live!

I'm down to throw some $ too. Maintainers please consider enabling sponshorship for this project if it means that things can move more quickly. A lot of people are dependent on this.

alexandrubese:
I don't know man... Migrating to v3 version, without making sure dataloaders and everything works, so that people need to implement new libraries and workarounds?! 

This bad decision made by the people of a package that is used by so many, without support to one of the core most important things of working with graphql.

To the point that people started offering money to get this to work, this is crazy.

graphene_django/debug/middleware.py
alexandrubese: I did not realise that for many-to-one relations you actually need another fix to get this working.
At this point, if someone does not fix dataloaders, I will just start using another graphql library because its getting ridiculous.

alexandrubese:
@dima-kov 
Use an older version of the library where things work? 

Which might not be ideal because you’re basically just adding “legacy code” that you will need to update once they will make v3 work (although it’s been almost a year this is discussed, I don’t know when this will work and they will fix dataloaders) 

Or use a different library

dima-kov:
Regarding the comment from @superlevure:

> I agree with releasing AsyncGraphQLView.

We are almost there; the only remaining tasks are performance benchmarks and documentation updates. This implies that most of the work is already done, and only a few percentages are left.

@jaw9c, it seems you might be short on time for this. I'm willing to contribute efforts at this point.

@superlevure, to be transparent, I haven't conducted performance testing before, but I'm attempting it now. Any suggestions on how to approach this appropriately would be highly appreciated.

dima-kov:
Benchmarks:
```16.547 seconds vs 83.194 seconds for i/o bound queries. 5x faster. async wins```

Details are here: https://github.com/dima-kov/django-graphene-benchmarks?tab=readme-ov-file#tldr

Async:
```
Concurrency Level:      10
Time taken for tests:   16.547 seconds
Complete requests:      1000
```

vs 

Sync
```
Concurrency Level:      10
Time taken for tests:   83.194 seconds
Complete requests:      1000
```

@superlevure are we fine now? What should be next steps to make this public?

**Review by dima-kov — APPROVED** [review]
🔥

**Comment by kamilglod** [comment]
@dima-kov shouldn't you run sync benchmark with 10 workers instead of 4? I know that measuring sync vs async is hard but with 10 workers we would have consistent number of requests that are handled in the same time by the server so we should get actual comparison. I guess we should get similar results which is fine as the biggest benefit of using async is less resources used.

**Comment by dima-kov** [comment]
hmm, that would eat lots of memory. 
from docs:

>Gunicorn should only need 4-12 worker processes to handle hundreds or thousands of requests per second. Gunicorn relies on the operating system to provide all of the load balancing when handling requests. Generally we recommend (2 x $num_cores) + 1 as the number of workers to start off with.

I'm running on 4 cores machine, so trying gunicorn with 9 workers I've got only slightly better result:
```
Concurrency Level:      10
Time taken for tests:   80.785 seconds
Complete requests:      1000
Failed requests:        0
```
and memory usage was ~80mb for 9 processes + 1 process 20mb (gunicorn) = 720. The bottleneck here might be sqlite3 database.

-------

vs async:
```
Concurrency Level:      10
Time taken for tests:   17.586 seconds
Complete requests:      1000
Failed requests:        0
```
and mem usage was 1 process 60mb.

**Comment by kamilglod** [comment]
Sqlite docs says that concurrent writes are locked, but reads should be fine so I think it's not a problem with sqlite itself. 

Maybe it's because of the differences in sync and async resolvers? In async you're fetching related `octopus_type`, in sync not. 
https://github.com/dima-kov/django-graphene-benchmarks/blob/main/project/api/schema/queries.py#L43-L51
https://github.com/dima-kov/django-graphene-benchmarks/blob/main/project/api/schema_async/queries.py#L48-L64

**Comment by dima-kov** [comment]
oh, shame on me. so here is fixed version comparison:


## Comparison

|             | **Sync** | **Async** | | **Sync** | **Async** | |
|-------------|----------|-----------|-|----------|-----------|-|
| Requests    | 1000     | 1000      | | 1000     | 1000      | |
| Concurrency | 100      | 100       | | 100      | 100       | |
| Processes   | 9        | 1         | | 4        | 1         | |
| Threads     | 1        | 100       | | 1        | 100       | |
| Mem         | ~720mb   | ~80mb     | | ~320mb   | ~80mb     | |
| Time        | 23.384s  | 13.411s   | |24.465s   | 13.670s   | |

**Comment by dima-kov** [comment]
Now, it can be asserted that the releasing of the Async version will result in a twofold acceleration of I/O endpoints.

**Comment by dima-kov** [comment]
Guys, we really need this (having async resolvers). Please tell us how can we help to make it public., I do not want to start using it unsure this is merged in main.

Or at least let us know your are going to release this, but we are missing: 1,2,3...

**Comment by superlevure** [comment]
Thanks for the benchmarks, I'll review the PR again tomorrow. Note that I don't have merge rights on this repo, we'll need a review from one of @firaskafri, @sjdemartini, @kiendang (or others)

**Comment by firaskafri** [comment]
@superlevure i think its is good to go as soon as we clarify the docs
What do you think @kiendang @jaw9c @sjdemartini

**Comment by superlevure** [comment]
@dima-kov I had a look at your benchmarks and I have few remarks. First, it looks like you're comparing

**Comment by superlevure** [comment]
@dima-kov I had a look at your benchmarks and I have few remarks. 

First, it looks like you're comparing the `sync` and `async` resolvers on the same branch of `graphene-django` (this PR's branch). I believe it would actually be more fair to compare the `async` / `sync` versions of this branch and the `sync` version of `graphene-django`'s `main` branch since this PR also affects `sync` resolvers and the point of the benchmarks is also to make sure no performance penalty is introduced to already existing code. 

Second, I noticed the `sync` resolver is returning [500 objects](https://github.com/dima-kov/django-graphene-benchmarks/blob/main/project/api/schema/queries.py#L51C13-L51C25) while the `async` version is only returning [10 objects](https://github.com/dima-kov/django-graphene-benchmarks/blob/main/project/api/schema_async/queries.py#L54).

I took the liberty of [pushing a PR](https://github.com/dima-kov/django-graphene-benchmarks/pull/1) to your repo that covers those point, as well as setting up a `docker` based environment and `postgres` as a DB to be closer to a real life use case. 

I obtain the following results:

```
# Sync version [main]
Concurrency Level:      100
Time taken for tests:   45.517 seconds
Complete requests:      1000
Failed requests:        0
Non-2xx responses:      1000
Total transferred:      60232000 bytes
Total body sent:        238000
HTML transferred:       59963000 bytes
Requests per second:    21.97 [#/sec] (mean)
Time per request:       4551.739 [ms] (mean)
Time per request:       45.517 [ms] (mean, across all concurrent requests)

# Sync version [this branch]
Concurrency Level:      100
Time taken for tests:   203.300 seconds
Complete requests:      1000
Failed requests:        0
Total transferred:      45805000 bytes
Total body sent:        244000
HTML transferred:       45382000 bytes
Requests per second:    4.92 [#/sec] (mean)
Time per request:       20330.019 [ms] (mean)
Time per request:       203.300 [ms] (mean, across all concurrent requests)

# Async version [this branch]
# didn't run till the end, see below
```

Data tends to show a huge perf hit for the `sync` version of this PR. Concerning the `async` version, I run into the following DB error which which makes all requests fail until I restart the DB: 

```
2024-02-11 16:31:39.310 UTC [4579] FATAL:  sorry, too many clients already
```

I suspect the `async` version is leaving DB connections opened somewhere (Postgres `max_connections` settings is set at `100` which is the concurrency level used in the benchmark)

I'll continue to play with the PR a bit tonight to make sure there's nothing wrong with my setup, but I'm curious if others can reproduce similar results.

dima-kov:
@superlevure thank you for looking into it! I missed that part

dima-kov:
regarding the results you've got: hmm, thats really strange. Both sync and async (100 requests concurrently and 1k requests in total) showed me +- 20s. 100+s is smth unbelievable for me.

Maybe the thing is with dockerization. I'm playing with main version now

virinchi123 (approved):

dolgidmi:
> 
> Data tends to show a huge perf hit for the `sync` version of this PR. Concerning the `async` version, I run into the following DB error which which makes all requests fail until I restart the DB:
> 
> ```
> 2024-02-11 16:31:39.310 UTC [4579] FATAL:  sorry, too many clients already
> ```
> 
> I suspect the `async` version is leaving DB connections opened somewhere (Postgres `max_connections` settings is set at `100` which is the concurrency level used in the benchmark)
> 
Django currently doesn't support persistent connections for ASGI https://code.djangoproject.com/ticket/33497

rw88:
> > Data tends to show a huge perf hit for the `sync` version of this PR. Concerning the `async` version, I run into the following DB error which which makes all requests fail until I restart the DB:
> > ```
> > 2024-02-11 16:31:39.310 UTC [4579] FATAL:  sorry, too many clients already
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I suspect the `async` version is leaving DB connections opened somewhere (Postgres `max_connections` settings is set at `100` which is the concurrency level used in the benchmark)
> 
> Django currently doesn't support persistent connections for ASGI https://code.djangoproject.com/ticket/33497

For those who use any external connection pooler like pgbouncer or AWS RDS Connection Pooling, this will not be a problem.

henadzit:
Thank you for all your work! How can we help to get it merged?

dima-kov:
> > > Data tends to show a huge perf hit for the `sync` version of this PR. Concerning the `async` version, I run into the following DB error which which makes all requests fail until I restart the DB:
> > > ```
> > > 2024-02-11 16:31:39.310 UTC [4579] FATAL:  sorry, too many clients already
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I suspect the `async` version is leaving DB connections opened somewhere (Postgres `max_connections` settings is set at `100` which is the concurrency level used in the benchmark)
> > 
> > 
> > Django currently doesn't support persistent connections for ASGI https://code.djangoproject.com/ticket/33497
> 
> For those who use any external connection pooler like pgbouncer or AWS RDS Connection Pooling, this will not be a problem.

https://docs.djangoproject.com/en/5.1/releases/5.1/#postgresql-connection-pools

django 5.1 will support connection pool and this might be an opportunity to merge this

pcraciunoiu:
I was getting an error that the view is not async on the latest Django release

Based on [this](https://github.com/django/django/blob/0e49a8c3bd9119795525d9f076f73740741479b7/django/views/generic/base.py#L63-L64), there must be an async method or the property returns false. As a quick fix, I pushed this classproperty to be true:

https://github.com/UpliftAgency/graphene-django/commit/cc665dfe2b45d7006971c66ddaebc5f7ae0c77fc

Feel free to add that in.

jaw9c:
Apologies for the radio silence on this thread. Appreciate the feedback and efforts from everyone on this PR - in particular the benchmarking. 

A fair few months ago, we rolled this into production at my work, utilising the new `AsyncGraphQLView` but found that we very quickly maxed out the DB connection pool. Since then we had to put the efforts on ice in lieu of making sure the startup was default alive... We've been using this branch in production since my last post, server under ASGI, but not using the AsyncGraphQLView.

I think this is safe to merge, but with an experimental warning for those willing to test and use the `AsyncGraphQLView`. (Perhaps we rename it to experimental_AsyncGraphQLView` or log a warning? Just to make sure people understand the DB connections point!

I've been carefully watching the movements by the Django team on making the ORM fully async, [this PR](https://github.com/django/django/pull/18408) is the latest, which is really encouraging!

henadzit:
> django 5.1 will support connection pool and this might be an opportunity to merge the PR

Looks like [Django 5.1](https://docs.djangoproject.com/en/dev/releases/5.1/) got released!

chris-stetter:
Are there plans to publish a tagged release? Happy to support if anything is missing.

superlevure:
> [..]
> I think this is safe to merge, but with an experimental warning for those willing to test and use the `AsyncGraphQLView`. (Perhaps we rename it to experimental_AsyncGraphQLView` or log a warning? Just to make sure people understand the DB connections point!
> 
> [..]

@jaw9c, do you think it's safe to merge even though my benchmarks showed this PR degrades performance for `sync` resolvers?

hunasdf:
Is there any update of this?

androane:
Maybe I'm missing something, but what's the point of graphene version 3 as long as it is simply not working?! or at least data loaders are not working. Not even with Django 5.1

ulgens:
@androane graphene-django project died long ago, but I wasn't able to give an EOL status to it when I was a maintainer. (then somebody revoked my rights without communicating first) This is not a maintained project, it's just a couple of people at the time didn't want to archive it. Occasional merges of a couple of small things here and there is not even remotely enough to keep an API-layer library modern.

I helped multiple startups to remove graphene from their stack, or replace with [Strawberry](https://strawberry.rocks/). Trying to keep this tool alive was a lot of work even 5 years ago and today, it's generally "Oh we made a mistake, please save us!" level thing. It's amazing to see how much time and energy is wasted just because there is no clear "NOT MAINTAINED" warning is not in the place.

androane:
@ulgens Thanks so much for your answer, that explains so much and confirms my hunches as well...

ulgens (changes requested):
Merging something of this importance into a dead project will only confuse people, and (considering the project's history and lack of latest activities) I don't think that people will be around to improve or fix it when it will be needed.

Keeping the PR open gives an impression that someday this will be a working part of graphene, so I guess the least I can do is to reject it.

ethagnawl:
That's _extremely_ useful context, @ulgens. Thanks.

dima-kov:
this is really bullshit. We chosen graphene-django long time ago when I was starting a project, somedays May-June 2023. We were considering graphene, and just seeing this PR, I thought nice, in several month it will be production ready.

Now need to move to alternative, and rewrite whole codebase

syrusakbary:
I've been pinged about this thread two times in the last week.
A bit of context, I created both the GraphQL-Python community and the Graphene and Graphene Django projects, although I've been a bit disconnected from it lately due to my duties and work at @wasmerio.

> graphene-django project died long ago, but I wasn't able to give an EOL status to it when I was a maintainer
@ulgens all maintainers need to be aligned for deprecating the project, which was not the case. It's true that Graphene-Django has not been active as of lately, but it's still alive and will making lots of progress moving forward.

I'll be taking on ownership of graphene-django back to add support of the learnings on using it with dataloaders over the last years in @wasmerio (we are using it heavily). Stay tuned

syrusakbary:
Also, @jaw9c thanks for your hard work on this PR. I have no idea why no one on the maintainers team decided to merge your PR, but I'll be reviewing it soon and hopefully merge.

As part of the team of maintainers of Graphene Django, my sincere apologies for taking so much time to review it.

dima-kov:
@syrusakbary thank you!

Graphene-Django still falls back to sync behavior in key execution paths. Given posted async load tests, do you see a feasible roadmap for full async compatibility? Do we need to wait full async compatibility from django core (if this will be merged: https://github.com/django/django/pull/18408)? 

Happy to contribute if there’s a focused plan—what do you see as the major blockers?

jaw9c:
Glad to have you back @syrusakbary! Thanks for your kind words - I think the main blocker became that the performance didn't increase due to the reasons outlined by @dima-kov above. The majority of people will be hitting the ORM during their queries, and running async with results in more (slightly more) computation from nested use of `sync_to_async`. There is still a net benefit if a workload is non ORM heavy - but otherwise we're waiting for https://github.com/django/django/pull/18408 to land AFAIK for the major benefits to be realised

syrusakbary:
I took a deeper look on the PR.
Few things:
* Middleware in GraphQL is very costly (it can slow down things heavily), so we should avoid it as much as possible (`DjangoSyncRequiredMiddleware`). We can simply wrap some of the resolvers with `@sync_to_async` if needed
* Data fetching could be done
* I'm working on [`graphql-server`](https://github.com/graphql-python/graphql-server) which adds support for tons of frameworks with Sync and Async views. Graphene-Django will be using this moving forward (as it adds compatibility with websockets/channels and more)

syrusakbary:
> The majority of people will be hitting the ORM during their queries, and running async with results in more (slightly more) computation from nested use of sync_to_async

Using async with Django Channels / ASGI improves things a tiiny bit (since they use the [`database_sync_to_async`](https://channels.readthedocs.io/en/stable/topics/databases.html#database-sync-to-async))

For non-asgi views performance will be on par (if no middleware is added).
I'd like to plan next steps better and then get back to you on this PR :)

syrusakbary:
> I'm working on [graphql-server](https://github.com/graphql-python/graphql-server) which adds support for tons of frameworks with Sync and Async views. Graphene-Django will be using this moving forward (as it adds compatibility with websockets/channels and more)

Small update: `graphql-server` is now published with the `3.0.0` version. Supporting a variety of web frameworks: Django, FastAPI, aiohttp, starlette and many more (with both sync requests, async and also websockets). We will now need to upgrade `graphene-django` to use it under the hood

PR #1398, opened 2023-04-19
fix: use execution_context_class attribute for GraphQLView

Arfey:
Simple change for using execution_context_class as class attribute

```python
class MyGraphQLView(GraphQLView):
    execution_context_class = CustomExecutionContext
```


graphene_django/views.py
```diff
@@ -112,7 +112,7 @@ def __init__(
         self.pretty = self.pretty or pretty
         self.graphiql = self.graphiql or graphiql
         self.batch = self.batch or batch
-        self.execution_context_class = execution_context_class
+        self.execution_context_class = execution_context_class or self.execution_context_class
```
jaw9c, line 115: We probs want to follow the same pattern as the other attributes (`graphiql`, `batch` etc):

```suggestion
        self.execution_context_class = self.execution_context_class or execution_context_class
```

jaw9c (changes requested):

graphene_django/views.py
Arfey: I think so this is not a good pattern. class attribute it's something like default value. when we create an instance we need to have approach to redefine it. thats why i use current order.

jaw9c: @Arfey Nice, thats a good point - could you update the other attributes to follow that pattern instead?

Arfey: done

firaskafri:
@Arfey great PR!
Seems we have some testing errors?

jaw9c (approved):
Looks like you just need to run the pre-commit hook so that the formatter runs, nice one on the PR

PR #1404, opened 2023-05-02
fix: fix graphiql request failure

### LINKED ISSUE #1402: GraphiQL v2.4.1 won't run IntrospectionQuery properly

GraphiQL update to v2.4.1 won't load properly. Loading up the page gives this error 
```json
{
  "errors": [
    {
      "message": "Must provide query string."
    }
  ]
}
```

Investigating the Network tab  shows the IntrospectionQuery is giving a 400 error. I've tested the query against my app using Postman and it works correctly.

Queries against the `/graphq/` endpoint using Postman also work.

Just the update to the new interface for GraphiQL seems to break.

* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via
a github repo, https://repl.it or similar (you can use this template as a starting point: https://repl.it/@jkimbo/Graphene-Django-Example).

[Example repo here](https://github.com/fullrobot/grapqhl_example)

1. Clone repo
2. pip install requirements
3. navigate to `localhost:8000/graphql/` to see error.

GraphiQL view works properly.

  - Version: graphene-django==3.0.1
  - Platform: MacOS

> NOTE: GraphiQL works fine on graphene-django==3.0.0

kiendang:
Fix #1402

After upgrading GraphiQL to v2.4.1 (#1396) all requests seem to fail on Chrome and Safari but works on Firefox. Unfortunately while working on that PR I tested exclusively on Firefox and didn't catch this.

The problem seems to be caused by this line.
https://github.com/graphql-python/graphene-django/blob/34cc86063b4d01212d30c6244ae58cde78b6a139/graphene_django/static/graphene_django/graphiql.js#L78

Also took this opportunity to simplify the GraphQL fetcher logic. `Accept` and `Content-Type` are automatically taken care of and don't need setting explicitly.

A more detailed explanation of the bug: `content-type: application/json` is automatically added and so we have both `content-type` and `Content-Type` in the `headers` object which somehow got turned into `content-type: application/json, application/json` which is invalid.

@sparktx-adam-gleason @dicknetherlands could you help test this with

```sh
pip install git+https://github.com/graphql-python/graphene-django.git@graphiql-headers-fix
```

### DIFF
--- a/graphene_django/static/graphene_django/graphiql.js
+++ b/graphene_django/static/graphene_django/graphiql.js
@@ -60,39 +60,26 @@
 
   function trueLambda() { return true; };
 
-  var fetcher = GraphiQL.createFetcher({
+  var headers = {};
+  var cookies = ("; " + document.cookie).split("; csrftoken=");
+  if (cookies.length == 2) {
+    csrftoken = cookies.pop().split(";").shift();
+  } else {
+    csrftoken = document.querySelector("[name=csrfmiddlewaretoken]").value;
+  }
+  if (csrftoken) {
+    headers['X-CSRFToken'] = csrftoken
+  }
+
+  var graphQLFetcher = GraphiQL.createFetcher({
     url: fetchURL,
     wsClient: graphqlWs.createClient({
       url: subscribeURL,
       shouldRetry: trueLambda,
       lazy: true,
-    })
+    }),
+    headers: headers
   })
-
-  function graphQLFetcher(graphQLParams, opts) {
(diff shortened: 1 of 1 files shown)


dicknetherlands:
This fix works for me (Chrome on Mac).

sparktx-adam-gleason:
This fix also works for me (Brave on Mac)

jaw9c (approved):
Looks good, nice cleanup too 👍

firaskafri (approved):

PR #1405, opened 2023-05-03
Bump version

firaskafri:
This release will address the issue caused by https://github.com/graphql-python/graphene-django/pull/1315

Just a heads up for folks who use get_queryset as an auth layer. If you need that, you can stick with versions between >=3.0.0b9 and <3.1.0 for now. Feel free to suggest a different way that won't lead to those pesky N+1 SQL query issues.

### DIFF
--- a/graphene_django/__init__.py
+++ b/graphene_django/__init__.py
@@ -1,7 +1,7 @@
 from .fields import DjangoConnectionField, DjangoListField
 from .types import DjangoObjectType
 
-__version__ = "3.0.1"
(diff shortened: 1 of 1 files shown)


jaw9c (approved):

kiendang:
We should include #1404 in this release too. v3.0.1 broke GraphiQL for browsers other than Firefox

kiendang:
Oops commented too late

firaskafri:
@kiendang no worries

PR #1415, opened 2023-05-27
Set pypi GH action version to release/v1

kiendang:
So it always uses the latest v1 release. We don't have to manually bump the minor and patch version. That's what the [doc](https://github.com/pypa/gh-action-pypi-publish/blob/unstable/v1/README.md) recommends anyway.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -20,7 +20,7 @@ jobs:
         pip install wheel
         python setup.py sdist bdist_wheel
     - name: Publish a Python distribution to PyPI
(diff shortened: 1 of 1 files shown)


ulgens:
I'd say it's better to keep it pinned, but activate automatic update checks.

kiendang:
Not a big issue so I'm fine either way and I also don't think there's much of a difference between 2 approaches, but also this is how we set the versions of other actions as well, `actions/checkout@v3` and `actions/setup-python@v4`.

ulgens (approved):

firaskafri (approved):

PR #1397, opened 2023-04-17
Add GraphiQL Explorer plugin

kiendang:
Close #1204.

Based on [`graphiql-plugin-explorer` CDN example](https://github.com/graphql/graphiql/blob/4879984ea1803a6e9f97d81c97e8ba27aacddae9/packages/graphiql-plugin-explorer/examples/index.html) and [`strawberry` `graphiql` template](https://github.com/strawberry-graphql/strawberry/blob/1b439b4aca3c0aba917126b8139a9d6b573806df/strawberry/static/graphiql.html).

PR #1418, opened 2023-06-01
Fix graphiql explorer styles by including official CSS, and update local example app for testing

sjdemartini:
Resolves https://github.com/graphql-python/graphene-django/issues/1417

(For the graphiql explorer styles recently added in https://github.com/graphql-python/graphene-django/pull/1397)

Followed the official graphiql-plugin-explorer example here
https://github.com/graphql/graphiql/blob/619864691941c46cc0b0848e8713028e20212c36/packages/graphiql-plugin-explorer/examples/index.html#L26-L29

Note the scrollbar that shows up in the "after" image below. Tested this locally with the cookbook-plain app, and was able to scroll to see the full content even on a short window. Also tested this with a larger app with many query definitions, and it worked there as well.

| Before | After |
|--------|--------|
|  Screenshot 2023-06-01 at 10 33 38 AM | Screenshot 2023-06-01 at 10 33 50 AM |


examples/cookbook-plain/cookbook/settings.py
```diff
@@ -5,10 +5,10 @@
 Generated by 'django-admin startproject' using Django 1.9.
 
 For more information on this file, see
-https://docs.djangoproject.com/en/1.9/topics/settings/
```
sjdemartini, line 8: The 1.9 links are dead, so I replaced them all with 3.2

kiendang (approved):
Awesome! I read that example in the graphiql repo too but didn't notice the css. Was a bit border by the styling issue but just thought it's just that way haha.

sjdemartini:
Yeah I also just thought that's the way it looked haha, didn't realize there was a separate CSS file until seeing that issue filed! Thanks for the review!

kiendang:
I took the chance to update GraphiQL and graphql-ws version as well. Note that graphql-ws hash does not change (coz changes from 5.12.1 -> 5.13.1 are server and type related, so I guess that's why the CDN file didn't change).

firaskafri (approved):

Xaelias:
👋 
Thanks for the fix. Any idea when this will make it to a public release?

firaskafri:
> 👋 Thanks for the fix. Any idea when this will make it to a public release?

its available in 3.1.2 now.

mnasiri:
the problem exists in the other place

PR #1419, opened 2023-06-05
Add check for `serializers.HiddenField` on `fields_for_serializer` function

sezginacer:
This pull request solves the issue that causes hidden fields to appear in Input when both `SerializerMutation` and serializer class with `HiddenField` are used. This field class is usually used when you want to auto-populate a field, not filled by request's data. So, this field should not be appear in Input.

### DIFF
--- a/graphene_django/rest_framework/mutation.py
+++ b/graphene_django/rest_framework/mutation.py
@@ -39,7 +39,9 @@ def fields_for_serializer(
                 field.read_only
                 and is_input
                 and lookup_field != name,  # don't show read_only fields in Input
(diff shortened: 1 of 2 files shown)


sjdemartini (approved):
Looks good to me! Agreed this should be the behavior based on [DRF docs on `HiddenField`](https://www.django-rest-framework.org/api-guide/fields/#hiddenfield) saying

> A field class that does not take a value based on user input, but instead takes its value from a default value or callable.

Thanks for adding the test!

firaskafri (approved):

PR #1421, opened 2023-06-07
remove JSONFIeld compat

dulmandakh:
Django 3.1 introduced universal JSONFIeld so compatibility is no longer required because we support Django 3.2 or above.

### DIFF
--- a/graphene_django/compat.py
+++ b/graphene_django/compat.py
@@ -10,16 +10,9 @@ try:
         IntegerRangeField,
         ArrayField,
         HStoreField,
-        JSONField as PGJSONField,
         RangeField,
     )
 except ImportError:
-    IntegerRangeField, ArrayField, HStoreField, PGJSONField, RangeField = (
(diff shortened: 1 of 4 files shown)


dulmandakh:
@ulgens fixed 😉

graphene_django/compat.py
```diff
@@ -10,16 +10,7 @@ def __init__(self, *args, **kwargs):
...
-    IntegerRangeField, ArrayField, HStoreField, PGJSONField, RangeField = (
-        MissingType,
-    ) * 5
-
-try:
-    # JSONField is only available from Django 3.1
-    from django.db.models import JSONField
-except ImportError:
-    JSONField = MissingType
+    IntegerRangeField, ArrayField, HStoreField, RangeField = (MissingType,) * 5
```
sjdemartini, line 16: Looks like this should be `* 4` now instead

firaskafri (approved):

sjdemartini (commented):
One quick comment, but thanks for the cleanup!

Issue #1428, opened 2023-07-08
[Critical] Breaking change in patch release


Natureshadow:
In version 3.1.2, the `JSONField` in `graphene_django.compat` was removed.

While removing this is a valid change, it constitutes a breaking change, and therefore a bump of the project's major version.

Due to semantic versioning, downstream applications will automatically update to `3.1.2` and break if they imported the field from there.

Please make a critical bugfix release reverting the change and release it to PyPI as version 3.1.3, if possible, today ☺!

PR #1429, opened 2023-07-08
Add JSONField back to graphene_django.compat module for backwards compatibility

sjdemartini:
Fixes https://github.com/graphql-python/graphene-django/issues/1428 (after the change merged in https://github.com/graphql-python/graphene-django/pull/1421)

This should improve backwards compatibility, fixing issues in downstream packages (notably graphene-django-cud
https://github.com/tOgg1/graphene-django-cud/issues/109, and also graphene-django-extras, both of which depended on
`graphene_django.compat.JSONField`). I've tested this in my downstream project and confirmed it resolved issues in both of those packages.

### DIFF
--- a/graphene_django/compat.py
+++ b/graphene_django/compat.py
@@ -1,3 +1,9 @@
+# For backwards compatibility, we import JSONField to have it available for import via
+# this compat module (https://github.com/graphql-python/graphene-django/issues/1428).
+# Django's JSONField is available in Django 3.2+ (the minimum version we support)
+from django.db.models import JSONField
+
+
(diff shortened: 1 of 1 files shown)


kiendang (approved):

sjdemartini:
@firaskafri any chance you could take a look and see about merging this? Just a one-line change to fix backwards incompatibility. This is blocking my team from upgrading past 3.1.1 (and seemingly others like https://github.com/graphql-python/graphene-django/issues/1428). Thanks!

firaskafri (approved):

Issue #1385, opened 2023-01-23
SerializerMutation not resolving Enums from serializer with ChoiceField


allen-munsch:
Why would this be the case?

I am using the latest branch.

```python
class SomeCreateSerializerMutation(SerializerMutation)
    ID = graphene.ID()

    class Meta:
        serializer_class = SomeListViewSerializer

    @classmethod 
    def mutate_and_get_payload(cls, root, info, **input):
        # kwargs = cls.get_serializer_kwargs(root, info, **input)
        kwargs = patched_get_serializer_kwargs(cls, root, info, **input)
```

```python
class SomeListViewSerializer(...):
    choice_type = serializers.ChoiceField(
           choices=choice.CHOICE_TYPE_CHOICES, allow_blank=True, required=False
    )
```

where i have to:

```python
def patched_get_serializer_kwargs(cls, root, info, **input):
    '''
    This is a patched version of:
        graphene_django.rest_framework.mutation.SerializerMutation.get_serializer_kwargs

    ############################
    # field.run_validation(primitive_value)
    # the serializer mutation type is not converting the `Choice_Type`,
    # after it is passed to the serializer kwargs
    # (Pdb) kwargs
    # {'instance': None, 'data': {'choice_type': <choice_type.TURN: 'TURN'>,
    # 'name': '', 'description': ''},
    # 'context': {'request': <WSGIRequest: POST '/graphql'>}, 'partial': False}
    # (Pdb) c
    # -> if serializer.is_valid():
    # (Pdb) serializer.is_valid()
    # False
    # (Pdb) serializer.errors
    # {'choice_type': [ErrorDetail(string='"choice_type.TURN" is not a valid choice.', code='invalid_choice')]}
    # input['choice_type'] = input['choice_type'].value
    #########################
    '''
    lookup_field = cls._meta.lookup_field
    model_class = cls._meta.model_class
    if model_class:
        ############ PATCH #################
        # i'm not sure why
        for k, v in input.items():
            attr = getattr(cls.Input, k)
            type_ = getattr(attr, 'type')
            if hasattr(v, 'value') and type_.get(v.value):
                input[k] = v.value
        ############ END PATCH #################

        if "update" in cls._meta.model_operations and lookup_field in input:
            instance = get_object_or_404(
                model_class, **{lookup_field: input[lookup_field]}
            )
            partial = True
        elif "create" in cls._meta.model_operations:
            instance = None
            partial = False
        else:
            raise Exception(
[...]

PR #1431, opened 2023-07-14
Update mutation.py to serialize Enum objects into input values

allen-munsch:
https://github.com/graphql-python/graphene-django/issues/1385

### DIFF
--- a/graphene_django/rest_framework/mutation.py
+++ b/graphene_django/rest_framework/mutation.py
@@ -1,3 +1,5 @@
+from enum import Enum
+
 from collections import OrderedDict
 
 from django.shortcuts import get_object_or_404
@@ -124,8 +126,11 @@ class SerializerMutation(ClientIDMutation):
     def get_serializer_kwargs(cls, root, info, **input):
         lookup_field = cls._meta.lookup_field
         model_class = cls._meta.model_class
-
         if model_class:
+            for input_dict_key, maybe_enum in input.items():
+                if isinstance(maybe_enum, Enum):
+                    choicefield_enum = maybe_enum
+                    input[input_dict_key] = choicefield_enum.value
             if "update" in cls._meta.model_operations and lookup_field in input:
                 instance = get_object_or_404(
                     model_class, **{lookup_field: input[lookup_field]}
--- a/graphene_django/rest_framework/mutation.py
+++ b/graphene_django/rest_framework/mutation.py
@@ -129,8 +129,7 @@ class SerializerMutation(ClientIDMutation):
         if model_class:
             for input_dict_key, maybe_enum in input.items():
                 if isinstance(maybe_enum, Enum):
-                    choicefield_enum = maybe_enum
-                    input[input_dict_key] = choicefield_enum.value
+                    input[input_dict_key] = maybe_enum.value
             if "update" in cls._meta.model_operations and lookup_field in input:
                 instance = get_object_or_404(
                     model_class, **{lookup_field: input[lookup_field]}
--- a/graphene_django/rest_framework/tests/test_mutation.py
+++ b/graphene_django/rest_framework/tests/test_mutation.py
@@ -7,7 +7,12 @@ from graphene import Field, ResolveInfo
 from graphene.types.inputobjecttype import InputObjectType
 
 from ...types import DjangoObjectType
-from ..models import MyFakeModel, MyFakeModelWithDate, MyFakeModelWithPassword
+from ..models import (
+    MyFakeModel,
+    MyFakeModelWithDate,
+    MyFakeModelWithPassword,
+    MyFakeModelWithChoiceField,
+)
 from ..mutation import SerializerMutation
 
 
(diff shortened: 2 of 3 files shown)


sjdemartini:
Thanks @allen-munsch! Would you be able to add a unit test for this? It would be great to ensure we handle this example and prevent any future regressions.

(Also, please run pre-commit formatting on the file per the failed CI job.)

allen-munsch:
@sjdemartini I added a unit test, and ran the precommit hook. Looked like it was all green locally.

Thank you for spending the time to review this. 

Much appreciated!

firaskafri (approved):

graphene_django/rest_framework/tests/test_mutation.py
```diff
@@ -268,6 +273,34 @@ class Meta:
     assert result.days_since_last_edit == 4
 
 
+def test_perform_mutate_success_with_enum_choice_field():
+    class ListViewChoiceFieldSerializer(serializers.ModelSerializer):
+        choice_type = serializers.ChoiceField(
+            choices=MyFakeModelWithChoiceField.ChoiceType, required=False
```
sjdemartini, line 279: Based on the issue reported https://github.com/graphql-python/graphene-django/issues/1385 and the DRF docs herehttps://www.django-rest-framework.org/api-guide/fields/#choicefield, shouldn't this be `choices=MyFakeModelWithChoiceField.ChoiceType.choices` instead? (fwiw, the tests passed for me either way when I ran it)

graphene_django/rest_framework/mutation.py
```diff
@@ -124,8 +126,13 @@ def __init_subclass_with_meta__(
     def get_serializer_kwargs(cls, root, info, **input):
         lookup_field = cls._meta.lookup_field
         model_class = cls._meta.model_class
-
         if model_class:
+            # check if input is Enum, and serialize value into input
+            for k, v in input.items():
+                attr = getattr(cls.Input, k)
+                type_ = getattr(attr, 'type')
```
sjdemartini, line 133: There's still a pre-commit black formatting issue here which is failing CI. `'type'` -> `"type"`

graphene_django/rest_framework/tests/test_mutation.py
```diff
@@ -268,6 +273,34 @@ class Meta:
     assert result.days_since_last_edit == 4
 
 
+def test_perform_mutate_success_with_enum_choice_field():
```
sjdemartini, line 276: I ran this test locally without the new changes in `mutation.py`, and it still passes. 🤔  Do you know why that would be? Does it relate to the version of django-rest-framework perhaps? Is the bug in https://github.com/graphql-python/graphene-django/issues/1385 actually still reproducible/present?

I think it would be worth ensuring that this test fails on master (presumably in the way that https://github.com/graphql-python/graphene-django/issues/1385 described) so that we're confident the fix resolves the original problem. Or if the change in `mutation.py` is no longer needed, we should remove that but still merge your test.

graphene_django/rest_framework/mutation.py
```diff
@@ -124,8 +126,13 @@ def __init_subclass_with_meta__(
...
+                if isinstance(v, Enum) and type_.get(v.value):
+                    input[k] = v.value
```
sjdemartini, line 135: Why isn't this new code just the following:

```python
            for k, v in input.items():
                if isinstance(v, Enum):
                    input[k] = v.value
```

i.e., what's the purpose of `type_.get(v.value)`? 

Related: would it be possible to use more verbose/explicit variable names so that it's clearer what this code means / is doing?

sjdemartini (commented):
Thank you for updating with a test! Really appreciate the contributions here 😄

graphene_django/rest_framework/tests/test_mutation.py
allen-munsch: @sjdemartini apologies, its been a minute since I worked with this code, and got the original error. 

I was able to repro with your DRF tip, where i got this on `main` with the :

```
choices=[(x.name, x.value) for x in MyFakeModelWithChoiceField.ChoiceType]
```

```
FAILED graphene_django/rest_framework/tests/test_mutation.py::test_perform_mutate_success_with_enum_choice_field - assert [ErrorType(field='choiceType', messages=[ErrorDetail(string='"asdf" is not a valid choice.', code='invalid_choice')])] is None
```

allen-munsch: @sjdemartini apologies, its been a minute since I worked with this code, and got the original error. 

I was able to repro with your DRF tip, where i got this on `main` with the :

```
choices=[(x.name, x.value) for x in MyFakeModelWithChoiceField.ChoiceType]
```

- https://github.com/graphql-python/graphene-django/pull/1431/commits/b1494da6be8adedbe3fef9871a7ae958e845d40d

```
FAILED graphene_django/rest_framework/tests/test_mutation.py::test_perform_mutate_success_with_enum_choice_field - assert [ErrorType(field='choiceType', messages=[ErrorDetail(string='"asdf" is not a valid choice.', code='invalid_choice')])] is None
```

graphene_django/rest_framework/tests/test_mutation.py
allen-munsch: @sjdemartini It is still present. I was able to repro with your DRF docs call out..

See: https://github.com/graphql-python/graphene-django/pull/1431#discussion_r1272697792

I wasn't instantiating the `choices` kwarg list correctly.

graphene_django/rest_framework/mutation.py
allen-munsch: You're right Thank you!

graphene_django/rest_framework/mutation.py
allen-munsch: Updated.

graphene_django/rest_framework/mutation.py
allen-munsch: I was a bit confused on where to use the `type_.get` I think i was supposed to pass it in for the test, I now have this, which also repros the issue on master.


In the test, now:

```
    kwargs = SomeCreateSerializerMutation.get_serializer_kwargs(
        None,
        mock_info(),
        **{
            "choice_type": SomeCreateSerializerMutation.Input.choice_type.type.get(
                "ASDF"
            )
        }
    )

``` 

Previous:

```
    kwargs = SomeCreateSerializerMutation.get_serializer_kwargs(
        None, mock_info(), **{"choice_type": MyFakeModelWithChoiceField.ChoiceType.ASDF}
    )

```

allen-munsch: updated the naming, and structure a bit to be more verbose and explicit in the code

I also removed the comment, I feel like now the code/naming/variables shows the intent more clearly, so it felt redundant to have the comment.

graphene_django/rest_framework/mutation.py
```diff
@@ -124,8 +126,11 @@ def __init_subclass_with_meta__(
     def get_serializer_kwargs(cls, root, info, **input):
         lookup_field = cls._meta.lookup_field
         model_class = cls._meta.model_class
-
         if model_class:
+            for input_dict_key, maybe_enum in input.items():
+                if isinstance(maybe_enum, Enum):
+                    choicefield_enum = maybe_enum
+                    input[input_dict_key] = choicefield_enum.value
```
sjdemartini, line 133: super minor: I think this could remove the intermediate variable to simplify, since it should be clear

```suggestion
                    input[input_dict_key] = maybe_enum.value
```

sjdemartini (approved):
Thanks so much for the quick followup and in investigating that question. That's great you were able to make that test surface the original bug on master! And I like how simple the fix ended up being! LGTM!

allen-munsch:
@sjdemartini No problem. It really means a lot you were working through the backlog of issues. So I wanted to make sure I didn't let this one sit around. Cheers!

PR #1440, opened 2023-08-03
Remove Python 3.7 (EOL since EOL since 2023-06-27) from CI

kiendang:
https://devguide.python.org/versions/

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -29,4 +29,3 @@ jobs:
       run: tox
       env:
         DJANGO: ${{ matrix.django }}
-        TOXENV: ${{ matrix.toxenv }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,8 +11,6 @@ jobs:
         django: ["3.2", "4.1", "4.2"]
         python-version: ["3.8", "3.9", "3.10"]
         include:
-          - django: "3.2"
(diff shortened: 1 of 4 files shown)


sjdemartini (approved):
Cool, LGTM!

One comment: we may want to change the `pre-commit` [pyupgrade](https://github.com/asottile/pyupgrade) arg to use `--py38-plus` now instead of py37: https://github.com/graphql-python/graphene-django/blob/8540a9332cda7bcbac92643cacbb5146f35080e0/.pre-commit-config.yaml#L22

That could presumably be a separate PR after this, since we'll likely want to re-run it on all existing files to upgrade any syntax that was kept around in support of 3.7.

kiendang:
Good catch! We could also take this opportunity to switch to [ruff](https://github.com/astral-sh/ruff)? Ruff implements both `pyupgrade` and `flake8` which we use in our pre-commit.

sjdemartini:
Yeah great suggestion. I'm using ruff instead of pyupgrade, flake8, etc in a project of mine, and I'm a big fan!

kiendang:
@sjdemartini I just include the change to `--py38-plus` here since after running pre-commit there's no change.
I was looking into changing to ruff but it's a bit more involved since we need to migrate the flake8 and isort config in `setup.cfg` so will leave to a future PR.

firaskafri (approved):

PR #1441, opened 2023-08-05
Use ruff in pre-commit

kiendang:


graphene_django/filter/utils.py
```diff
@@ -50,7 +51,7 @@ def get_filtering_args_from_filterset(filterset_class, type):
             ):
                 # Get the filter field for filters that are no explicitly declared.
                 if filter_type == "isnull":
-                    field = graphene.Boolean(required=required)
+                    graphene.Boolean(required=required)
```
kiendang, line 54: `field = ...` was removed here as it's not referenced later. Was this a typo and should have been `model_field = ...` instead?

.ruff.toml
```diff
@@ -0,0 +1,32 @@
+select = [
+    "E",    # pycodestyle
+    "W",    # pycodestyle
+    "F",    # pyflake
+    "I",    # isort
+    "B",    # flake8-bugbear
+    "UP",   # pyupgrade
```
sjdemartini, line 7: It's nice to also include `C4` in addition to `UP`, since `C4` handles "unnecessary comprehensions" which was formerly handled by pyupgrade https://beta.ruff.rs/docs/rules/#flake8-comprehensions-c4

.ruff.toml
```diff
@@ -0,0 +1,32 @@
+select = [
+    "E",    # pycodestyle
+    "W",    # pycodestyle
+    "F",    # pyflake
+    "I",    # isort
+    "B",    # flake8-bugbear
```
sjdemartini, line 6: Thanks for including comments for each of the rule categories!

.ruff.toml
```diff
@@ -0,0 +1,32 @@
...
+]
+
+target-version = "py38"
+
+[per-file-ignores]
+# Ignore unused imports (F401) in these files
+"__init__.py" = ["F401"]
+"graphene_django/compat.py" = ["F401"]
+
+[isort]
```
sjdemartini, line 28: It looks like this should be `[tool.ruff.isort]` since we're using the isort rules via ruff now (https://beta.ruff.rs/docs/settings/#isort), right? It seems ruff has support for each of these options.

graphene_django/filter/utils.py
sjdemartini: Oh dang, this shows the value of a good linter! Looking back at the PR that introduced this code https://github.com/graphql-python/graphene-django/pull/1119/files#diff-cdc00f7a891a74f7c16d107ad5346f75d56a657b9316c8b207283c7b507cecb6R52 and trying to understand what this code is doing, I think this line should actually be:

```python
                field_type = graphene.Boolean(required=required)
```

The function is trying to return a graphene type given a filter, and it's saying if the filter is "isnull", then we should be returning a `graphene.Boolean` field for that. That matches up with the rest of the `type`s of the various variables and logically makes sense to me.

graphene_django/forms/tests/test_mutation.py
```diff
@@ -280,7 +279,7 @@ def test_model_form_mutation_mutate_invalid_form():
     result = PetMutation.mutate_and_get_payload(None, None)
 
     # A pet was not created
-    Pet.objects.count() == 0
+    assert Pet.objects.count() == 0
```
sjdemartini, line 282: Good catch!

setup.py
```diff
@@ -26,10 +26,6 @@
 
 
 dev_requires = [
-    "black==23.3.0",
```
sjdemartini, line 29: Maybe worth including `ruff` here? I guess not strictly required but it probably makes it easier to use in folks' editors

sjdemartini (commented):
Sweet! Thanks for doing this! Overall this LGTM!

.ruff.toml
kiendang: Ah it's only `[tool.ruff.isort]` if we use `pyproject.toml`. if `ruff.yaml` or `.ruff.yaml` we can omit the `tool.ruff.`. I have checked this. It's working.

sjdemartini: Ahh gotcha, I hadn't even noticed this was `.ruff.toml` rathr than `pyproject.toml` 👍

.ruff.toml
kiendang: Added. Thanks! Didn't know about that.

.ruff.toml
kiendang: I saw this in some other repo and found it such a great idea as well.

sjdemartini (approved):

PR #1442, opened 2023-08-05
Prevent duplicate CI runs

kiendang:
We have this in our workflows

```yaml
on: [push, pull_request]
```

This is redundant since when someone pushes to a PR branch, it triggers both `push` and `pull_request` `synchronized` which makes the action runs twice. Checking every PR CI and you will see we have the every check run twice `(push)` and `(pull_request)`.

Just `on: push` is enough.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,6 +1,6 @@
 name: Lint
 
-on: [push, pull_request]
+on: push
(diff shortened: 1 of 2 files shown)


sjdemartini (approved):
Good call!

kiendang:
...

PR #1446, opened 2023-08-09
fix: foreign key nullable and custom resolver

superlevure:
Fix `model.DoesNotExist` exception when trying to resolve a null one-to-one relationship with a custom `get_queryset` implemented on the `DjangoObjectType` (see test)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,7 +10,7 @@ dev-setup:
 
 .PHONY: tests ## Run unit tests
 tests:
-	py.test graphene_django --cov=graphene_django -vv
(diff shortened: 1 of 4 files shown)


tcleonard (approved):
Looks good to me

superlevure:
Thank you for your review @firaskafri, it seems that both @tcleonard and I lost our merge permissions, would you be able to do it?

firaskafri:
> Thank you for your review @firaskafri, it seems that both @tcleonard and I lost our merge permissions, would you be able to do it?

Sure! I just checked and I believe you and @tcleonard still have the write permissions though.

PR #1447, opened 2023-08-09
Miscellaneous CI fixes

kiendang:
- Fix `Makefile` following the switch to ruff
- Some miscellaneous `tox.ini` fixes
- Updating `ruff` version in `pre-commit`. Previously running `ruff` directly yields different result compared to `pre-commit`. Updating `ruff` fixed that.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ tests:
 
 .PHONY: format ## Format code
 format:
-	tox -e pre-commit
+	black graphene_django examples setup.py
 
 .PHONY: lint ## Lint code
 lint:
--- a/Makefile
+++ b/Makefile
@@ -10,15 +10,15 @@ dev-setup:
 
 .PHONY: tests ## Run unit tests
 tests:
-	PYTHONPATH=. py.test graphene_django --cov=graphene_django -vv
+	PYTHONPATH=. pytest graphene_django --cov=graphene_django -vv
 
 .PHONY: format ## Format code
 format:
-	black graphene_django examples setup.py
+	tox -e pre-commit
 
 .PHONY: lint ## Lint code
 lint:
-	flake8 graphene_django examples
+	ruff graphene_django examples
 
 .PHONY: docs ## Generate docs
 docs: dev-setup
--- a/setup.py
+++ b/setup.py
@@ -26,8 +26,8 @@ tests_require = [
 
 
 dev_requires = [
-    "black",
-    "ruff",
+    "black==23.7.0",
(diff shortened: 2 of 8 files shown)


Makefile
```diff
@@ -10,15 +10,15 @@ dev-setup:
 
 .PHONY: tests ## Run unit tests
 tests:
-	py.test graphene_django --cov=graphene_django -vv
+	pytest graphene_django --cov=graphene_django -vv
 
 .PHONY: format ## Format code
 format:
-	black graphene_django examples setup.py
+	tox -e pre-commit
```
sjdemartini, line 17: I think this actually has a different effect, since it won't apply to all existing files and will just apply to uncommitted changes. Perhaps this should be `pre-commit run --all-files` instead of just `pre-commit`?

Makefile
```diff
@@ -10,15 +10,15 @@ dev-setup:
 
 .PHONY: tests ## Run unit tests
 tests:
-	py.test graphene_django --cov=graphene_django -vv
+	pytest graphene_django --cov=graphene_django -vv
 
 .PHONY: format ## Format code
 format:
```
sjdemartini, line 16: side suggestion here: perhaps we add back `black` in the `dev_requires` dependencies in `setup.py` (so `dev-setup` in this Makefile also installs black, since that will likely be useful for folks' editor configurations just like `ruff`)

sjdemartini (approved):
Overall looks good! A couple quick suggestions

sjdemartini:
And thanks again for making these improvements!

Makefile
kiendang: ah this runs the `pre-commit` tox env, not the pre-commit command. In `tox.ini` you will see https://github.com/graphql-python/graphene-django/blob/7bd4a684b243c9bd61d015ef4a4ce651a93bbc01/tox.ini#L37-L41

that's the command being run

but I agree on adding back black and reversing this change.

setup.py
```diff
@@ -26,6 +26,7 @@
 
 
 dev_requires = [
+    "black",
     "ruff",
```
sjdemartini, line 30: We should presumably pin these to the same versions as in pre-commit so we ensure the same formatting

Makefile
sjdemartini: Ah gotcha, okay 👍

firaskafri:
Great work @kiendang ! Is it ready to be merged?

kiendang:
Yup I think so. @sjdemartini? Thanks for the reviews btw!

PR #1450, opened 2023-08-10
fix: empty list is not an empty value for list filters even when a custom filtering method is provided

tcleonard:
…


graphene_django/filter/tests/test_typed_filter.py
```diff
@@ -87,14 +102,15 @@ def test_typed_filter_schema(schema):
...
+        "headlineSearch": "[String]",
     }
 
     all_articles_filters = (
         schema_str.split("  articles(")[1]
         .split("): ArticleTypeConnection\n")[0]
         .split(", ")
     )
-
+    print(all_articles_filters)
```
superlevure, line 113: to be removed?

sjdemartini (commented):
Good catch, and thanks for adding the tests!

graphene_django/filter/filters/list_filter.py
```diff
@@ -1,12 +1,32 @@
...
 class ListFilter(TypedFilter):
     """
     Filter that takes a list of value as input.
     It is for example used for `__in` filters.
     """
 
+    @TypedFilter.method.setter
+    def method(self, value):
+        """
+        Override method setter so that in case a custom `method` is provided, it doesn't fall back to
```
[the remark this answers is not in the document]
tcleonard: The array filter is relying on the `EMPTY_VALUES` check so that's why I didn't add the same implementation. 
I'll still put some tests on it to see if the behaviour of that filter makes sense

firaskafri:
@tcleonard long time! glad you have time to make great contributions as before! 
is this still a WIP?

graphene_django/filter/filters/list_filter.py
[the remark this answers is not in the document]
tcleonard: While adding tests I realized that none of the array field tests were getting run because we were not having postgres in the test environment, so I ended up revamping all of those tests to fake an `ArrayField` with a `JSONField` instead for unit tests only.

And you are correct that the same logic was needed in the `ArrayFilter`

tcleonard:
I just pushed an update taking into account your suggestion and revamping the tests. Let me know what you think

graphene_django/filter/tests/test_array_field_exact_filter.py
```diff
@@ -1,18 +1,14 @@
 import pytest
 
-from graphene import Schema
-
 from ...compat import ArrayField, MissingType
 
 
-@pytest.mark.skipif(ArrayField is MissingType, reason="ArrayField should exist")
-def test_array_field_exact_no_match(Query):
+# @pytest.mark.skipif(ArrayField is MissingType, reason="ArrayField should exist")
```
superlevure, line 6: leftover?

superlevure (changes requested):

graphene_django/compat.py
```diff
@@ -19,4 +22,16 @@ def __init__(self, *args, **kwargs):
         RangeField,
     )
 except ImportError:
-    IntegerRangeField, ArrayField, HStoreField, RangeField = (MissingType,) * 4
+    IntegerRangeField, HStoreField, RangeField = (MissingType,) * 3
+
+    # For unit tests we fake ArrayField using JSONFields
+    if os.path.basename(sys.argv[0]) in ["pytest", "py.test"]:
```
kiendang, line 28: nitpick: this would not work if someone calls `python -m pytest` since `sys.argv[0]` would be `../pytest/__main__.py`. This would be a bit more robust

```python
from pathlib import PurePath

if any(
    PurePath(sys.argv[0]).match(p)
    for p in [
        "**/pytest",
        "**/py.test",
        "**/pytest/__main__.py",
    ]
):
```

superlevure (approved):

graphene_django/filter/filters/array_filter.py
```diff
@@ -1,13 +1,36 @@
...
+    @TypedFilter.method.setter
+    def method(self, value):
+        """
+        Override method setter so that in case a custom `method` is provided
+        (see documentation https://django-filter.readthedocs.io/en/stable/ref/filters.html#method),
+        it doesn't fall back to checking if the value is in `EMPTY_VALUES` (from the `__call__` method
+        of the `FilterMethod` class) and instead use our ArrayFilterMethod that consider empty lists as values.
+
+        Indeed when providing a `method` the `filter` method bellow is overridden and replaced by `FilterMethod(self)`
+        which means that the validation of the empty value is made by the `FilterMethod.__call__` method instead.
```
sjdemartini, line 28: typo: bellow -> below

graphene_django/filter/filters/list_filter.py
```diff
@@ -1,12 +1,36 @@
...
 
+    @TypedFilter.method.setter
+    def method(self, value):
+        """
+        Override method setter so that in case a custom `method` is provided
+        (see documentation https://django-filter.readthedocs.io/en/stable/ref/filters.html#method),
+        it doesn't fall back to checking if the value is in `EMPTY_VALUES` (from the `__call__` method
+        of the `FilterMethod` class) and instead use our ListFilterMethod that consider empty lists as values.
+
+        Indeed when providing a `method` the `filter` method bellow is overridden and replaced by `FilterMethod(self)`
```
sjdemartini, line 27: typo: bellow -> below

sjdemartini (approved):
Thank you for handling the ArrayField as well and for adding more test coverage there too (and nice job getting a workaround for non-Postgres environments)!

tcleonard:
if anyone of you has merge permissions (I don't seem to have them anymore), could you merge this please? 🙏

PR #1452, opened 2023-08-11
Only release on pypi after tests pass

kiendang:
Context https://github.com/graphql-python/graphene-django/issues/1448#issuecomment-1672195085

Relevant GHA docs [`workflow_call`](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_call), [`needs`](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds)

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -6,8 +6,13 @@ on:
       - 'v*'
 
 jobs:
-  build:
+  lint:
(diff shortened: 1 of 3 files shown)


sjdemartini (approved):
Looks great, thanks so much for adding this! This is the same approach/syntax I used in a frontend package I maintain and it works well there

PR #1454, opened 2023-08-22
Remove default output fields from SerializerMutation

nossila:
I found a little annoying that when we use a SerializerMutation all fields in the Serializer, used mostly for the Input, also gets added in the Output. 

The best reason to use a SerializerMutation is for the input validations that DRF brings to the table, the Output is very easy to set and most of the times we want to return our own ObjectType, not all fields used as the input.

What you guys think?

### DIFF
--- a/graphene_django/rest_framework/mutation.py
+++ b/graphene_django/rest_framework/mutation.py
@@ -99,14 +99,6 @@ class SerializerMutation(ClientIDMutation):
             convert_choices_to_enum=convert_choices_to_enum,
             lookup_field=lookup_field,
         )
-        output_fields = fields_for_serializer(
-            serializer,
-            only_fields,
-            exclude_fields,
-            is_input=False,
-            convert_choices_to_enum=convert_choices_to_enum,
-            lookup_field=lookup_field,
-        )
 
         if not _meta:
             _meta = SerializerMutationOptions(cls)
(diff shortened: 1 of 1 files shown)


sjdemartini:
I don't personally use any of the DRF Serializer functionality within graphene-django, so I'm not very familiar with the reasons for or against this, but it's seemingly a breaking change to how things have worked to date and would probably need to be "opt-out" rather than changed universally. Tests are breaking with the change as is (e.g. see [here](https://github.com/graphql-python/graphene-django/actions/runs/5933590750/job/16090880862?pr=1454#step:5:564)). 

Could you show a more complete example of a serializer, the resulting schema, what fields you do/don't want, and why the existing graphene-django logic here is problematic?

(For what it's worth, I've found https://github.com/tOgg1/graphene-django-cud to be really well suited for handling write-operations as a more "GraphQL-first" approach, versus DRF Serializers where it all feels more shoehorned and a bit less clean.)

nossila:
@sjdemartini the biggest reason I was using the SeralizerMutation was because of valitations and for partial updates, I found the django form integration wasn't very good for PATCH but seems like [graphene-django-cud](https://github.com/tOgg1/graphene-django-cud) really covers everything, thanks a lot for showing that.

I'm closing tis PR.

PR #1457, opened 2023-08-26
Reduce cognitive complexity and typo fixes

lilac-supernova-2:
Hello there!

Just wanna say I love this project so I went ahead and did some small contributions.
Pretty much just splitting out some functions that were too big, and fixed some typos.

While doing this I got paranoid with unit tests so I added a `tests-repeat` command to the `Makefile` that will allow you to run all unit tests, 100 times, in parallel - this can be useful in the future to catch bugs and/or flaky tests that can fail from time to time (which, after implementing, got some occasions where the tests failed but worked nicely after re-running - so maybe there are either undiscovered bugs and/or some flaky tests whose configuration - could be due to using some randomness on some tests - fail, well, randomly)

Tested and formatted the code as per the contributing guidelines. Let me know if you have any questions and if you're ok with this contribution!


graphene_django/views.py
```diff
@@ -103,16 +103,15 @@ def __init__(
     ):
         if not schema:
             schema = graphene_settings.SCHEMA
+        self.schema = schema or self.schema
 
         if middleware is None:
             middleware = graphene_settings.MIDDLEWARE
+        if isinstance(middleware, MiddlewareManager):
```
lilac-supernova-2, line 110: Let me know if I should've kept the check for `middleware is not None` here - my thought process was that we already check above if it's none, so probably we don't need to check if now it is _not_ none.

Unless `graphene_settings.MIDDLEWARE` can be `None` - then this is needed and I can re-add it back in.

graphene_django/utils/testing.py
```diff
@@ -128,7 +128,7 @@ def _client(self, client):
         )
         self.client = client
 
-    def assertResponseNoErrors(self, resp, msg=None):
+    def assert_response_no_errors(self, resp, msg=None):
```
sjdemartini, line 131: This is a breaking change for consumers of this package. I would recommend at least preserving an alias with the original camelCase name.  I imagine it was named this way for consistency with Django (and Python's) camelCase assertion methods: https://docs.djangoproject.com/en/4.2/topics/testing/tools/#assertions

sjdemartini (changes requested):
Thanks for making improvements throughout graphene-django! Would it be possible to break this up into a couple different PRs: one with just the typo/content fixes, and one with actual code changes? I haven't taken the time to go through meticulously and validate that the refactored code matches the original behavior, and personally less sure how necessary that is (I think complexity scores only tell you so much). The typo fixes and whatnot seem like an easy thing to merge first. (On the code changes, I just left a couple comments where there are breaking changes to method names, which I don't think we should merge as is. If there are other breaking changes to graphene-django method signatures we should avoid those as well.)

graphene_django/utils/testing.py
```diff
@@ -138,7 +138,7 @@ def assertResponseNoErrors(self, resp, msg=None):
         self.assertEqual(resp.status_code, 200, msg or content)
         self.assertNotIn("errors", list(content.keys()), msg or content)
 
-    def assertResponseHasErrors(self, resp, msg=None):
+    def assert_response_has_errors(self, resp, msg=None):
```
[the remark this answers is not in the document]
lilac-supernova-2: @sjdemartini thanks for the replies and for your _very_ important suggestions - I'll take care of these soon enough.

PR #1458, opened 2023-09-01
Support displaying deprecated input fields in GraphiQL docs

romainletendart:
GraphiQL supports displaying deprecated input fields through an opt-in configuration property.
This PR exposes a new GRAPHENE setting that is forwarded all the way through the aforementioned configuration property.

### DIFF
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -197,9 +197,6 @@ Set to ``False`` if you want to disable GraphiQL headers editor tab for some rea
 
 This setting is passed to ``headerEditorEnabled`` GraphiQL options, for details refer to GraphiQLDocs_.
 
-.. _GraphiQLDocs: https://github.com/graphql/graphiql/tree/main/packages/graphiql#options
-
-
 Default: ``True``
 
 .. code:: python
@@ -230,8 +227,6 @@ Set to ``True`` if you want to persist GraphiQL headers after refreshing the pag
 
 This setting is passed to ``shouldPersistHeaders`` GraphiQL options, for details refer to GraphiQLDocs_.
 
-.. _GraphiQLDocs: https://github.com/graphql/graphiql/tree/main/packages/graphiql#options
-
 
 Default: ``False``
 
@@ -240,3 +235,6 @@ Default: ``False``
    GRAPHENE = {
       'GRAPHIQL_SHOULD_PERSIST_HEADERS': False,
    }
+
+
+.. _GraphiQLDocs: https://graphiql-test.netlify.app/typedoc/modules/graphiql_react#graphiqlprovider-2
--- a/docs/settings.rst
+++ b/docs/settings.rst
@@ -237,4 +237,35 @@ Default: ``False``
    }
 
 
+``GRAPHIQL_INPUT_VALUE_DEPRECATION``
+------------------------------------
+
+Set to ``True`` if you want GraphiQL to show any deprecated fields on input object types' docs.
+
+For example, having this schema:
+
+.. code:: python
+
+    class MyMutationInputType(graphene.InputObjectType):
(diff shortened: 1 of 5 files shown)


docs/settings.rst
```diff
@@ -240,3 +235,23 @@ Default: ``False``
...
    }
+
+
+``GRAPHIQL_INPUT_VALUE_DEPRECATION``
+------------------------------------
+
+Set to ``True`` if you want GraphiQL to display a deprecation section on field doc view.
+
+This setting is passed to ``inputValueDeprecation`` GraphiQL options, for details refer to GraphiQLDocs_.
+Make sure your GraphQL server supports deprecation of input values before setting this to ``True``.
```
sjdemartini, line 246: I'm confused by this comment. Is this referring to the version of `graphql-core` (https://github.com/graphql-python/graphql-core), or `graphene`, or something else? In what situation would it not be supported?

docs/settings.rst
```diff
@@ -240,3 +235,23 @@ Default: ``False``
    GRAPHENE = {
       'GRAPHIQL_SHOULD_PERSIST_HEADERS': False,
    }
+
+
+``GRAPHIQL_INPUT_VALUE_DEPRECATION``
+------------------------------------
+
+Set to ``True`` if you want GraphiQL to display a deprecation section on field doc view.
```
sjdemartini, line 243: I _think_ I know what this means, but I'm not actually positive. I think it would be worth giving an example of a graphene/graphene-django schema definition for when this would come into play and what the effect of setting it to `True` is.

sjdemartini (commented):
Neat, thanks for adding this and updating the docs! 

Do you happen to have a screenshot or something illustrating the effect of having this set to `False` (the default, and presumably the behavior today) and `True`? The docs from graphiql don't provide much info about `inputValueDeprecation`.

docs/settings.rst
romainletendart: This sentence is actually wrong. I meant checking `graphql-js` version.

Where we could have a risk is when we would use a version of GraphiQL that would depend on an old version of `graphql` (i.e. `graphql-js`) that does not support input field deprecation, see: https://github.com/graphql/graphiql/blob/57b5fcd8749bd1c54f4a6a04a9ab07316fd05e71/packages/graphiql/resources/renderExample.js#L103

Only `graphql>15.5` supports input field deprecation.

After checking, it seems anyway that graphene-django pins `graphiql`'s version here: https://github.com/graphql-python/graphene-django/blob/main/graphene_django/views.py#L68 to `2.4.7` and this version of `graphiql` in turn uses `graphql@16.5.0`: https://github.com/graphql/graphiql/blob/graphiql%402.4.7/yarn.lock#L9991 which is above `15.5`.

So I think I can just remove this warning as it should no longer be a problem.

docs/settings.rst
romainletendart: I've added an example and some more details on what the setting actually does.

docs/settings.rst
```diff
@@ -240,3 +235,37 @@ Default: ``False``
...
+
+.. code:: python
+
+    class MyMutationInputType(graphene.InputObjectType):
+        old_field = graphene.String(deprecation_reason="You should now use 'newField' instead.")
+        new_field = graphene.String()
+
+    class MyMutation(graphene.Mutation):
+        class Arguments:
+            input = types.MyMutationInputType()
```
sjdemartini, line 255: minor: `types.MyMutationInputType()` -> `MyMutationInputType()` since the class is defined immediately above in this example and not in a `types` module

sjdemartini (approved):
Thank you for updating the docs, super helpful! I tested this end-to-end and it worked well for me. Nice addition! Just one super minor comment

PR #1460, opened 2023-09-13
Fix graphiql explorer styles by sending missing param to render_graphiql

mnasiri:
Resolves #1418 
Fix graphiql explorer styles by sending graphiql_plugin_explorer_css_sri param to render_graphiql function of the GraphQlView 

The problem:
The param `graphiql_plugin_explorer_css_sri` is used in `graphene_django/templates/graphene/graphiql.html`  ([this snippet](https://github.com/graphql-python/graphene-django/blob/67def2e074c3c03f0fedff2416efe950bb44aefa/graphene_django/templates/graphene/graphiql.html#L24-L27)) but doesn't set when send params to render graphiql [here](https://github.com/graphql-python/graphene-django/blob/67def2e074c3c03f0fedff2416efe950bb44aefa/graphene_django/views.py#L154-L175).

I had this problem locally.
so, I only add a line to the view.py

Thanks for your attempts.


sjdemartini (approved):
Good catch, thanks!

PR #1465, opened 2023-09-18
Fix test Client headers for Django 4.2

### LINKED ISSUE #1464: GraphQLTestCase does not send headers using Django 4.2
**Note: for support questions, please use stackoverflow**. This repository's issues are reserved for feature requests and bug reports.

Trying to send headers using GraphQLTestCase doesn't work. This is my test:

```
class TestCase(GraphQLTestCase):
    def test(self):
        response = self.query(
            '''
            query {
                puzzles {
                    id
                    title
                }
            }
            ''',
            headers={
                'Some-Header': 'test,
            },
        )

        self.assertResponseNoErrors(valid_response)
```

and if I try to log those headers in my code I get `None`. However if I follow the code to `graphene_django/utils/testing.py` and change line 58 from:

```
resp = client.post(
            graphql_url, json.dumps(body), content_type="application/json", **headers
        )
```

to 

```
resp = client.post(
            graphql_url, json.dumps(body), content_type="application/json", headers=headers
        )
```

Then it works fine.

  - Version: 
 
python = "^3.11"
django = "^4.2.1"
graphene-django = "^3.1.1"
 
  - Platform: Ubuntu

kiendang:
Fix #1464

### DIFF
--- a/graphene_django/utils/utils.py
+++ b/graphene_django/utils/utils.py
@@ -1,6 +1,5 @@
 import inspect
 
-import pkg_resources
 from django.db import connection, models, transaction
 from django.db.models.manager import Manager
 from django.utils.encoding import force_str
@@ -149,11 +148,15 @@ def bypass_get_queryset(resolver):
 
 
 def __django_version():
-    return pkg_resources.get_distribution("django").parsed_version
+    from pkg_resources import get_distribution
+
+    return get_distribution("django").parsed_version
 
 
 def __parse_version(v):
-    return pkg_resources.parse_version(v)
+    from pkg_resources import parse_version
+
+    return parse_version(v)
 
 
 _DJANGO_VERSION_AT_LEAST_4_2 = __django_version() >= __parse_version("4.2")
--- a/graphene_django/utils/utils.py
+++ b/graphene_django/utils/utils.py
@@ -1,5 +1,6 @@
 import inspect
 
+import pkg_resources
 from django.db import connection, models, transaction
 from django.db.models.manager import Manager
 from django.utils.encoding import force_str
@@ -145,3 +146,14 @@ def bypass_get_queryset(resolver):
     """
     resolver._bypass_get_queryset = True
     return resolver
+
+
+def __django_version():
+    return pkg_resources.get_distribution("django").parsed_version
+
+
+def __parse_version(v):
+    return pkg_resources.parse_version(v)
+
+
+_DJANGO_VERSION_AT_LEAST_4_2 = __django_version() >= __parse_version("4.2")
--- a/graphene_django/utils/utils.py
+++ b/graphene_django/utils/utils.py
@@ -1,5 +1,6 @@
 import inspect
 
+import django
 from django.db import connection, models, transaction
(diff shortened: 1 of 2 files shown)


firaskafri (approved):

graphene_django/utils/utils.py
```diff
@@ -145,3 +145,18 @@ def bypass_get_queryset(resolver):
     """
     resolver._bypass_get_queryset = True
     return resolver
+
+
+def __django_version():
+    from pkg_resources import get_distribution
```
sjdemartini, line 151: There's a warning at the top here https://setuptools.pypa.io/en/latest/pkg_resources.html saying

> Use of pkg_resources is deprecated in favor of [importlib.resources](https://docs.python.org/3/library/importlib.resources.html#module-importlib.resources), [importlib.metadata](https://docs.python.org/3/library/importlib.metadata.html#module-importlib.metadata) and their backports ([importlib_resources](https://pypi.org/project/importlib_resources), [importlib_metadata](https://pypi.org/project/importlib_metadata)). Some useful APIs are also provided by [packaging](https://pypi.org/project/packaging) (e.g. requirements and version parsing). Users should refrain from new usage of pkg_resources and should work to port to importlib-based solutions.

Instead of `parse_version()`, it looks like we could use https://packaging.pypa.io/en/stable/version.html (`from packaging.version import parse`). And for getting the currently-installed django version string, it looks like `import django; django.get_version()` would work (or alternatively [`importlib.metadata.version`](https://docs.python.org/3/library/importlib.metadata.html#module-importlib.metadata) seems to be the substitute for `get_distribution("...").parsed_version`).

Unsure how important this deprecation warning is, but seems like it may be worthwhile to avoid the new code depending on this, as they suggest.

---

Another alternative could be something like this (if we want to avoid `packaging` for instance):

```python
import django

_DJANGO_VERSION_AT_LEAST_4_2 = django.VERSION[0] > 4 or (django.VERSION[0] >= 4 and django.VERSION[1] >= 2)
``` 
```

**Review by sjdemartini — COMMENTED** [review]
Thanks for quickly adding this @kiendang! Looks good. just one point for quick discussion

File: graphene_django/utils/utils.py (hunk 1 above)
Comment (superlevure) [inline graphene_django/utils/utils.py:151]: The usage of `pkg_resources` is deprecated: https://setuptools.pypa.io/en/latest/pkg_resources.html

How about using:

```python
from django import VERSION
print(VERSION)
# (4, 1, 11, 'final', 0)
```

instead?

superlevure: Looks like @sjdemartini beat me to it

kiendang: @sjdemartini @superlevure Thanks for the review. I know about `pkg_resources` being deprecated in favor of `importlib...` modules. However went for `pkg_resources.get_distribution("...").parsed_version` here since we need to compare semver while `importlib.metadata.version` only returns the version string, unparsed.

`django.VERSION` looks like exactly what we need which I was not aware of. Will convert to that. Thanks for pointing that out!

sjdemartini (approved):

Issue #1472, opened 2023-10-24
Query Limiting Depth


AstRonin:
**Is your feature request related to a problem? Please describe.**
Graphene has query validation about Depth limit Validator (https://docs.graphene-python.org/en/latest/execution/queryvalidation/#depth-limit-validator). 
```
validation_errors = validate(
    schema=schema.graphql_schema,
    document_ast=parse('THE QUERY'),
    rules=(
        depth_limit_validator(
            max_depth=20
        ),
    )
)
```

Would be great we will have it (depth) in django project.

**Describe the solution you'd like**
Limit depth of query. OWASP suggest add limit for security reason https://cheatsheetseries.owasp.org/cheatsheets/GraphQL_Cheat_Sheet.html#query-limiting-depth-amount

AstRonin:
Can this 
https://github.com/graphql-python/graphene-django/issues/1342#issuecomment-1230943989
example usable also for my question?

kiendang:
@AstRonin I believe that would work.

@firaskafri @sjdemartini @ulgens This use case has been requested a couple of time. If we go the route of porting over graphql-core [graphql_impl](https://github.com/graphql-python/graphql-core/blob/0c93b8452eed38d4f800c7e71cf6f3f3758cd1c6/src/graphql/graphql.py#L155) to improve performance (see https://github.com/graphql-python/graphene-django/pull/1439#discussion_r1283484438 and https://github.com/graphql-python/graphene-django/pull/1393#issuecomment-1474258711), this would be quite easy to add `validation_rules` as another option to `GraphQLView`. I could submit a PR for this, but currently I'm not using it so defer to you on it's worth implementing or not.

sjdemartini:
@kiendang I like the idea of adding validation rules as an option to GraphQLView. I didn't realize that was already a feature of graphene (and just not graphene-django). Sounds like it's worthwhile given requests like this and the ability to support both graphene's built-in validation like depth-limits, as well as potentially other user-defined validators.

sjdemartini:
@kiendang now that you've merged in the performance improvements and refactoring around graphene-django's query execution (thank you!), are you still planning on adding an option for validation rules in the `GraphQLView`? No rush, just curious. I've recently been thinking more about validation in a GraphQL app I work on, and this seems like it'd be a nice addition to graphene-django to enable some of that.

kiendang:
Yup I just put up a PR in #1475. Implementation is simple but I haven't added any test though.

PR #1393, opened 2023-03-14
Remove redundant call to validate

shukryzablah:
The call to `validate` in the django view is redundant with the validation call in graphql-core. Related to #1198

Issue #1342, opened 2022-08-29
How to disable introspection via validation rules?


smac89:
**Is your feature request related to a problem? Please describe.**
I need to disable introspection for graphql, how would I do that using the method described [here](https://docs.graphene-python.org/en/latest/execution/queryvalidation/#disable-introspection)

**Describe the solution you'd like**
An example of accomplishing this would be nice

**Describe alternatives you've considered**
I am thinking to either create a subclass of `GraphQLView` or `graphene.Schema` and add the query validation there, but I don't know if that's the recommended approach

**Additional context**

smac89:
I ended up going with this:
```py
from graphql import ExecutionResult, parse, validate
from graphql.validation import NoSchemaIntrospectionCustomRule

class ValidatingSchema(graphene.Schema):
    def __init__(self, *args, validation_rules=(), **kwargs):
        super().__init__(*args, **kwargs)
        self.validation_rules = validation_rules

    def execute(self, *args, **kwargs):
        return self.validate(*args, **kwargs) or super().execute(*args, **kwargs)

    async def execute_async(self, *args, **kwargs):
        return self.validate(*args, **kwargs) or await super().execute_async(*args, **kwargs)

    def validate(self, *args, **kwargs):
        if query := (kwargs.get("source") or kwargs.get("request_string")):
            errors = validate(self.graphql_schema, parse(query), rules=self.validation_rules, max_errors=3)
            if errors:
                return ExecutionResult(errors=errors)


schema = ValidatingSchema(
    query=...,
    mutation=...,
    directives=...,
    validation_rules=(
        *filter(None, (NoSchemaIntrospectionCustomRule if not settings.DEBUG else None,)),
    ),
)
```

PR #1474, opened 2023-10-26
Optimize execute_graphql_request

kiendang:
This ported over graphql-core [graphql_impl](https://github.com/graphql-python/graphql-core/blob/0c93b8452eed38d4f800c7e71cf6f3f3758cd1c6/src/graphql/graphql.py#L155) to speed up graphql execution by getting rid of duplicated steps. See https://github.com/graphql-python/graphene-django/pull/1393#issuecomment-1474258711 for context. This also paves the way for implementing support for validations rules (#1342, #1472)

This PR is basically just #1439 with some minor changes
- Added the missing schema validation step in [graphql_impl](https://github.com/graphql-python/graphql-core/blob/0c93b8452eed38d4f800c7e71cf6f3f3758cd1c6/src/graphql/graphql.py#L170)
- Keep the `show_graphiql` arg. Agree with @danthewildcat that this arg is somewhat redundant but removing it is a non-backward compatible change.

### DIFF
--- a/graphene_django/views.py
+++ b/graphene_django/views.py
@@ -12,6 +12,7 @@ from django.views.generic import View
 from graphql import ExecutionResult, OperationType, execute, get_operation_ast, parse
 from graphql.error import GraphQLError
 from graphql.execution.middleware import MiddlewareManager
+from graphql.language import OperationDefinitionNode
 from graphql.validation import validate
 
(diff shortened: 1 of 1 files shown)


kiendang:
Actually I'll just push my commits into #1439

PR #1477, opened 2023-11-15
Provide setting to enable/disable converting choices to enums globally

### LINKED ISSUE #1480: Provide setting to enable/disable converting choices to enums globally
In our project we want to disable the auto-convertion of Django choice fields to Enums for every model.
To specify convert_choices_to_enum = False on every Type is a bit cumbersome so a global setting would help a lot.

I've created a PR which showcases the improvement I have in mind: #1477

Flauschbaellchen:
In our project we want to disable the auto-convertion of Django choice fields to Enums for every model.
To specify `convert_choices_to_enum = False` on every `Type` is a bit cumbersome so a global setting would help a lot.

Any feedback is highly appreciated.

Closes #1480

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -133,13 +133,17 @@ def convert_choice_field_to_enum(field, name=None):
 
 
 def convert_django_field_with_choices(
-    field, registry=None, convert_choices_to_enum=True
+    field, registry=None, convert_choices_to_enum=None
 ):
     if registry is not None:
         converted = registry.get_converted_field(field)
         if converted:
             return converted
     choices = getattr(field, "choices", None)
+    if convert_choices_to_enum is None:
+        convert_choices_to_enum = bool(
+            graphene_settings.DJANGO_CHOICE_FIELD_ENUM_CONVERT
+        )
     if choices and convert_choices_to_enum:
         EnumCls = convert_choice_field_to_enum(field)
         required = not (field.blank or field.null)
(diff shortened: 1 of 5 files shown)


graphene_django/converter.py
```diff
@@ -133,13 +133,17 @@ def convert_choice_field_to_enum(field, name=None):
...
 def convert_django_field_with_choices(
-    field, registry=None, convert_choices_to_enum=True
+    field, registry=None, convert_choices_to_enum=None
 ):
     if registry is not None:
         converted = registry.get_converted_field(field)
         if converted:
             return converted
     choices = getattr(field, "choices", None)
+    if convert_choices_to_enum is None:
```
Flauschbaellchen, line 143: A reason why I moved it into `convert_django_field_with_choices` and not implemented it on `construct_fields` or `DjangoObjectType` is, that `django-graphene-plus` uses `convert_django_field_with_choices` directly and would skip/ignore the setting or would need to manually look it up.
See: https://github.com/0soft/graphene-django-plus/blob/master/graphene_django_plus/input_types.py#L28

kiendang (commented):
Thanks @Flauschbaellchen for the PR. This sounds like a valid use case to me and the changes you proposed look backward-compatible too. I don't see a problem with merging this. Thoughts @sjdemartini @firaskafri?

sjdemartini (approved):
These changes look good to me! Thanks for adding the thorough tests and for suggesting this; the motivation makes sense to me. I support merging 👍

kiendang (approved):

PR #1481, opened 2023-12-04
Add Python 3.12 to CI

kiendang:
Django 4.2.8 which added support for Python 3.12 has just been released.
https://docs.djangoproject.com/en/4.2/faq/install/#what-python-version-can-i-use-with-django

Django 5.0.0 is out as well but Django REST is not fully compatible yet, failing just 1 of our tests. Will settle that in a future PR.


sjdemartini (approved):
Thanks, LGTM!

firaskafri (approved):

PR #1483, opened 2023-12-08
Remove Django 4.1 (EOL) and add Django 5.0 to CI

kiendang:
https://docs.djangoproject.com/en/5.0/faq/install/#what-python-version-can-i-use-with-django
https://www.djangoproject.com/download/#supported-versions


kiendang:
> Looks great! Nice to see 5.0 passed all tests without any changes 😄

Yup. Turned out that 1 failed test was with djangomain and we don't need to worry about that until future releases.

PR #1487, opened 2023-12-23
Make use of http.HTTPStatus for response status code checks

ulgens:
I got this idea from https://www.b-list.org/weblog/2023/dec/04/python-http-status-codes/ . I always liked [DRF's status](https://www.django-rest-framework.org/api-guide/status-codes/) and I think it makes the code easier to read. Python's `http` version of the same thing is not looking nice as DRF's in my opinion (`.HTTP_200_OK` vs `.OK`), but still an improvement over using directly the numeric codes.

### DIFF
--- a/graphene_django/tests/test_views.py
+++ b/graphene_django/tests/test_views.py
@@ -1,4 +1,5 @@
 import json
+from http import HTTPStatus
 from unittest.mock import patch
 
 import pytest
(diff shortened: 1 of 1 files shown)


kiendang (commented):
We do have DRF as test dependency so we can use their status here. I agree that DRF's statuses look nicer and more dev-friendly too since sometimes we need to reference GraphQL specs and they mostly reference the status code directly.

kiendang:
And if you want to use it in the main codebase we can alias Python's `http` statuses?

Issue #1225, opened 2021-07-16
Cannot reference Object defined later in file when using DjangoListField


liliserf:
When trying to reference an object that's defined later in the file...

This works:
```
class Foo(graphene.ObjectType):
    bars = graphene.List(graphene.NonNull(lambda: Bar))


class Bar(graphene.ObjectType):
    baz = graphene.String()
```

This does not work:
```
class Foo(graphene.ObjectType):
    bars = DjangoListField(lambda: Bar)


class Bar(graphene.ObjectType):
    baz = graphene.String()
```

When using DjangoListField it returns:
```
web_1       | Traceback (most recent call last):
web_1       |   File "/usr/local/lib/python3.9/threading.py", line 954, in _bootstrap_inner
web_1       |     self.run()
web_1       |   File "/usr/local/lib/python3.9/threading.py", line 892, in run
web_1       |     self._target(*self._args, **self._kwargs)
web_1       |   File "/usr/local/lib/python3.9/site-packages/django/utils/autoreload.py", line 54, in wrapper
web_1       |     fn(*args, **kwargs)
web_1       |   File "/usr/local/lib/python3.9/site-packages/django/core/management/commands/runserver.py", line 117, in inner_run
web_1       |     self.check(display_num_errors=True)
web_1       |   File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 387, in check
web_1       |     all_issues = self._run_checks(
web_1       |   File "/usr/local/lib/python3.9/site-packages/django/core/management/base.py", line 377, in _run_checks
web_1       |     return checks.run_checks(**kwargs)
web_1       |   File "/usr/local/lib/python3.9/site-packages/django/core/checks/registry.py", line 72, in run_checks
web_1       |     new_errors = check(app_configs=app_configs)
web_1       |   File "/usr/local/lib/python3.9/site-packages/django/core/checks/urls.py", line 13, in check_url_config
web_1       |     return check_resolver(resolver)
web_1       |   File "/usr/local/lib/python3.9/site-packages/django/core/checks/urls.py", line 23, in check_resolver
web_1       |     return check_method()
web_1       |   File "/usr/local/lib/python3.9/site-packages/django/urls/resolvers.py", line 399, in check
web_1       |     for pattern in self.url_patterns:
web_1       |   File "/usr/local/lib/python3.9/site-packages/django/utils/functional.py", line 80, in __get__
web_1       |     res = instance.__dict__[self.name] = self.func(instance)
[...]

PR #1492, opened 2024-01-18
fix: same type list

superlevure:
Fix for https://github.com/graphql-python/graphene-django/issues/1225

Allow to use a recursive type for `DjangoListField` 

(fix suggested by @tcleonard)

### DIFF
--- a/graphene_django/tests/test_fields.py
+++ b/graphene_django/tests/test_fields.py
@@ -22,8 +22,13 @@ class TestDjangoListField:
         class Query(ObjectType):
             something = DjangoListField(String)
 
-        with pytest.raises(TypeError):
+        with pytest.raises(TypeError) as excinfo:
             Schema(query=Query)
+
+        assert (
+            "Query fields cannot be resolved. DjangoListField only accepts DjangoObjectType types as underlying type"
+            in str(excinfo.value)
+        )
 
     def test_only_import_paths(self):
         list_field = DjangoListField("graphene_django.tests.schema.Human")
--- a/graphene_django/tests/test_fields.py
+++ b/graphene_django/tests/test_fields.py
@@ -12,17 +12,18 @@ from .models import (
     Article as ArticleModel,
     Film as FilmModel,
     FilmDetails as FilmDetailsModel,
+    Person as PersonModel,
     Reporter as ReporterModel,
 )
 
 
 class TestDjangoListField:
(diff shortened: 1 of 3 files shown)


graphene_django/tests/test_fields.py
```diff
@@ -12,17 +12,18 @@
...
 class TestDjangoListField:
     def test_only_django_object_types(self):
-        class TestType(ObjectType):
-            foo = String()
+        class Query(ObjectType):
+            something = DjangoListField(String)
 
-        with pytest.raises(AssertionError):
-            DjangoListField(TestType)
+        with pytest.raises(TypeError):
```
sjdemartini, line 25: It looks like in code, this failure condition is still caused by an `assert` statement. Just curious then why this is manifesting as a TypeError now.

sjdemartini (approved):
Looks good to me based on the tests, thanks for adding support for this!

graphene_django/tests/test_fields.py
superlevure: Good question! The `AssertionError` is now caught and re-raise by `graphql-core` here: https://github.com/graphql-python/graphql-core/blob/0c93b8452eed38d4f800c7e71cf6f3f3758cd1c6/src/graphql/type/definition.py#L734

sjdemartini: Nice, thanks! Maybe it's worth adding a final line to the test to validate that the assertion message (about only using DjangoObjectTypes) is present in the raised exception string, since it seems it will be and that would make for a more thorough test

superlevure: That's a good point, done in https://github.com/graphql-python/graphene-django/pull/1492/commits/cc45b0bcdbde8b8502c6da0927e853c10bc83830

kiendang (approved):
LGTM

On a side note I notice a lot of `property`s (if not all of them) in `DjangoListField` and `DjangoConnectionField` are things that don't change. We could consider using `functools.cached_property`, but that's for a future issue/PR.

superlevure:
Thanks for the review @kiendang @sjdemartini, I'll let one of you merge as I don't have the authorization

sjdemartini:
I'm also unable to merge ("Merging is blocked...base branch restricts merging to authorized users"), but looks good to go!

tcleonard (approved):

PR #1495, opened 2024-01-29
feat!: check Django model has a default ordering when used in a Relay Connection

tcleonard:
Since predictable ordering is a pre-requisite for a Relay Connection pagination to work, we should enforce it.
Indeed if you don't have a consistent ordering, since the pagination is based on an offset under the hood, when requesting page 2 you wouldn't have guaranteed that it's actually the same order that was used to define page 1.

### DIFF
--- a/examples/starwars/models.py
+++ b/examples/starwars/models.py
@@ -24,6 +24,9 @@ class Faction(models.Model):
 
 
 class Ship(models.Model):
+    class Meta:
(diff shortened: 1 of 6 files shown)


superlevure (approved):
Good catch! Thanks for the PR @tcleonard

firaskafri (approved):

PR #1496, opened 2024-02-01
Associate GlobalIDFilter with BigAutoField and SmallAutoField

iamthad:
`AutoField` is set up to use `GlobalIDFilter`, but `BigAutoField` and `SmallAutoField` are not.

This MR adds them to `GRAPHENE_FILTER_SET_OVERRIDES`, associating them with the filter.

It also removes some conditional logic around `SmallAutoField` that was introduced in #1212 because [`SmallAutoField` was introduced in Django 3.0](https://docs.djangoproject.com/en/stable/releases/3.0/#models) and this project requires a minimum Django version of 3.2.

### DIFF
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -183,17 +183,11 @@ def convert_field_to_string(field, registry=None):
     )
 
 
+@convert_django_field.register(models.SmallAutoField)
 @convert_django_field.register(models.BigAutoField)
 @convert_django_field.register(models.AutoField)
 def convert_field_to_id(field, registry=None):
     return ID(description=get_django_field_description(field), required=not field.null)
-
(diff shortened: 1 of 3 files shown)


pizzapanther:
I originally added the `BigAutoField` support. This was definitely missed so thanks!  LGTM

sjdemartini (approved):
Cool, thanks for updating this!

iamthad:
I realized today, that in the spirit of https://xkcd.com/1172/, this will be a breaking change for those who were relying on the old behavior. Their APIs will start requiring global IDs on ID filters instead of integers.

How should this be addressed? Maybe a note in the changelog about how to restore the old behavior would be sufficient?

PR #1497, opened 2024-02-07
Allows field's choices to be a callable

nossila:
Starting in Django 5 field's choices can also be a callable https://docs.djangoproject.com/en/5.0/ref/models/fields/#django.db.models.Field.choices


kiendang:
Thanks @nossila could you also add a test for this?

sjdemartini (approved):
👍  thanks!

PR #1509, opened 2024-03-24
Fix ReadTheDocs builds

### LINKED ISSUE #1504: Read the docs is broken

Read The Docs is not currently able build our documentation: https://readthedocs.org/projects/graphene-django/builds/

A user should see the latest documentation when visiting https://docs.graphene-python.org/projects/django/en/latest/installation/

For the documentation to reflect current behavior

Seems we need to update some configuration, it also seems to be building against the "master" branch which no longer exists. Maybe we should create that branch off main to see if we can get it to work.

PabloAlexis611:
Closes #1504

Here we just add the `.readthedocs.yaml` file at the root of the repo.
The basic configuration seems to work out-of-the-box.

I tested this on a fork of the ReadTheDocs build as well, which you can (for now) access here: https://graphene-django-pabloalexis611-fork.readthedocs.io/en/latest/

I will delete the RTD-fork once the PR gets merged.

Also, we may need to make sure that the `graphene-django` RTD project has `main` as it's "latest" branch setup correctly (which I don't have access to).

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -1,0 +1,18 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.io/en/stable/config-file/v2.html for details
+
+version: 2
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.12"
+
+# Build documentation in the "docs/" directory with Sphinx
+sphinx:
+  configuration: docs/conf.py
+
+# See https://docs.readthedocs.io/en/stable/guides/reproducible-builds.html
+python:
+   install:
+      - requirements: docs/requirements.txt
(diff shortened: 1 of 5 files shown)


PabloAlexis611:
Link to latest build: https://beta.readthedocs.org/projects/graphene-django-pabloalexis611-fork/builds/23850776/

PabloAlexis611:
Just fixed some references to the old `master` branch and switched them to `main`.

The `main` branch doesn't seem to be up yet in coveralls, so I didn't update those links for the coverage badge.

sjdemartini (approved):
I'm not super familiar with ReadTheDocs setup, but this LGTM. Thank you for adding this and for confirming that RTD is showing successful builds. I also don't have access to the RTD project for graphene-django, but seems like this would make sense to merge regardless.

kiendang (approved):

Issue #1516, opened 2024-04-09
Enforcing model default ordering (#1495) may have been overzealous


gsvr:
#1495 - released in [3.2.1](https://github.com/graphql-python/graphene-django/releases/tag/v3.2.1) forces every model in a project to have a default ordering specified.

The impact of this is not trivial:
1. Each project using this will have to set ordering defaults on every model that a DjangoConnectionField is being applied to (in only one of my projects this is over 100 changes)
2. Ordering is not a free operation (please see https://docs.djangoproject.com/en/5.0/ref/models/options/#ordering)

I do agree the problem raised in #1495 is a good catch, but I think a better approach to handling this could be:
1. If there is no default ordering specified, use the models primary key as the default for sorting, since every model should have one (https://docs.djangoproject.com/en/5.0/topics/db/models/#automatic-primary-key-fields) and/or
2. Warn the user of potential problems rather than enforcing model updates

Please feel free to disagree with me if I'm misunderstanding or overlooking something here, in the meantime I'm pinning to 3.2.0 😊

kiendang:
Tagging @tcleonard.
@gsvr I agree with this. Requiring declaring `ordering` does break backward compatibility and if we do require it it should not have been a patch release. I think solution 1 is best here, setting default `ordering` to pk if it's not declared. @tcleonard @gsvr do one of you mind submitting a PR for this?

gsvr:
@kiendang I looked at opening a PR, but with my limited knowledge of the codebase I couldn't easily understand what the correct place would be to apply the default ordering.

I'd love to contribute here but I think it would probably take me too long to fully understand all the nuances and class inheritances here before I can come up with something useful. If you could point me in the right direction I might be able to attempt a change, but it might be worth reverting the original change in the meantime.

pecheneff:
Totally agree with @gsvr .
We also have a large project with many models so using default ordering everywhere is not free. Using `3.2.0` is the only way for now to avoid issues.
Please avoid adding breaking changes like this to a patch versions. Thanks in advance.

PR #1518, opened 2024-04-11
Not require explicitly set `ordering` in `DjangoConnectionField`

kiendang:
I'm working on fixing #1516, probably by setting `ordering = ["pk"]` by default in `DjangoConnectionField` `class Meta`. 

However I don't find any reference of `ordering` in both this or Graphene's code base. Does anyone have any idea? Or is setting `ordering` actually not doing anything? @sjdemartini @tcleonard

### DIFF
--- a/examples/starwars/models.py
+++ b/examples/starwars/models.py
@@ -24,9 +24,6 @@ class Faction(models.Model):
 
 
 class Ship(models.Model):
-    class Meta:
-        ordering = ["pk"]
-
     name = models.CharField(max_length=50)
     faction = models.ForeignKey(Faction, on_delete=models.CASCADE, related_name="ships")
 
--- a/graphene_django/fields.py
+++ b/graphene_django/fields.py
@@ -101,19 +101,13 @@ class DjangoConnectionField(ConnectionField):
             non_null = True
         assert issubclass(
             _type, DjangoObjectType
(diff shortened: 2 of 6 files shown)


gsvr:
@kiendang It looks like you're getting stuck at the same place I did - there doesn't seem to be an easy way to define ordering in a Meta class somewhere. The only ordering method I could find in the docs was using an [OrderingFilter](https://docs.graphene-python.org/projects/django/en/latest/filtering/#ordering), but setting a default with that doesn't look easily viable. I would suggest opening another feature issue to implement default ordering so this PR can be wrapped up :)

kiendang:
@gsvr my problem is that I don't know if setting

```py
class Meta
    ordering = ["pk"]
```

actually does anything because I can't seem to find any code that references `ordering` in this or Graphene code base.

If that is the case, we should just revert #1518.

In case that `ordering` actually works, setting a default `ordering` is doable, but even in this case I'm fine with reverting #1518 first then implementing that later.

gsvr:
@kiendang Yeah, I also searched around, and I don't think setting ordering in class Meta actually does anything. I couldn't find any usage of that in the docs or the code.

sjdemartini (approved):
lgtm to revert, thanks for handling this!

firaskafri (approved):

PR #1525, opened 2024-06-11
Bump to 3.2.2

### LINKED ISSUE #1521: Enforcing default ordering is too invasive (Django model <model> has to have a default ordering to be used in a Connection.)
Since version 3.2.1 a default ordering is enforced for models.   
First of all, I want to say that I run into the problem which the PR #1495 tries to fix several times so the change is high appreciated.

Previously, I resolved this by always let the client to choose the ordering (as the pagination is, in my optinion, more a frontend-related problem and nothing the backend should care about).

Enforcing the ordering on the model level with `Meta.ordering` has (IMO) some disadvantages:

* The ORM would always order all queries independent of pagination is required/used or not. As the model is (obviously) used on other places than the GraphQL endpoints, this would reduce the overall performance of the app.
* Third-party apps/models would need to be adjusted, too (see below)

Additionally, the change feels more like a breaking change, as it requires the codebase to be adjusted.

For our own models, we had a base class which was adjusted in the following
```python
class MyBaseModel(models.Model):
    class Meta:
        abstract = True
        ordering = ["pk"]
```

But as we do not only use our own models for the API, we had the problem that for third-party models the same error occurs:

```
TypeError: Query fields cannot be resolved.
Django model contenttypes.ContentType has to have a default ordering to be used in a Connection.
```

As said before, the intention of the change itself is high appriciated.
However, I would like to have a more non-invasive approach.

Would it be possible to just "fall back" to an ordering of `pk` in case `Meta` does not specify it in the context of the `Connection`?
Idk if this is something the user should be warned about or if it should be even be a setting the developer can adjust if it is "enforced", "fall-back-to-pk" or "silenced".

  - Version: Django 4.2.7
  - Platform: Python 3.10.x

Flauschbaellchen:
@kiendang Version bump after reverting #1518 as requested in https://github.com/graphql-python/graphene-django/issues/1521#issuecomment-2153737586

Closes #1521

### DIFF
--- a/graphene_django/__init__.py
+++ b/graphene_django/__init__.py
@@ -2,7 +2,7 @@ from .fields import DjangoConnectionField, DjangoListField
 from .types import DjangoObjectType
 from .utils import bypass_get_queryset
 
-__version__ = "3.2.1"
+__version__ = "3.2.2"
 
 __all__ = [
     "__version__",


kiendang (approved):
Thanks @Flauschbaellchen!

sjdemartini (approved):
🚢

PR #1529, opened 2024-07-04
Bugfix: call `resolver` function in `DjangoConnectionField` as documented

PoByBolek:
This fixes a bug in that `DjangoConnectionField` doesn't call its `resolver` function despite it being documented in the [Graphene docs: "Resolvers outside the class"](https://docs.graphene-python.org/en/latest/types/objecttypes/#resolvers-outside-the-class).

```python
# currently, this will never be called :(
def resolve_ships(root, info):
    return [ShipModel()]

class Query(ObjectType):
    ships = DjangoConnectionField(Ship, resolver=resolve_ships)
```

I also fixed some warnings and running the tests now [treats warnings as errors](https://docs.python.org/3/using/cmdline.html#envvar-PYTHONWARNINGS). But I really only care about c127aec75b33617b8eb10c84fb0733654ce70bf3 here 😄

### DIFF
--- a/examples/starwars/schema.py
+++ b/examples/starwars/schema.py
@@ -1,5 +1,5 @@
 import graphene
-from graphene import Schema, relay, resolve_only_args
+from graphene import Schema, relay
 from graphene_django import DjangoConnectionField, DjangoObjectType
 
 from .data import create_ship, get_empire, get_faction, get_rebels, get_ship, get_ships
@@ -62,16 +62,13 @@ class Query(graphene.ObjectType):
     node = relay.Node.Field()
     ships = DjangoConnectionField(Ship, description="All the ships.")
 
-    @resolve_only_args
-    def resolve_ships(self):
+    def resolve_ships(self, info):
         return get_ships()
 
-    @resolve_only_args
-    def resolve_rebels(self):
+    def resolve_rebels(self, info):
         return get_rebels()
 
-    @resolve_only_args
-    def resolve_empire(self):
+    def resolve_empire(self, info):
         return get_empire()
 
 
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -190,17 +190,11 @@ def convert_field_to_string(field, registry=None):
     )
 
 
-@convert_django_field.register(models.BigAutoField)
 @convert_django_field.register(models.AutoField)
+@convert_django_field.register(models.BigAutoField)
+@convert_django_field.register(models.SmallAutoField)
 def convert_field_to_id(field, registry=None):
     return ID(description=get_django_field_description(field), required=not field.null)
-
-
-if hasattr(models, "SmallAutoField"):
-
-    @convert_django_field.register(models.SmallAutoField)
-    def convert_field_small_to_id(field, registry=None):
-        return convert_field_to_id(field, registry)
 
 
 @convert_django_field.register(models.UUIDField)
--- a/graphene_django/forms/tests/test_converter.py
+++ b/graphene_django/forms/tests/test_converter.py
@@ -1,4 +1,4 @@
-from django import forms
+from django import VERSION as DJANGO_VERSION, forms
 from pytest import raises
 
 from graphene import (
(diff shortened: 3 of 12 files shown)


graphene_django/converter.py
```diff
@@ -190,19 +190,13 @@ def convert_field_to_string(field, registry=None):
     )
 
 
-@convert_django_field.register(models.BigAutoField)
 @convert_django_field.register(models.AutoField)
+@convert_django_field.register(models.BigAutoField)
+@convert_django_field.register(models.SmallAutoField)
 def convert_field_to_id(field, registry=None):
     return ID(description=get_django_field_description(field), required=not field.null)
```
PoByBolek, line 199: `SmallAutoField` has been arouns since [Django 3.0](https://docs.djangoproject.com/en/5.0/releases/3.0/#models) from 2019-12-02, which we don't support anymore. So the check for `models.SmallAutoField` shouldn't be necessary anymore.

graphene_django/forms/tests/test_converter.py
```diff
@@ -75,8 +84,7 @@ def test_should_regex_convert_string():
 
 
 def test_should_uuid_convert_string():
-    if hasattr(forms, "UUIDField"):
-        assert_conversion(forms.UUIDField, UUID)
+    assert_conversion(forms.UUIDField, UUID)
```
PoByBolek, line 87: `forms.UUIDField` has been around since [Django 1.8](https://docs.djangoproject.com/en/5.0/releases/1.8/#new-data-types) from 2015-04-01.

graphene_django/rest_framework/tests/test_field_converter.py
```diff
@@ -96,8 +96,7 @@ def test_should_regex_convert_string():
 
 
 def test_should_uuid_convert_string():
-    if hasattr(serializers, "UUIDField"):
-        assert_conversion(serializers.UUIDField, graphene.String)
+    assert_conversion(serializers.UUIDField, graphene.String)
```
PoByBolek, line 100: DRF added support for `UUIDField` in [3.0.4](https://www.django-rest-framework.org/community/release-notes/#304) from 2015-01-28.  We [require DRF>= 3.6.3](https://github.com/graphql-python/graphene-django/blob/28c71c58f73e969bc05ebfdde06f578dd96e15d5/setup.py#L13)

graphene_django/tests/test_converter.py
```diff
@@ -166,14 +164,14 @@ def test_field_with_choices_convert_enum():
         help_text="Language", choices=(("es", "Spanish"), ("en", "English"))
     )
 
-    class TranslatedModel(models.Model):
+    class ChoicesModel(models.Model):
```
PoByBolek, line 167: This fixes a Django warning about the `TranslatedModel` being already registered:

> RuntimeWarning: Model 'test.translatedmodel' was already registered. Reloading models is not advised as it can lead to inconsistencies, most notably with related models.

examples/starwars/schema.py
```diff
@@ -62,16 +62,13 @@ class Query(graphene.ObjectType):
     node = relay.Node.Field()
     ships = DjangoConnectionField(Ship, description="All the ships.")
 
-    @resolve_only_args
-    def resolve_ships(self):
+    def resolve_ships(self, info):
         return get_ships()
```
PoByBolek, line 66: `resolve_only_args` is deprecated

PoByBolek:
I just noticed that this is basically a duplicate of #1513 :sweat_smile:

PoByBolek:
@sjdemartini @kiendang any thoughts on this?

sjdemartini (approved):
@PoByBolek Sorry for the delay, I haven't had much time to look at graphene-django recently. Overall the changes look good to me, and I appreciate the comments clarifying the updates you made (as well as the efforts to reduce noise with warnings, etc). I would generally prefer that the changes unrelated to the `DjangoConnectionField` `resolver` fix were put into their own PR to make this simpler to review and more targeted, but the separate commits helps. This seems a tad cleaner than the solution in https://github.com/graphql-python/graphene-django/pull/1513 (where the `super` call looks a bit odd, even if the effect is the same), so merging this seems good to me.

graphene_django/converter.py
```diff
@@ -190,19 +190,13 @@ def convert_field_to_string(field, registry=None):
...
 
 
-if hasattr(models, "SmallAutoField"):
-
-    @convert_django_field.register(models.SmallAutoField)
-    def convert_field_small_to_id(field, registry=None):
-        return convert_field_to_id(field, registry)
-
-
```
PoByBolek, line 205: Note that this could be considered a backwards-incompatible change as it removes the (undocumented) `convert_field_small_to_id()` function from the `graphene_django.converter` module.

So if you want, I can turn this into two PR: one only including c127aec75b33617b8eb10c84fb0733654ce70bf3 (the bugfix) and the remaining commits that fix the warnings.

PR #1534, opened 2024-09-15
remove dead code: singledispatch has been in the standard library ...

a-detiste:
…for many years

(BTW this function does not seems to be used anywhere anymore)


sjdemartini (approved):
Like like it's been stable since python 3.7 https://docs.python.org/3/library/functools.html#functools.singledispatch. LGTM!

kiendang (approved):

PR #1539, opened 2024-12-26
fix typed choices, make working with different Django 5x choices options

### LINKED ISSUE #1476: Mutation can not represent TextChoice value
**Note: for support questions, please use stackoverflow**. This repository's issues are reserved for feature requests and bug reports.

When executing a mutation that returns a Type of a Model that has a CharField with a TextChoices enum, it returns the following error:
`"Enum 'Status' cannot represent value: <AppointmentStatus.CANCELLED: 'C'>"`



* **If the current behavior is a bug, please provide the steps to reproduce and if possible a minimal demo of the problem** via
a github repo, https://repl.it or similar (you can use this template as a starting point: https://repl.it/@jkimbo/Graphene-Django-Example).


Define these classes

```
class AppointmentStatus(models.TextChoices):
    BOOKED = "B", "Booked"
    CANCELLED = "C", "Cancelled"
    

class Appointment(models.Model):
    id = models.UUIDField(default=uuid.uuid4, editable=False, primary_key=True)
    status = models.CharField(max_length=1, choices=AppointmentStatus)


class AppointmentType(DjangoObjectType):
    class Meta:
        model = Appointment


class CancelAppointment(Mutation):
    appointment = graphene.Field(AppointmentType, required=False)

    class Arguments:
        appointment_id = graphene.UUID(required=True)

    def mutate(self, info, appointment_id):
        appointment = Appointment.objects.get(id=appointment_id)
        return CancelAppointment(appointment=appointment)


class Mutation(object):
    cancel_appointment = CancelAppointment.Field()
```

When using this mutation and querying the result status

```
mutation cancelAppointment($appointmentId: UUID!) {
  cancelAppointment(
    appointmentId: $appointmentId
  ) {
    appointment {
      id
      status
    }
  }
}
```

and querying the result status I receive the following error:

```
"errors": [
    {
      "message": "Enum 'Status' cannot represent value: <AppointmentStatus.CANCELLED: 'C'>'",
      "locations": [
        {
          "line": 10,
          "column": 7
        }
      ],
      "path": [
        "cancelAppointment",
        "appointment",
        "status"
      ]
    }

```


The appointment status values should be shown without errors.


  - Version: graphene-django==3.1.5, python 3.9.13
  - Platform: Ubuntu 22.04

I'm trying to migrate from graphene-django 2.15.0 to 3.1.5.  

These are the whole traces of the problem:
```
File "/usr/local/lib/python3.9/site-packages/graphql/execution/execute.py", line 540, in execute_field       
  completed = self.complete_value(                                                                           
File "/usr/local/lib/python3.9/site-packages/graphql/execution/execute.py", line 639, in complete_value      
  return self.complete_leaf_value(cast(GraphQLLeafType, return_type), result)                                
File "/usr/local/lib/python3.9/site-packages/graphql/execution/execute.py", line 774, in complete_leaf_value 
  serialized_result = return_type.serialize(result)                                                          
File "/usr/local/lib/python3.9/site-packages/graphene/types/definitions.py", line 58, in serialize           
  return super(GrapheneEnumType, self).serialize(value)                                                      
File "/usr/local/lib/python3.9/site-packages/graphql/type/definition.py", line 1233, in serialize            
  raise GraphQLError(           
```

GeyseR:
An attempt to fix #1476

While working on the fix, I realized that Graphene doesn't support choice fields initialized with a types class itself without calling `choices`. Fixed this as well.

Django 5.1 introduced a convenient `normalize_choices` method that I use if a newer django version installed

### DIFF
--- a/graphene_django/compat.py
+++ b/graphene_django/compat.py
@@ -1,10 +1,11 @@
 import sys
+from collections.abc import Callable
 from pathlib import PurePath
 
 # For backwards compatibility, we import JSONField to have it available for import via
 # this compat module (https://github.com/graphql-python/graphene-django/issues/1428).
 # Django's JSONField is available in Django 3.2+ (the minimum version we support)
-from django.db.models import JSONField
+from django.db.models import Choices, JSONField
 
 
 class MissingType:
@@ -42,3 +43,23 @@ except ImportError:
 
     else:
         ArrayField = MissingType
+
+
+try:
+    from django.utils.choices import normalize_choices
+except ImportError:
+
+    def normalize_choices(choices):
+        if isinstance(choices, type) and issubclass(choices, Choices):
+            choices = choices.choices
+
+        if isinstance(choices, Callable):
+            choices = choices()
+
+        # In restframework==3.15.0, choices are not passed
+        # as OrderedDict anymore, so it's safer to check
+        # for a dict
+        if isinstance(choices, dict):
+            choices = choices.items()
+
+        return choices
--- a/graphene_django/converter.py
+++ b/graphene_django/converter.py
@@ -1,5 +1,4 @@
 import inspect
-from collections.abc import Callable
 from functools import partial, singledispatch, wraps
 
 from django.db import models
@@ -37,7 +36,7 @@ except ImportError:
     from graphql import assert_valid_name as assert_name
 from graphql.pyutils import register_description
 
-from .compat import ArrayField, HStoreField, RangeField
+from .compat import ArrayField, HStoreField, RangeField, normalize_choices
 from .fields import DjangoConnectionField, DjangoListField
 from .settings import graphene_settings
 from .utils.str_converters import to_const
@@ -61,6 +60,24 @@ class BlankValueField(Field):
         return blank_field_wrapper(resolver)
 
 
+class EnumValueField(BlankValueField):
+    def wrap_resolve(self, parent_resolver):
+        resolver = super().wrap_resolve(parent_resolver)
+
+        # create custom resolver
+        def enum_field_wrapper(func):
+            @wraps(func)
+            def wrapped_resolver(*args, **kwargs):
+                return_value = func(*args, **kwargs)
+                if isinstance(return_value, models.Choices):
+                    return_value = return_value.value
+                return return_value
+
+            return wrapped_resolver
+
+        return enum_field_wrapper(resolver)
+
+
 def convert_choice_name(name):
     name = to_const(force_str(name))
     try:
@@ -72,15 +89,7 @@ def convert_choice_name(name):
 
 def get_choices(choices):
     converted_names = []
-    if isinstance(choices, Callable):
-        choices = choices()
-
-    # In restframework==3.15.0, choices are not passed
-    # as OrderedDict anymore, so it's safer to check
-    # for a dict
-    if isinstance(choices, dict):
-        choices = choices.items()
-
+    choices = normalize_choices(choices)
     for value, help_text in choices:
         if isinstance(help_text, (tuple, list)):
             yield from get_choices(help_text)
@@ -157,7 +166,7 @@ def convert_django_field_with_choices(
 
         converted = EnumCls(
             description=get_django_field_description(field), required=required
-        ).mount_as(BlankValueField)
+        ).mount_as(EnumValueField)
     else:
         converted = convert_django_field(field, registry)
     if registry is not None:
--- a/graphene_django/tests/test_converter.py
+++ b/graphene_django/tests/test_converter.py
@@ -25,7 +25,7 @@ from ..converter import (
 )
 from ..registry import Registry
 from ..types import DjangoObjectType
-from .models import Article, Film, FilmDetails, Reporter
+from .models import Article, Film, FilmDetails, Reporter, TypedIntChoice, TypedStrChoice
 
 # from graphene.core.types.custom_scalars import DateTime, Time, JSONString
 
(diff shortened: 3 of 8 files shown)


graphene_django/compat.py
```diff
@@ -34,11 +36,49 @@ def __init__(self, *args, **kwargs):
...
 
     else:
         ArrayField = MissingType
+
+
+try:
+    from django.utils.choices import normalize_choices
+except ImportError:
+
+    def normalize_choices(choices):
```
sjdemartini, line 53: Good find on `normalize_choices` from Django! I see it seems to have a [different signature](https://github.com/django/django/blob/fcd9d08379a2aee3b2c49eab0d0b8db6fd66d091/django/utils/choices.py#L72) and much more extensive behavior than the `normalize_choices` fallback defined here, but I suppose it's a superset of the functionality (with the additional `depth` arg being optional) and not particularly harmful to have less extensive fallback behavior, so seems alright to me.

graphene_django/converter.py
```diff
@@ -61,6 +60,24 @@ def wrapped_resolver(*args, **kwargs):
         return blank_field_wrapper(resolver)
 
 
+class EnumValueField(Field):
```
sjdemartini, line 63: Looks like this effectively replaces `BlankValueField`. Can you add a comment or something stating as much, and "marking" `BlankValueField` as deprecated? I assume it's effectively only ever been used internally, but we shouldn't remove it without a major version bump, unfortunately.

graphene_django/tests/test_converter.py
```diff
@@ -475,3 +475,44 @@ def resolve_reporter(root, info):
     assert result.data == {
         "reporter": {"firstName": "Bridget", "aChoice": None},
     }
+
+
+def test_typed_choice_value():
+    """Test that choice fields with blank values work"""
```
sjdemartini, line 481: This doc comment is not correct.

sjdemartini (commented):
Thanks for the PR. You mention that this should fix https://github.com/graphql-python/graphene-django/issues/1476. Could you please add a test or two accordingly for the mutation behavior, to ensure it's fixed and won't regress in the future?

graphene_django/compat.py
GeyseR: You are right about the different signature, but the depth parameter is used for the recursive calls from inside the function, I don't think it is useful for outside callers

graphene_django/converter.py
GeyseR: actually, my idea was to inherit from the `BlankValueField` class, but I forgot to. The interesting thing is that I've seen the test that was added when this class was added (`test_choice_enum_blank_value`), but it didn't catch the bug in my code as the test itself was wrong (the field in the test instance was null instead of blank).
I've tested a bug in the test and classes inheritance, so they match my initial idea, but if you want me to move all the logic to the `EnumValueField` and deprecate `BlankValueField`, I can do that

GeyseR:
> You mention that this should fix https://github.com/graphql-python/graphene-django/issues/1476. Could you please add a test or two accordingly for the mutation behavior, to ensure it's fixed and won't regress in the future?

the source of the problem was that the library incorrectly resolved typed choice values assigned to model choice fields (which is a totally valid case and can happen not only in mutations). I've added a `test_typed_choice_value` test that covers that

sjdemartini:
> > You mention that this should fix #1476. Could you please add a test or two accordingly for the mutation behavior, to ensure it's fixed and won't regress in the future?
> 
> the source of the problem was that the library incorrectly resolved typed choice values assigned to model choice fields (which is a totally valid case and can happen not only in mutations). I've added a `test_typed_choice_value` test that covers that

Would you be able to add a test for the mutation specifically, for comprehensiveness and good measure?

graphene_django/tests/test_converter.py
```diff
@@ -443,35 +443,69 @@ def test_choice_enum_blank_value():
...
+    class ReporterType(DjangoObjectType):
+        class Meta:
+            model = Reporter
+            fields = ("typed_choice", "class_choice", "callable_choice")
+
+    class Query(graphene.ObjectType):
+        reporter = graphene.Field(ReporterType)
+
+        def resolve_reporter(root, info):
+            # assign choice values to the fields instead of their ste or int values
```
sjdemartini, line 484: What is "ste"? Typo?

GeyseR:
> Would you be able to add a test for the mutation specifically, for comprehensiveness and good measure?

extended the added test with the mutation call. Thanks for the suggestion!

sjdemartini (approved):
Sorry for the delay, thanks again for the work on this and for adding the extra mutation test!

graphene_django/converter.py
GeyseR: @sjdemartini how do you want to resolve this one?

graphene_django/compat.py
```diff
@@ -1,10 +1,12 @@
 import sys
+from collections.abc import Callable
 from pathlib import PurePath
 
 # For backwards compatibility, we import JSONField to have it available for import via
 # this compat module (https://github.com/graphql-python/graphene-django/issues/1428).
 # Django's JSONField is available in Django 3.2+ (the minimum version we support)
```
sjdemartini, line 7: minor: this comment should appear above the import of JSONField instead of above `import django`

graphene_django/converter.py
sjdemartini: I think what you have (with inheritance from BlankValueField) is fine 👍

graphene_django/compat.py
GeyseR: huh, it was an unused import of django, but the IDE showed it similar to the comment above, so I didn't notice it
image

remove the import.

Though, do you think it would make sense to add a linter rule for unused imports? (in a separate PR)

sjdemartini: Looks like the unused imports lint rule was disabled for the whole file https://github.com/graphql-python/graphene-django/blob/8d4a64a40dca6b2c0b355bdb89b9205309884deb/.ruff.toml#L28 but we should probably remove that and instead disable it on a per-import basis

GeyseR: removing the ruff rule for the compat.py file didn't break linting

PR #1540, opened 2024-12-26
add official Django 5.1 support

GeyseR:
adds explicit support for Django 5.1

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,8 +2,7 @@
 envlist =
     py{38,39,310}-django32
     py{38,39}-django42
-    py{310,311}-django{42,50,main}
-    py312-django{42,50,main}
+    py{310,311,312}-django{42,50,51,main}
     pre-commit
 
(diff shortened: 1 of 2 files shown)


sjdemartini (approved):
Thank you, LGTM!

tox.ini
```diff
@@ -2,8 +2,8 @@
 envlist =
     py{38,39,310}-django32
     py{38,39}-django42
-    py{310,311}-django{42,50,main}
-    py312-django{42,50,main}
+    py{310,311}-django{42,50,51,main}
+    py312-django{42,50,51,main}
```
kiendang, line 6: I think we can simplify this to

```suggestion
    py{310,311,312}-django{42,50,51,main}
```

GeyseR: good point, done!

kiendang (approved):
Thank you @GeyseR

PR #1544, opened 2025-04-14
chore: Add support for Django 5.2

rapsealk:
Hello graphene-django! This pull request introduces a support for Django 5.2, as a follow-up of #1540.

According to the docs, Django 5.2 supports Python 3.10, 3.11. 3.12, and 3.13, so I excluded Python 3.8 and 3.9.

https://docs.djangoproject.com/en/5.2/releases/5.2/#python-compatibility

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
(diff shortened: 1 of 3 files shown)


rapsealk:
#1546 may resolve the failure in GitHub Actions workflow.

PR #1550, opened 2025-06-14
Optional internal Python error masking with whitelist in GraphQL responses

DrillableBit:
By default, Graphene returns Django errors (e.g., `ValidationError`) inside the GraphQL `data` object, while `GraphQLErrors` -including internal Python exceptions like `ZeroDivisionError` - are returned outside the `data` field.

For public APIs, exposing full internal error details can be a security risk and a poor user experience. Currently, Graphene lacks built-in support for masking such unhandled internal exceptions.

This PR introduces an opt-in error masking feature that:

 - Masks internal exceptions with a generic error message by default

 - Allows whitelisting of specific exception classes to be returned fully

- Enables configuration via `GRAPHENE_ERRORS` settings, supporting custom messages and whitelisted error classes

This improves API security and usability by preventing leakage of sensitive error information while still exposing known, safe errors.

### DIFF
--- a/docs/error-masking.rst
+++ b/docs/error-masking.rst
@@ -1,0 +1,49 @@
+Custom GraphQL Error Masking
+============================
+
+This project includes a custom error formatting function for GraphQL
+responses that masks sensitive error details from clients.
+
+Purpose
+-------
+
+- Prevent exposing internal error details for security and user experience.
+- Allow whitelisting of exception classes that should be exposed as-is.
+- Return a generic error message for all other exceptions.
+
+Configuration
+-------------
+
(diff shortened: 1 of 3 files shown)


DrillableBit:
There might be an error in this PR - in that it might require native GraphQLErrors to be whitelisted when masking is set to True. 

If there's any interest in this PR, I'll fix it, but I ended up making a custom view for my project.

syrusakbary:
The upcoming changes into `graphql-server` will make of this a non-issue, as they will make most of the current error behavior fixed.

I'll keep the PR opened until graphene-django gets updated with the new server

